PR #11, opened 2024-12-06
AXON-26 fix create JIRA issue

bwieger-atlassian-com:
- Fix create JIRA issue screen. 
- Bug: create JIRA page was blank. 
- Root Cause: A new i18n requirement from a version bump in atlaskit/panel
- Resolution: Roll back the version of the dependency 
- Drawback: exposes a vulnerability. 
- Drawback Mitigation: will fix forward. This is exposed in the last stable version, so we aren't regressing.


sdzh-atlassian (approved):
Thanks for digging into this!

PR #12, opened 2024-12-06
[AXON-23] fix: BB schema issue and add tests

sdzh-atlassian:
### What is this?

This is a fix for the problem reported to us in Atlascode `3.0.13`:
image

Apparently at some point someone introduced a change to the `bitbucket-pipelines` file schema that rendered the schema itself invalid.

To catch such issues in the future, this change includes a new unit test to specifically make sure that the schema compiles and works on a simple file

### How was this tested?

```
npm run lint # OK
npm run test # OK
npm run extension:install # validated that the error went away:
```

image


sdzh-atlassian:
I've found a much more recent version of the pipeline file schema - will update it in a separate PR

bwieger-atlassian-com:
LGTM

bwieger-atlassian-com (approved):

PR #14, opened 2024-12-06
[AXON-19] fix: pass FX3 client vars to build

sdzh-atlassian:
### What is this?

When doing the round of bug bashing today, we caught that the `atlascode` feature flag client initialization vars are not getting passed to the build environment correctly.

Now they are 😉 

### How was this tested?

This is related to CI - so I've only run a pipeline to make sure things work, and checked that the environment variables propagate to the build env correctly. Once this is merged, I'll check the next `nightly` for any feature flag initialization issues

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,8 +6,17 @@ on:
       - '**'
 
 jobs:
+
   build:
+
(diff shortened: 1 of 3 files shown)


bwieger-atlassian-com (approved):
LGTM

PR #16, opened 2024-12-09
[AXON-23] chore: remove echo

sdzh-atlassian:
### What is this:

Remove the `echo` we forgot in commands


sdzh-atlassian:
Also in this: `secrets: inherit` for when we call nightly build from release

bwieger-atlassian-com (approved):

PR #32, opened 2024-12-13
[AXON-37] (PR-1) Fix .vscodeignore to not bloat the extension artifact

sdzh-atlassian:
### What is this?

I've noticed that our extension artifact has become quite bloated - and the problem became 1000 times worse once I've tried creating a local cache for e2e tests :D

This should hopefully curtail the bloat ;)

| | Latest nightly (3.1.7) | After the fix |
|-|-|-|
| Size | 3.77 MB | 1.7 MB |
| Composition | image | image |

### How was this tested?

* _Not thoroughly_ - but I've installed the version and poked around a bunch to make sure that icons still work, and components don't look weird

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,3 +1,32 @@
+# Allowed:
+# build/
+# resources/
+# images/
+# LICENSE
+
+.env
+.env.example
+.eslintignore
+.mocharc.js
+.nvmrc
+.prettierignore
+CODEOWNERS
+HOMEPAGE.md
+changelog.md
+nightlyver.sh
+package.json
+readme.md
+setupTests.js
+tsconfig.notest.json
+tsconfig.ui-test.json
+ui-test-settings.json
+vscode-stylenames.txt
+.devcontainer/
+.github/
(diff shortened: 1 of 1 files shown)


.vscodeignore
```diff
@@ -1,3 +1,32 @@
...
+.env.example
+.eslintignore
+.mocharc.js
+.nvmrc
+.prettierignore
+CODEOWNERS
+HOMEPAGE.md
+changelog.md
+nightlyver.sh
+package.json
```
bwieger-atlassian-com, line 17: Interested to see how this might impact the marketplace page after we do a release

bwieger-atlassian-com (approved):
Approved with some lingering concern

.vscodeignore
sdzh-atlassian: True, that's not something I could test - however, looks like `readme` and `changelog` are pulled in regardless, as well as package.json:
image

I think most of the contents of the marketplace page should be built on links - so it might be fine? But we'll only see after the 1st nightly

PR #34, opened 2024-12-14
[AXON-37] Run E2E tests in CI

sdzh-atlassian:
### What is this?

A little something-something to run our fancy new E2E test in CI, as well

An keen-eyed reader may have already spotted the brand new docker image published to `ghcr.io` and linked to `atlascode`:
image

This PR is how it got there! Now that we have a Dockerfile, why let it just sit boring-ly in our repo and do nothing? We should instead:

 * Create a workflow to build and publish it for our other workflows to reuse
 * Hook up the end-to-end tests to be executed in our `build` and `release` workflows!

> Note: one thing I'm not super happy about - it would be nice if we could first explicitly build a `.vsix` package file, and then reuse it in the E2E tests. The current E2E action is a straight upgrade from the regular `vsce package` in the `build` workflow, so it's OK to hotswap them. However, for the `release` flows we want a fine control over the version and release type - so we can't do it there. Perhaps we could address this in a follow-up PR? For now I'd be happy to just let the E2E tests run in builds in _some_ capacity :)

### How was this tested?

* It's purely a CI change - so through CI? 🤔 
* `release` workflows were deliberately excluded - `nigtly` will test itself soon enough 😉


bwieger-atlassian-com (dismissed):
LGTM, but you may need to change command verbiage based on previous PR comments.

sdzh-atlassian:
Addressed the command naming in the [follow-up PR](https://github.com/atlassian/atlascode/pull/36)

PR #41, opened 2024-12-17
Patch: VSCODE-1050: PAT auth for hosted BitBucket(like for hosted Jira) 

bwieger-atlassian-com:
Address comments on https://github.com/atlassian/atlascode/pull/28/files

### DIFF
--- a/src/atlclients/loginManager.ts
+++ b/src/atlclients/loginManager.ts
@@ -29,8 +29,6 @@ import { Logger } from '../logger';
 import { OAuthDancer } from './oauthDancer';
 import { SiteManager } from '../siteManager';
 
-const slugRegex = /[\[\:\/\?#@\!\$&'\(\)\*\+,;\=%\\\[\]]/gi;
-
 export class LoginManager {
     private _dancer: OAuthDancer = OAuthDancer.Instance;
     private _jiraAuthenticator: JiraAuthenticator;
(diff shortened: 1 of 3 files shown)


src/atlclients/loginManager.ts
```diff
@@ -156,24 +154,29 @@ export class LoginManager {
...
             case ProductJira.key:
                 siteDetailsUrl = `${protocol}//${site.host}${contextPath}/rest/api/2/myself`;
                 avatarUrl = `${protocol}//${site.host}${contextPath}/images/fav-jcore.png`;
                 apiUrl = `${protocol}//${site.host}${contextPath}/rest`;
                 break;
             case ProductBitbucket.key:
-                const bbCredentials = credentials as BasicAuthInfo;
-                siteDetailsUrl = `${protocol}//${
-                    site.host
-                }${contextPath}/rest/api/1.0/users/${bbCredentials.username.replace(slugRegex, '_')}?avatarSize=64`;
```
Fotiman, line 169: This line would replace the `@` symbol in a username like "john.doe@example.com" with an underscore ("john.doe_example.com").  This no longer happens where `ausername` is defined below, so the `siteDetailsUrl` contains a value like "john.doe%40example.com" instead, and this fails.

bwieger-atlassian-com: @Fotiman Thank you! Will get a fix going for this.

Issue #46, opened 2024-12-18
[JIRA] JIRA Issue Details Blank Page (no bitbucket)


manuartero:
**Describe the bug**
The issue details won't open when clicking on an issue

**To Reproduce**

1. Go to Jira ISSUES PANE
2. Click on any issue
3. the issue details won't open

**Expected behavior**
The issue detail is opened in the editor

**Screenshots**

check video:

https://github.com/user-attachments/assets/c3beaf7a-2a4c-4853-ba68-35e496b71932

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version: **3.2.1**

![Screenshot 2024-12-18 at 09 56 51](https://github.com/user-attachments/assets/4f92430f-6478-4b63-b30a-89b96dd7af6d)

 - Visual Studio Code Version: **1.96.0** (Universal)

```
Version: 1.96.0 (Universal)
Commit: 138f619c86f1199955d53b4166bef66ef252935c
Date: 2024-12-11T02:29:09.626Z (1 wk ago)
Electron: 32.2.6
ElectronBuildId: 10629634
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 23.6.0
```

 - Is this a local or cloud environment? local, connected to a regular jira instance
 - If Cloud, Browser (chrome, safari),---
 - If Local, OS (macOS, windows, linux) macOs

**Additional context**

I've seen this extension working actually, cant tell when did it stop working but should be some update 

pasting an error from the developer tools inspector:

![Screenshot 2024-12-18 at 09 40 54](https://github.com/user-attachments/assets/e1282288-1565-4b8e-a4d6-d59c4997346b)

***

NOTE: this is different issue from 

- #26

there is no BitBucket here, please do **not** mark as duplicated

sdzh-atlassian:
Hi @manuartero, thank you very much for reporting this! Sorry you had to run into a bad experience 😦 We certainly won't be making this as a duplicate

A special thanks for the error from the inspector - that's actually kind of super helpful to us ;D

I couldn't reproduce this right away, but I think that might have to do with something different going on with your issues that we don't have at the moment. I also have a sneaking suspicion that it's an error coming from the JIRA clients we are using.

Just to exclude some things, could I trouble you to try reproducing this in the extension version `3.0.10`? This would help us rule out any recent changes we've made. I would expect the behaviour to be the same - but let's see what actually happens 😉

[...]

PR #48, opened 2024-12-18
[AXON-40] fix: add slightly more graceful failures in JiraIssuePage

sdzh-atlassian:
### What is this?

A temporary workaround to put into `nightly` so that we can hopefully dig a little deeper into #46 

With this, if we try to render markup for an `undefined` field - we  skip rendering that field, and we get an error message:
image

Some considerations on this:

 * I might be missing something in the implementation - but this is super weird. The only field rendered on `JiraIssuePage` which could fail with that exact step should be `Summary` - but that feels _impossible_?
 * By no means is this a long-term solution
 * We _really_ need proper error handling in our pages :( Might be a good early task for @cabella-dot? 😉
 * Randomly spotted a bunch of copy-pasted code and moved it to a loop

### How was this tested?

* Rendered the page with some fields hardcoded to `undefined as any`, saw the error message and the page without the fields
* Regular CI stuff

### DIFF
--- a/src/webviews/components/issue/JiraIssuePage.tsx
+++ b/src/webviews/components/issue/JiraIssuePage.tsx
@@ -72,6 +72,17 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
         this.state = emptyState;
     }
 
+    // TODO: proper error handling in webviews :'(
+    // This is a temporary workaround to hopefully troubleshoot
+    // https://github.com/atlassian/atlascode/issues/46
+    override getInputMarkup(field: FieldUI, editmode?: boolean, context?: String) {
+        if (!field) {
+            console.warn(`Field error - no field when trying to render ${context}`);
+            return null;
+        }
+        return super.getInputMarkup(field, editmode);
+    }
+
     getProjectKey = (): string => {
         return this.state.key.substring(0, this.state.key.indexOf('-'));
     };
@@ -476,7 +487,7 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
                             />
                         </Tooltip>
                     </div>
-                    <h2>{this.getInputMarkup(this.state.fields['summary'], true)}</h2>
+                    <h2>{this.getInputMarkup(this.state.fields['summary'], true, 'summary')}</h2>
                 </div>
                 {this.state.isErrorBannerOpen && (
                     <ErrorBanner onDismissError={this.handleDismissError} errorDetails={this.state.errorDetails} />
@@ -491,7 +502,7 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
                 {this.state.fields['description'] && (
                     <div className="ac-vpadding">
                         <label className="ac-field-label">{this.state.fields['description'].name}</label>
-                        {this.getInputMarkup(this.state.fields['description'], true)}
+                        {this.getInputMarkup(this.state.fields['description'], true, 'description')}
                     </div>
                 )}
                 {this.state.fields['attachment'] &&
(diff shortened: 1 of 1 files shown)


src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -72,6 +72,17 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
         this.state = emptyState;
     }
 
+    // TODO: proper error handling in webviews :'(
+    // This is a temporary workaround to hopefully troubleshoot
+    // https://github.com/atlassian/atlascode/issues/46
+    getInputMarkup(field: FieldUI, editmode?: boolean, context?: String) {
```
bwieger-atlassian-com, line 78: Refactor suggestion.

Goal: Reduce copy-paste error on the callers

```
getInputMarkup(fields: /*typeOf(this.state.fields)*/, fieldKey?: String, editmode?: boolean) {
     const field = fields[fieldKey];
     if (!field) {
          console.error(`Field error - no field when trying to render ${fieldKey}`);
          return null;
      }
      return super.getInputMarkup(field, editmode);
}
```

bwieger-atlassian-com: Also, it looks like the `editmode` param is always true.. Maybe we can drop from the signature here?

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -72,6 +72,17 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
...
+        if (!field) {
+            console.error(`Field error - no field when trying to render ${context}`);
```
bwieger-atlassian-com, line 80: Not sure this should be error given all the versions of JIRAs out there. Perhaps a warn instead?

src/webviews/components/issue/JiraIssuePage.tsx
cabella-dot, line 78: Looks good to me!

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -491,7 +502,7 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
                 {this.state.fields['description'] && (
                     <div className="ac-vpadding">
```
cabella-dot, line 503: not related to this PR but omg this frontend code

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -775,46 +786,26 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
...
-                        {this.getInputMarkup(this.state.fields['components'], true)}
-                    </div>
-                )}
-                {this.state.fields['fixVersions'] && (
-                    <div className="ac-vpadding">
-                        <label className="ac-field-label">{this.state.fields['fixVersions'].name}</label>
-                        {this.getInputMarkup(this.state.fields['fixVersions'], true)}
-                    </div>
+
+                {['assignee', 'reporter', 'labels', 'priority', 'components', 'fixVersions'].map((field) =>
```
cabella-dot, line 790: nice!

cabella-dot (dismissed):

src/webviews/components/issue/JiraIssuePage.tsx
sdzh-atlassian: Unfortunately, we have limited flexibility with the signature here since this is an override from base class xD
I went this route to minimize changes made to code.

Added an explicit `override` for clarity though - and replaced `console.error` with `console.warn`:

image

src/webviews/components/issue/JiraIssuePage.tsx
sdzh-atlassian: Replaced `console.error` with `console.warn`:

image

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -826,7 +817,7 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
                 markups.push(
                     <div className="ac-vpadding">
                         <label className="ac-field-label">{field.name}</label>
-                        {this.getInputMarkup(field, true)}
+                        {this.getInputMarkup(field, true, `Advanced sidebar`)}
```
sdzh-atlassian, line 820: this is where `field` is resolved explicitly

bwieger-atlassian-com (approved):

PR #49, opened 2024-12-18
AXON-32 First iteration on e2e tests - no auth

bwieger-atlassian-com:
- Some more in depth tests 
- Doesn't use any auth 
- There's a bug with the library we are using, so skipping that type of test for the moment. See comments in file for more.


.mocharc.js
```diff
@@ -1,6 +1,6 @@
 // increase default test case timeout to 5 seconds
 module.exports = {
-	"timeout": 5000,
+	"timeout": 60000,
```
bwieger-atlassian-com, line 3: Extend so that we can have enough time to debug stuff when using the debugger.

src/react/atlascode/common/StatusBarPanel.tsx
```diff
@@ -52,7 +52,7 @@ export const StatusBarPanel: React.FunctionComponent<StatusBarPanelProps> = memo
                 <ExpansionPanelSummary
                     expandIcon={<ExpandMoreIcon />}
                     aria-controls={`${configSection}-${ConfigSubSection.Auth}-content`}
-                    id={`${configSection}-${ConfigSubSection.Auth}-header`}
+                    id={`${configSection}-${ConfigSubSection.Status}-header`}
```
bwieger-atlassian-com, line 55: Bug from before. Overlapping IDs

src/react/atlascode/config/auth/CloudAuthButton.tsx
```diff
@@ -18,7 +18,13 @@ export const CloudAuthButton: React.FunctionComponent<CloudAuthButtonProps> = ({
     };
 
     return (
-        <Button variant="contained" color="primary" startIcon={<CloudIcon />} onClick={() => handleCloudProd()}>
+        <Button
+            id="settings.login.cloud.button"
```
sdzh-atlassian, line 22: NIT: I'd strongly recommend using `kebab-case` for element ids because it is in line with other `id` values in the repo - as well as best practices, to my knowledge.

TIL dots were even allowed in `id`s , seeing them kind of breaks my brain a little bit because of selector syntax where `#a.b.c` means "Element with id `a` that has classes `b` and `c`" - [relevant example](https://css-tricks.com/multiple-class-id-selectors/)

sdzh-atlassian (approved):
Please consider changing the `id` naming convention, but otherwise SHIP IIIIIT 🚢

.vscode/launch.json.example
```diff
@@ -5,6 +5,27 @@
 {
     "version": "0.2.0",
     "configurations": [
+        {
+            "name": "Debug UI Tests",
```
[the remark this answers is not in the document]
bwieger-atlassian-com: Fair enough. Next PR.

.vscode/launch.json.example
```diff
@@ -5,6 +5,27 @@
...
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceFolder}/node_modules/.bin/extest",
+            "windows": {
+              "program": "${workspaceFolder}/node_modules/vscode-extension-tester/out/cli.js",
```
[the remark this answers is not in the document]
bwieger-atlassian-com: Oh... this is just copy-pasta from the example code. And it worked for me.

src/react/atlascode/config/auth/CloudAuthButton.tsx
bwieger-atlassian-com: Ahhh okay. Will fix up in a follow up PR.

bwieger-atlassian-com:
Will change IDs to kebab style in follow up PR.

PR #51, opened 2024-12-19
kebab id for some recent id changes 

bwieger-atlassian-com:
- Changes some html ids from dot style to kebab style.


sdzh-atlassian (approved):
Looks good, let's ship it! 🚢

PR #54, opened 2024-12-20
AXON-32 - Auth Test framework - Auth we go! (works only on local, no CI) 

bwieger-atlassian-com:
- Logins a test user using API Token. API token is set via environment variables. Check Repo secrets

- For community members, if needed, please use CI builds to run these tests.


sdzh-atlassian (approved):
SHIP ITTTTT (from pair work) 🚢

PR #58, opened 2024-12-24
[AXON-41] Implement AtlascodeErrorBoundary

sdzh-atlassian:
### What is this?

This is the 1st PR in a series related to improving our observability when something goes wrong.

In React, there's a concept of an [Error Boundary](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary) - a special component which is intended to address rendering errors in the tree, and which is very notably missing in our pages 😉

Further, we are already pulling [@atlaskit/analytics-next](https://atlaskit.atlassian.com/packages/analytics/analytics-next) as a dependency, which conveniently offers an [implementation](https://atlaskit.atlassian.com/packages/analytics/analytics-next/docs/error-boundary) we could use.

After the changes in this PR, we can receive an event with some data on rendering errors occuring in prod:
image

> ⚠️ Note: The exact structure and content of the analytics event are temporary and subject to change

#### What does this PR have, mechanically?

* Basic support for firing error-related events from webviews with this architecture:
image
    * New "common" action type `SendAnalytics`
    * Handling for it in all controllers
    * Type definitions
 * `AtlascodeErrorBoundary` implementation and usage 

#### What is notably NOT included here?
 * **Tests** - I ran into dependency conflicts when trying to pull react testing utils. Might make sense to add them in a separate PR after bumping `react` version
 * `AtlascodeErrorBoundary` usage in other pages - to not overwhelm reviewers further ;D

### How was this tested?

I've introduced the new error boundary to our `Settings` page, and then added a special component that fails on rendering. With that, when running our settings page, it was possible to show alternative HTML on render errors, while also receiving an event in the analityics service:

image
(all the extra components were removed during clean-up)

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -2,6 +2,7 @@ import { ScreenEvent, TrackEvent, UIEvent } from './analytics-node-client/src/ty
 import { DetailedSiteInfo, isEmptySiteInfo, Product, ProductJira, SiteInfo } from './atlclients/authInfo';
 import { BitbucketIssuesTreeViewId, PullRequestTreeViewId } from './constants';
 import { Container } from './container';
+import { UIErrorInfo } from './analyticsTypes';
 
 // IMPORTANT
 // Make sure there is a corresponding event with the correct attributes in the Data Portal for any event created here.
@@ -254,6 +255,13 @@ export async function viewScreenEvent(
 }
 
 // UI Events
+
+export async function uiErrorEvent(errorInfo: UIErrorInfo): Promise<TrackEvent> {
+    const e = trackEvent('failedTest', 'ui', {
+        attributes: { ...errorInfo },
+    });
+    return e;
+}
 
 export async function bbIssuesPaginationEvent(): Promise<UIEvent> {
     const e = {
--- a/src/lib/webview/controller/welcome/welcomeWebviewController.ts
+++ b/src/lib/webview/controller/welcome/welcomeWebviewController.ts
@@ -76,6 +76,7 @@ export class WelcomeWebviewController implements WebviewController<WelcomeInitMe
                 break;
             }
 
+            case CommonActionType.SendAnalytics:
             case CommonActionType.CopyLink:
             case CommonActionType.OpenJiraIssue:
             case CommonActionType.ExternalLink:
--- a/src/react/atlascode/common/ErrorBoundary.tsx
+++ b/src/react/atlascode/common/ErrorBoundary.tsx
@@ -1,0 +1,70 @@
+import React, { useCallback } from 'react';
+// eslint-disable-next-line no-restricted-imports
+import { AnalyticsErrorBoundary, AnalyticsListener, UIAnalyticsEvent } from '@atlaskit/analytics-next';
+import { CommonActionType } from 'src/lib/ipc/fromUI/common';
+import { AnalyticsChannels, UIAnalyticsContext } from 'src/analyticsTypes';
+
+const STACK_LIMIT = 150;
+const COMPONENT_GLOBAL = 'global';
+
+export type AtlascodeErrorBoundaryProps = {
+    postMessageFunc: (action: any) => void;
+    context: UIAnalyticsContext;
+    children?: React.ReactNode;
+};
+
+/**
+ * Global error boundary for Atlascode UI components
+ *
+ * Put this somewhere close to the root of your component tree.
+ * It will catch all rendering errors in the subtree and send them to the main process.
+ *
+ * Add `AnalyticsErrorBoundary` around the specific components you'd like to handle and track separately.
+ *
+ * @param {AtlascodeErrorBoundaryProps} props
+ * @returns
+ */
+export const AtlascodeErrorBoundary: React.FC<AtlascodeErrorBoundaryProps> = ({
+    children,
+    postMessageFunc,
+    context,
+}: AtlascodeErrorBoundaryProps) => {
+    const onRenderError = useCallback(
+        (event: UIAnalyticsEvent) => {
+            const { error, info, ...rest } = event.payload.attributes;
+            postMessageFunc({
+                type: CommonActionType.SendAnalytics,
+                errorInfo: {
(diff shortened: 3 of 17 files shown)


src/analytics.ts
```diff
@@ -255,6 +256,13 @@ export async function viewScreenEvent(
 
 // UI Events
 
+export async function uiErrorEvent(errorInfo: UIErrorInfo): Promise<TrackEvent> {
+    const e = trackEvent('failedTest', 'ui', {
```
sdzh-atlassian, line 260: Intentionally leaving the temporary event here for now while we figure out a more permanent structure for events and metrics

src/lib/webview/controller/welcome/welcomeWebviewController.ts
```diff
@@ -76,6 +76,7 @@ export class WelcomeWebviewController implements WebviewController<WelcomeInitMe
                 break;
             }
 
+            case CommonActionType.SendAnalytics:
```
bwieger-atlassian-com, line 79: Does this break D.R.Y? 

Can we import a function that can be parameterized with the right messaging tags?

src/react/atlascode/common/ErrorBoundary.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useCallback } from 'react';
+// eslint-disable-next-line no-restricted-imports
+import { AnalyticsErrorBoundary, AnalyticsListener, UIAnalyticsEvent } from '@atlaskit/analytics-next';
+import { CommonActionType } from 'src/lib/ipc/fromUI/common';
+import { AnalyticsChannels, UIAnalyticsContext } from 'src/analyticsTypes';
+
+const STACK_LIMIT = 150;
```
bwieger-atlassian-com, line 7: What reason for this number?

bwieger-atlassian-com (commented):
Some questions

src/react/atlascode/common/ErrorBoundary.tsx
sdzh-atlassian: It's quite arbitrary ;D Just some reasonable max length for the top of the stack.
I originally had some considerations about using it as a `GROUP BY` field in splunk (where it would be good if similar errors could fold into one top-of-the-stack group) - but honestly, it's not very well thought-through xD

Happy to change it or nuke it altogether

src/react/atlascode/common/ErrorBoundary.tsx
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Put this somewhere close to the root of your component tree.
+ * It will catch all rendering errors in the subtree and send them to the main process.
+ *
+ * Add `AnalyticsErrorBoundary` around the specific components you'd like to handle and track separately.
+ *
+ * @param {AtlascodeErrorBoundaryProps} props
+ * @returns
+ */
+export const AtlascodeErrorBoundary: React.FC<AtlascodeErrorBoundaryProps> = ({
```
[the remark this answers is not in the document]
sdzh-atlassian: Here we are reusing the 2 elements which are exactly like that - `AnalyticsListener` and `AnalyticsErrorBoundary`
Normally, you're not really supposed to have them right next to each other, `AnalyticsListener` goes to the root of the tree, and `AnalyticsErrorBoundary` goes to the specific place somewhere (e.g. around your form).

However, we're introducing a very high-level error boundary - so it felt sensible to put this element in place, and unfortunately there isn't really an already-available component for this.

As a bonus, having this component becomes particularly useful if we want to add some extra stuff into every page down the line (e.g. I've been heavily leveraging it to add a top panel which lets you induce errors on any page:
image

sdzh-atlassian (commented):
oh my god it took so long to figure out how to submit pending comments 💀

src/lib/webview/controller/welcome/welcomeWebviewController.ts
sdzh-atlassian: Sorry, I'm not sure I understand what you mean - could you elaborate on the idea of importing a function?

For some context on the repetition here - the need for it here comes from [this line](https://github.com/atlassian/atlascode/blob/3a11fea633c6e8272b6c3ff646dd3f6d66f46755/src/lib/webview/controller/welcome/welcomeWebviewController.ts#L94) in all the files, which makes typescript check on compile-time that the `switch` statement can't possibly go to `default`

I don't really like it, but I kind of understand the rationale 🤔 For this PR, I wanted to avoid changing things which aren't absolutely necessary - but I would be onboard with refactoring this in the future 😉

bwieger-atlassian-com: Fair enough. Making the code pretty can take a backseat for now.

src/react/atlascode/common/ErrorBoundary.tsx
bwieger-atlassian-com: Can we make it a huge number until we break something?

src/react/atlascode/common/ErrorBoundary.tsx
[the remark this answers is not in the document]
bwieger-atlassian-com: Sounds good. 

Outside of testing purposes, I'm a little hesitant to use this to "add some extra stuff into every page down the line" 

Thinking that each component / wrapper should have a single job to do.

bwieger-atlassian-com (approved):
Approving so you can move forward. 

I am concerned about the 150 stack limit being too low.

src/react/atlascode/common/ErrorBoundary.tsx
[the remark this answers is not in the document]
sdzh-atlassian: Agreed! :) I was mainly thinking about testing/debugging tools and callback error interception logic

src/react/atlascode/common/ErrorBoundary.tsx
sdzh-atlassian: Sure thing, will do in a follow-up PR

PR #59, opened 2024-12-24
[AXON-41] Add error boundary to all pages except JIRA

sdzh-atlassian:
### What is this?

`AtlascodeErrorBoundary` is now added to all pages of the newer architecture

### How was this tested?

 * Built the extension, poked around, induced and observed some failures

### DIFF
--- a/src/analyticsTypes.ts
+++ b/src/analyticsTypes.ts
@@ -8,6 +8,24 @@ export enum AnalyticsChannels {
     AtlascodeUiErrors = 'atlascode.ui.errors',
 }
 
+export enum AnalyticsView {
+    OnboardingPage = 'onboarding',
+    SettingsPage = 'settings',
+    WelcomePage = 'welcome',
+
+    BitbucketIssuePage = 'bitbucketIssue',
+    PullRequestPage = 'pullRequest',
+
+    CreateBitbucketIssuePage = 'createBitbucketIssue',
+    CreatePullRequestPage = 'createPullRequest',
+
+    PipelineSummaryPage = 'pipelineSummary',
+
+    StartWorkPage = 'startWork',
+
+    Other = 'other',
+}
+
 export type UIAnalyticsContext = {
     view: string;
 };
--- a/src/react/atlascode/config/ConfigPage.tsx
+++ b/src/react/atlascode/config/ConfigPage.tsx
@@ -34,6 +34,7 @@ import ToggleButtonGroup from '@material-ui/lab/ToggleButtonGroup';
 import WorkIcon from '@material-ui/icons/Work';
 import equal from 'fast-deep-equal/es6';
 import { AtlascodeErrorBoundary } from '../common/ErrorBoundary';
+import { AnalyticsView } from 'src/analyticsTypes';
 
 const useStyles = makeStyles(
     (theme: Theme) =>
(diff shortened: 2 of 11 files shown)


src/analyticsTypes.ts
```diff
@@ -8,6 +8,24 @@ export enum AnalyticsChannels {
     AtlascodeUiErrors = 'atlascode.ui.errors',
 }
 
+export enum AnalyticsView {
+    OnboardingPage = 'onboarding',
+    SettingsPage = 'settings',
+    WelcomePage = 'welcome',
+
+    BitbucketIssuePage = 'bitbucketIssue',
```
bwieger-atlassian-com, line 16: Are we happy with the naming conventions here as we look at the next several years?

src/react/atlascode/config/ConfigPage.tsx
```diff
@@ -170,7 +171,10 @@ const ConfigPage: React.FunctionComponent = () => {
     return (
         <ConfigControllerContext.Provider value={controller}>
             <AuthDialogControllerContext.Provider value={authDialogController}>
-                <AtlascodeErrorBoundary postMessageFunc={controller.postMessage} context={{ view: 'config' }}>
+                <AtlascodeErrorBoundary
```
bwieger-atlassian-com, line 174: Are we missing the closing tag here? 

`</AtlascodeErrorBoundary>`

src/react/atlascode/pipelines/pipelineSummaryController.ts
```diff
@@ -138,6 +140,7 @@ export function usePipelineSummaryController(): [PipelineSummaryState, PipelineS
             refresh: sendRefresh,
             rerun: rerun,
             fetchLogs: fetchLogs,
+            postMessage: postMessage,
         };
     }, [sendRefresh, rerun, fetchLogs]);
```
bwieger-atlassian-com, line 145: See GitHub Action warning about React Hook

bwieger-atlassian-com (commented):
Couple of questions

src/react/atlascode/config/ConfigPage.tsx
sdzh-atlassian: Sorry about the confusion - this page had a boundary added in a previous PR, so the closing tag is already there and not in the changeset

src/analyticsTypes.ts
sdzh-atlassian: The honest answer is that I'm not particularly happy with any of this, lol :D The idea was to somehow list the pages we have today, to somehow represent them in analytics. This isn't very future proof - for what it's worth, it's kind of already not very "present-proof", with how there are 2 `start work` pages xD

Super open to alternative suggestions here

bwieger-atlassian-com: Here are some prior art on this: 
https://posthog.com/docs/product-analytics/best-practices

In summary: 
- use colon to separate main ideas `:`
- use underscore to separate words `_` in a given idea
- start with the high level and go to the low level 
- version the appropriate idea (eg: v1, v2, v3 ...) 
- do keep a consistent vocabulary (eg: always `...button:click`, never `...clicky_thing:tap`)

Good Example: 
`start_work_page:start_work_button:click`
`start_work_page_v2:start_work_button:click`

Don'ts:
- Don't use old vs new. What happens when you need the next news?

src/analyticsTypes.ts
```diff
@@ -8,6 +8,24 @@ export enum AnalyticsChannels {
     AtlascodeUiErrors = 'atlascode.ui.errors',
 }
 
+export enum AnalyticsView {
+    OnboardingPage = 'onboarding',
```
bwieger-atlassian-com, line 12: Example here: 
`onboarding_page:view`

src/analyticsTypes.ts
sdzh-atlassian: Hmm, I'm happy with any naming convention here, so let's go with this :) Let's address it in a follow-up PR though if that's OK? Just to minimize the number of rebases :D

bwieger-atlassian-com (approved):
Approving with the note that we will have a follow up for renaming the analytics events

sdzh-atlassian:
Rebased onto main, merging

PR #60, opened 2024-12-26
[AXON-41] Add error boundary to JIRA pages

sdzh-atlassian:
### What is this?

This PR adds `AtlascodeErrorBoundary` to the remaining 3 pages, implemented with the older architecture and IPC.

### How was this tested?

Built the extension, ran things in debug, made sure analytics events are emitted

### DIFF
--- a/src/analyticsTypes.ts
+++ b/src/analyticsTypes.ts
@@ -19,9 +19,13 @@ export enum AnalyticsView {
     CreateBitbucketIssuePage = 'createBitbucketIssue',
     CreatePullRequestPage = 'createPullRequest',
 
+    CreateJiraIssuePage = 'createJiraIssue',
+    JiraIssuePage = 'jiraIssue',
+
     PipelineSummaryPage = 'pipelineSummary',
 
     StartWorkPage = 'startWork',
+    OldStartWorkPage = 'oldStartWork',
 
     Other = 'other',
 }
--- a/src/webviews/abstractWebview.ts
+++ b/src/webviews/abstractWebview.ts
@@ -19,6 +19,7 @@ import { isAction, isAlertable, isPMFSubmitAction } from '../ipc/messaging';
 import { iconSet, Resources } from '../resources';
 import { OnlineInfoEvent } from '../util/online';
 import { UIWebsocket } from '../ws';
+import { CommonActionType } from '../lib/ipc/fromUI/common';
 
 // ReactWebview is an interface that can be used to deal with webview objects when you don't know their generic typings.
 export interface ReactWebview extends Disposable {
@@ -206,6 +207,13 @@ export abstract class AbstractReactWebview implements ReactWebview {
                 }
             }
         }
+
+        // Special case for analytics which are normally routed through the new IPC system
+        if (a?.type === CommonActionType.SendAnalytics && a?.errorInfo) {
+            Container.analyticsApi.fireUIErrorEvent(a.errorInfo);
+            return true;
+        }
+
         return false;
     }
 
--- a/src/webviews/components/issue/JiraIssuePage.tsx
+++ b/src/webviews/components/issue/JiraIssuePage.tsx
@@ -47,6 +47,8 @@ import VotesForm from './VotesForm';
 import WatchesForm from './WatchesForm';
 import WorklogForm from './WorklogForm';
 import Worklogs from './Worklogs';
+import { AtlascodeErrorBoundary } from 'src/react/atlascode/common/ErrorBoundary';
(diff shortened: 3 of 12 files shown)


webpack.extension.dev.js
```diff
@@ -77,6 +77,7 @@ module.exports = [
                 'process.env.ATLASCODE_FX3_ENVIRONMENT': JSON.stringify(process.env.ATLASCODE_FX3_ENVIRONMENT),
                 'process.env.ATLASCODE_FX3_TARGET_APP': JSON.stringify(process.env.ATLASCODE_FX3_TARGET_APP),
                 'process.env.ATLASCODE_FX3_TIMEOUT': JSON.stringify(process.env.ATLASCODE_FX3_TIMEOUT),
+                'process.env.CI': JSON.stringify(process.env.CI),
```
sdzh-atlassian, line 80: Apparently some transitive dependencies have special behaviour depending on `$CI`. We probably will too, eventually :D

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -910,36 +912,43 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
...
-                            );
-                        }
-                        return (
-                            <div style={{ maxWidth: '1200px', margin: '20px auto 0 auto' }}>
-                                <Grid layout="fluid">
-                                    <GridColumn medium={8}>
+                <AtlascodeErrorBoundary
+                    context={{ view: AnalyticsView.JiraIssuePage }}
+                    postMessageFunc={(e) => {
+                        this.postMessage(e); /* just {this.postMessage} doesn't work */
```
sdzh-atlassian, line 918: Apparently just putting `this.postMessage` here doesn't work. I'm guessing it's because postMessage is not a callback, and not memo-ized properly? I'd really rather not mess with these pages more than absolutely necessary :D

src/analyticsTypes.ts
```diff
@@ -19,9 +19,13 @@ export enum AnalyticsView {
     CreateBitbucketIssuePage = 'createBitbucketIssue',
     CreatePullRequestPage = 'createPullRequest',
 
+    CreateJiraIssuePage = 'createJiraIssue',
+    JiraIssuePage = 'jiraIssue',
+
     PipelineSummaryPage = 'pipelineSummary',
 
     StartWorkPage = 'startWork',
+    OldStartWorkPage = 'oldStartWork',
```
bwieger-atlassian-com, line 28: Do we like this naming convention? 

Eg: should we be starting or ending with the "old" word?

src/webviews/abstractWebview.ts
```diff
@@ -206,6 +207,13 @@ export abstract class AbstractReactWebview implements ReactWebview {
                 }
             }
         }
+
+        // Special case for analytics which are normally routed through the new IPC system
+        if (a?.type === CommonActionType.SendAnalytics && a?.errorInfo) {
```
bwieger-atlassian-com, line 212: I'm assuming the "type" is added in this PR. Continue to review

bwieger-atlassian-com (commented):
Couple of questions

src/analyticsTypes.ts
sdzh-atlassian: Oh, forgot I had it here lol :) As per the conversation in the other PR - let's fix this in a follow-up ;)

src/webviews/abstractWebview.ts
sdzh-atlassian: It actually won't, I had to look into this :) The reason for this is the (unbelievably frustrating) difference in implementation for the JIRA pages and all the other ones

The older (JIRA) pages use [this definition](https://github.com/atlassian/atlascode/blob/main/src/ipc/messaging.ts#L26-L29) for events from `src/ipc`:
```js
{
    action: "SOME_STRING",
    ...payload
}
```


The newer (other) pages use [this definition](https://github.com/atlassian/atlascode/blob/main/src/lib/ipc/fromUI/common.ts#L19) in `src/lib/ibc` (!!?? what a nice trick to see if we're paying attention, eh?)
```js
{
    type: "SOME_STRING",    // from ReducerAction<K,V>
    ...payload,
}
```

For this new event, I didn't want to have 2 parallel implementations - and just added handling for a _new-style_ event in the _old-style_ handler - hence the weird out-of-place check ;D

bwieger-atlassian-com (approved):

sdzh-atlassian:
Rebased and fixed conflicts, merging

PR #62, opened 2024-12-28
[AXON-42] fix: show correct JIRA issues under "My X issues"

sdzh-atlassian:
### What is this?

A long overdue fix for a very annoying problem.

Before, when several JIRA instances shared a credential, client from one of them would be reused for everything, leading to incorrect results. Now, we properly create clients for each JIRA instance independently

| Before | After |
|--|--|
| image | image |

### How was this tested?

By doing a before and after 😉

### DIFF
--- a/src/atlclients/clientManager.ts
+++ b/src/atlclients/clientManager.ts
@@ -250,7 +250,7 @@ export class ClientManager implements Disposable {
     }
 
     private keyForSite(site: DetailedSiteInfo): string {
-        return site.credentialId;
+        return `${site.credentialId} - ${site.baseApiUrl}`;
     }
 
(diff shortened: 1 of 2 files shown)


src/atlclients/clientManager.ts
```diff
@@ -250,7 +250,7 @@ export class ClientManager implements Disposable {
     }
 
     private keyForSite(site: DetailedSiteInfo): string {
-        return site.credentialId;
+        return `${site.credentialId} - ${site.baseApiUrl}`;
```
bwieger-atlassian-com, line 253: Well that was easy 😆

bwieger-atlassian-com (approved):

PR #63, opened 2025-01-02
[AXON-43] chore: fix CI warnings about release actions

sdzh-atlassian:
### What is this?

Small PR to fix these warnings that appear in our builds:

image

Apparently the official github action for releases is no longer maintained, so this makes the repo use the highest-rated proposed alternative - [softprops/action-gh-release](https://github.com/softprops/action-gh-release)

It's almost a drop-in replacement, with an added benefit of combining the upload of artifacts into one action

### How was this tested?

Test release from this branch using the new action
image

Also, made sure that the build fails if the uploaded artifact isn't found:
image

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -89,22 +89,14 @@ jobs:
 
       - name: Create Release
         id: create_release
-        uses: actions/create-release@v1
+        uses: softprops/action-gh-release@v2
         env:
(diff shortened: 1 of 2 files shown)


bwieger-atlassian-com (approved):
Clean!

sdzh-atlassian:
Hopefully this doesn't break the actual nightlies ;D

PR #65, opened 2025-01-06
[AXON-32] allow for testing e2e in non-VPN environments

bwieger-atlassian-com:
Our prior implementation of e2e testing relied on jira-dev.com

This was a flawed starting point, because jira-dev.com is only accessible while on the Atlassian VPN, which this GitHub repo is not. 

Idea: Use a site that is still free to use, but not protected by a VPN. 

Solution: Make a site in the free-tier of JIRA usage on atlassian.net 

Name: axontest2025.atlassian.net 

- Make an exception for handling this so we aren't forced to use OAuth. 

How was this tested? 
- I test the VPN issue with jira-dev.com by trying to visit it on a personal device without VPN access. Resulted in a 403 
- I tested this locally to confirm the e2e test can run while on VPN 
- I am raising this PR to test if the GitHub actions pipeline will allow this to work. (May the odds be ever in my favor)

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -371,7 +371,7 @@ export class Container {
             return;
         }
         const authInfo = {
-            username: 'axon-test@polli.tlp.usersinbuckets.com',
+            username: 'axontest2025@gmail.com',
             password: process.env.ATLASCODE_TEST_USER_API_TOKEN,
             user: {
                 id: '',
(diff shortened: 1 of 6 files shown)


src/container.ts
```diff
@@ -371,7 +371,7 @@ export class Container {
             return;
         }
         const authInfo = {
-            username: 'axon-test@polli.tlp.usersinbuckets.com',
+            username: 'axontest2025@gmail.com',
```
sdzh-atlassian, line 374: NIT: we didn't do this last time, but should these values perhaps be a constant somewhere?

Especially given that they are used across several files

sdzh-atlassian (approved):
Approved! 👍

sdzh-atlassian:
image
Nice!

sdzh-atlassian (approved):
ship iiiiit 🚢

PR #66, opened 2025-01-07
[AXON-49] fix: prevent JIRA page crash on Reporter edit

sdzh-atlassian:
### What is this?

Small rabbit hole started from chasing an easy win 🐰 

We got a user report that editing the `Reporter` field on the JIRA issue page causes the whole page to crash. 

This PR fixes the error banner used there - which was very ungracefully failing to render.

This PR ALSO addresses the root cause of the error that needed the banner in the first place - by temporarily replacing the `/rest/api/2/user/recommend` API with `/rest/api/2/user/search` for that field

#### Before

https://github.com/user-attachments/assets/19ec081a-a2b3-4492-9582-1ce18a9e58de

#### After the ErrorBanner Fix

https://github.com/user-attachments/assets/e50391ac-0dd6-4ef7-b129-3df568efb0a4

#### Final State

https://github.com/user-attachments/assets/95d4fd61-fd24-423b-bf2e-99ec38e32ded

### How was this tested

Rather painfully 😢

Built the extension, used a project where editing `Reporter` is allowed, edited the field a bunch and made sure the search works (albeit with worse UX)

The E2E test for this is notably missing in the PR 😉

### DIFF
--- a/src/webviews/components/ErrorBanner.tsx
+++ b/src/webviews/components/ErrorBanner.tsx
@@ -69,20 +69,16 @@ export default class ErrorBanner extends React.Component<
 
         const title: string = this.state.errorDetails.title ? this.state.errorDetails.title : 'Something went wrong';
         return (
-            <SectionMessage
-                appearance="warning"
-                title={title}
-                actions={[
-                    {
-                        text: 'Dismiss',
-                        onClick: () => {
-                            this.setState({ errorDetails: undefined });
-                            this.props.onDismissError();
-                        },
-                    },
-                ]}
-            >
+            <SectionMessage appearance="warning" title={title}>
                 <div>{errorMarkup}</div>
+                <button
+                    onClick={() => {
+                        this.setState({ errorDetails: undefined });
+                        this.props.onDismissError();
+                    }}
+                >
+                    Dismiss
+                </button>
             </SectionMessage>
         );
     }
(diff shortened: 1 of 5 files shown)


src/webviews/components/ErrorBanner.tsx
```diff
@@ -69,20 +69,16 @@ export default class ErrorBanner extends React.Component<
...
-                        onClick: () => {
-                            this.setState({ errorDetails: undefined });
-                            this.props.onDismissError();
-                        },
-                    },
-                ]}
-            >
+            <SectionMessage appearance="warning" title={title}>
                 <div>{errorMarkup}</div>
+                <button
```
sdzh-atlassian, line 74: All other `SectionMessage`-s elsewhere in the code have children, and no `actions` prop at all

src/webviews/components/issue/AbstractIssueEditorPage.tsx
```diff
@@ -284,9 +284,21 @@ export abstract class AbstractIssueEditorPage<
         });
     };
 
+    /**
+     * AXON-49 This is a workaround for the suggested autocomplete URL
+     * (.../user/recommend/...) not being accessible
+     */
+    protected fixAutocompleteUrl(url: string): string {
```
sdzh-atlassian, line 291: Not proud of it, but it works

bwieger-atlassian-com (approved):
LGTM, but clearly missing e2e test.

PR #81, opened 2025-01-23
AXON-80  - Fix Blank JIRA bug due to attachment variable mishandling 

bwieger-atlassian-com:
[JIRA](https://softwareteams.atlassian.net/browse/AXON-80)

[Repro Video: 
](https://www.loom.com/share/2a8845fb08ee42d9be9850e4c8b0302c)

[Root Cause Video
](https://www.loom.com/share/a6f377870c794c609cdc5783904ce37c)

[Solution Testing  
](https://www.loom.com/share/9e39749b9343461ca9de5e06e848e9a7)

Splitting out the testing because we don't have a pattern for unit testing with the "Container" dependency that is required here. https://softwareteams.atlassian.net/browse/AXON-88


cabella-dot (approved):
LGTM!

PR #92, opened 2025-02-04
Revert to vscode engine compatibility >=1.77.0

bwieger-atlassian-com:
Project IDX looks for extensions that are compatible with the engine that it is running on. If the extension claims it runs on a higher version, it won't show up to be installed. 

To fix this, we are lowering our compatiblity requirement to match what we are actually compatible with. 

Note: we could go lower, we just haven't tested if we can. 1.77.0 is from March 2023

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "theme": "light"
     },
     "engines": {
-        "vscode": "^1.93.0"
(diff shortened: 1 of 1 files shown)


sdzh-atlassian (approved):
SHIP IIIIIT

PR #93, opened 2025-02-05
[AXON 95] Fix file attachment to jira tickets

marcomura:
The root of the issue was the inability to retrieve the absolute path of the picked file from the file picker, which probably it's done for security reasons, and the subsequent attempt to read the file from a path that we don't have.

Originally, the file was read from the controller, which receives the file info from the view via post message.

After the fix, the file is read by the view before the post message, and the controller receives the file content along with the file metadata.

Tested:
- Jira cloud, attaching files to existing jira ticket
- Jira DC, attaching files to existing jira ticket
- Jira DC, attaching files during a jira ticket creation


src/util/files.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function readFilesContentAsync(files: File[]): Promise<FileWithContent[]> {
+    const promise = new Promise<any>((resolve) => {
+        let doneCount = 0;
+        for (let i = 0; i < files.length; ++i) {
+            const index = i;
+            const reader = new FileReader();
+            reader.onloadend = (event) => {
+                (files[index] as FileWithContent).fileContent = arrayBufferToBase64(reader.result as ArrayBuffer);
+                if (++doneCount === files.length) {
```
sdzh-atlassian, line 16: Haven't seen `++` used in a while, this takes me back :D

sdzh-atlassian (commented):
Would you mind adding tests for the new `util` files? Otherwise, looks good to me :)

src/util/files.ts
marcomura: I'll look into UTs for the `util` files tomorrow :) thanks

src/util/bufferTransformers.ts
```diff
@@ -0,0 +1,21 @@
+export function arrayBufferToBase64(buffer: ArrayBuffer | null | undefined): string | undefined {
```
bwieger-atlassian-com, line 1: Have you considered using `npm install base64-arraybuffer`

https://github.com/niklasvh/base64-arraybuffer

marcomura: To be honest I haven't, I usually avoid taking dependencies for small things, and it was frowned upon on my previous job :)
What's the recommendation here?

marcomura: I've added the package

src/util/files.ts
marcomura: Reverted that since I'm now using a package to handle the conversion.
The new `util/file` remains untested as I don't know how to mock a File object.

sdzh-atlassian (commented):
ship it! 🚢🇮🇹

sdzh-atlassian (dismissed):

src/util/files.ts
marcomura: Jumped into a call, it looks like this code isn't well testable because of jest and the lack of FileReader in node.
Skipping for now

src/util/files.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface FileWithContent extends File {
+    /** base64-encoded file content */
+    fileContent: string | undefined;
+}
+
+export function readFilesContentAsync(files: File[]): Promise<FileWithContent[]> {
+    const promise = new Promise<any>((resolve) => {
+        let doneCount = 0;
+        for (let i = 0; i < files.length; ++i) {
+            const index = i;
```
bwieger-atlassian-com, line 12: I believe you can parrallelize this. 

Just ask GitHub Copilot to do it. If not, no worries.

src/webviews/components/issue/JiraIssuePage.tsx
```diff
@@ -384,23 +386,32 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
...
-                path: file.path,
-            };
-        });
-        this.postMessage({
-            action: 'addAttachments',
-            site: this.state.siteDetails,
-            issueKey: this.state.key,
-            files: serFiles,
+    handleAddAttachments = (files: File[]) => {
+        if (this.attachingInProgress) {
```
bwieger-atlassian-com, line 390: Interesting. What does the user behavior on this look like? 

What happens if I try to attach to files, one after the other?

bwieger-atlassian-com (dismissed):
Some performance enhancements. 

Happy to ship "good" and then make it "better"

src/util/files.ts
marcomura: not sure if I got it right, but reading N file contents is already parallelized.
FileReader executes the callback `onloadend` when the read is done.

src/webviews/components/issue/JiraIssuePage.tsx
marcomura: Good question. I believe the code is much faster than any user interaction, so this case shouldn't pose a problem, but I'll run some tests to verify it. 
I also forgot to put a guard on CreateIssuePage as well, and I'll fix it before merging.
Thanks!

marcomura: Attaching one file after the other works well :)

src/util/files.ts
bwieger-atlassian-com: I'm looking at this for loop: 

```
for (let i = 0; i < files.length; ++i) {
            const index = i;
```

bwieger-atlassian-com: It looks like I/O is embedded in the for-loop. So the total run time would be 

`i * io_time` 

But if we parallelize, it could just be: 

`max(io_time_i)`

marcomura: no it's not, `readAsArrayBuffer` is an async call that returns immediately, and the content is read asynchronously when the `onloadend` call back is invoked.
all the reads are happening in parallel, so the time can be already calculated as max(io_time_i)

bwieger-atlassian-com (approved):

marcomura:
> Some performance enhancements.
> 
> Happy to ship "good" and then make it "better"

I believe it's already "best" 😆

Issue #77, opened 2025-01-16
Error authenticating with Bitbucket 404


aparker-cityworks:
I'm trying to connect to Bitbucket datacenter. It is version v8.19.13.
I try to authenticate and I get an error
`[2025-01-16 00:09:49:192]   Error: Error authenticating with Bitbucket: AxiosError: Request failed with status code 404
[2025-01-16 00:09:49:193]  Error: Authentication error: Error authenticating with Bitbucket: AxiosError: Request failed with status code 404`

**To Reproduce**
1. In Atlassian Setting click Bitbucket tab
2. Click Add Custom Bitbucket site
3. Enter my base url https://bitbucket.mycompany.tools
4. Enter my Personal Access Token
5. Click Save site
6. Error

**Expected behavior**
I would expect it to authenticate successfully.

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (3.2.2) 
 - Visual Studio Code Version (1.96.3) 
 - Is this a local or cloud environment? (Local) 
 - If Local, OS (macOS)

bwieger-atlassian-com:
@aparker-cityworks, can you confirm that you are on VPN when trying to authenticate? 

I am assuming your company protects its stuff behind a VPN.

aparker-cityworks:
@bwieger-atlassian-com No, I am not on VPN. I can access Bitbucket in a browser without a VPN. Authentication is done through Okta in a browser.

I am trying to use a Personal Access Token to login in VS Code.
I have successfully used the same Personal Access Token to authenticate using the Gitkraken application, so I know it's possible.

aparker-cityworks:
@bwieger-atlassian-com , I had a person from our cloud tools team look at this. Here is his response:

> looks like a plugin issue so far 
Logs from the extension, sending in a contenttype of undefined for the first request, and the second URL does not seem to be valid - we might have to compare the version of Bitbucket they tested and our version.

> [2025-01-17 00:42:44:662] Error while creating agent  Error: ENOENT: no such file or directory, open 'bitbucket.mycompany.tools'
[2025-01-17 00:42:44:664] ---------------------------------------------------------------------- 
[...]

PR #100, opened 2025-02-06
AXON-111 revert changes that made special chars not work for BB DC

bwieger-atlassian-com:
Video of solution working: 

https://github.com/user-attachments/assets/0b8bab9e-8136-45a0-9e95-f14b1cfedacf

Issue: https://github.com/atlassian/atlascode/issues/77

Community pointing out the issue: https://github.com/atlassian/atlascode/pull/41/files 

Solution: 
1. Unfortunately, unlike before the credential object didn't contain the username. That why in the past PR, the `/rest/api/latest/build/capabilities` API was introduced to get the username info 
2. The header returns this in URI encoding eg: bwieger%40atlassian.com 
3. So that needs to be convert to standard text eg: bwieger@atlassian.com
4. And then the special characters can be replaced with the prior regex.

### DIFF
--- a/src/atlclients/loginManager.ts
+++ b/src/atlclients/loginManager.ts
@@ -172,13 +172,13 @@ export class LoginManager {
                     ...getAgent(site),
                 });
                 const slugRegex = /[\[\:\/\?#@\!\$&'\(\)\*\+,;\=%\\\[\]]/gi;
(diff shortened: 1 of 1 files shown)


marcomura (approved):
it looks lgtm

PR #104, opened 2025-02-10
[Axon 97] Fix image rendering in Jira Cloud tickets' description and comments

marcomura:
To render descriptions and comments of Jira tickets, we use a "rendered html" field returned by the server.

When we request the jira ticket information, Jira DC responds with a rendered html that contains images referenced with an absolute URL, which work fine.
Jira Cloud instead responds with a rendered html that contains images referenced with a relative URL, which cannot be resolved because the `<base>` url is set to some vscode react stuff.

The fix is to pre-process the "rendered html" before displaying it, fixing the relative URLs to absolute.

Tested:
- Jira Cloud, image in description and comments
- Jira DC, image in description and comments

Before:
image
image

After:
image
image

### DIFF
--- a/src/util/html.test.ts
+++ b/src/util/html.test.ts
@@ -1,72 +1,75 @@
 import * as html from './html';
 
 describe('html util', () => {
-
     describe('fix relative URLs in html', () => {
-
         it('replaceRelativeURLsWithAbsolute correctly replaces the relative URL', () => {
-            const baseUrl = "https://www.domain.com/path1/path2/path3";
-            const htmlText = "<p><label>/imgs/test.png</label><img src=\"/imgs/test.png\" /></p>";
+            const baseUrl = 'https://www.domain.com/path1/path2/path3';
+            const htmlText = '<p><label>/imgs/test.png</label><img src="/imgs/test.png" /></p>';
 
-            const expectedHtml = "<p><label>/imgs/test.png</label><img src=\"https://www.domain.com/path1/path2/imgs/test.png\" /></p>";
+            const expectedHtml =
+                '<p><label>/imgs/test.png</label><img src="https://www.domain.com/path1/path2/imgs/test.png" /></p>';
 
             const fixedHtml = html.replaceRelativeURLsWithAbsolute(htmlText, baseUrl);
             expect(fixedHtml).toBe(expectedHtml);
         });
 
         it('replaceRelativeURLsWithAbsolute correctly replaces all relative URLs', () => {
-            const baseUrl = "https://www.domain.com/path1/path2/path3";
-            const htmlText = "<p>" +
-                "<label>/imgs/test1.png</label><img src=\"/imgs/test1.png\" />" +
-                "<label>/imgs/test2.png</label><img src=\"/imgs/test2.png\" />" +
+            const baseUrl = 'https://www.domain.com/path1/path2/path3';
+            const htmlText =
+                '<p>' +
+                '<label>/imgs/test1.png</label><img src="/imgs/test1.png" />' +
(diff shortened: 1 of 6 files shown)


sdzh-atlassian:
If you don't mind, small request for the future PRs like this - would you mind throwing in a pair of before-and-after screenshots?
I typically try to do something like [this here](https://github.com/atlassian/atlascode/pull/79) - doesn't have to be a gif though

Basically, for this kind of visual issues, it really helps to see the before and after IMO 😄

bwieger-atlassian-com:
Yes to what @sdzh-atlassian says. 

Any sort of visual artifact will do (Loom, GitHub Video upload, screenshots, gifs...)

src/util/html.test.ts
```diff
@@ -0,0 +1,75 @@
+import * as html from './html';
+
+describe('html util', () => {
```
bwieger-atlassian-com, line 3: GREAT! Love the tests!!!! 💯

sdzh-atlassian (dismissed):
Ship it! ⛵👍

marcomura:
> If you don't mind, small request for the future PRs like this - would you mind throwing in a pair of before-and-after screenshots? I typically try to do something like [this here](https://github.com/atlassian/atlascode/pull/79) - doesn't have to be a gif though
> 
> Basically, for this kind of visual issues, it really helps to see the before and after IMO 😄

Makes sense!
I've added a couple of screenshots to the description :)

Thanks!

src/util/html.ts
```diff
@@ -0,0 +1,13 @@
+const regex1 = / src="\/[^"]+/g;
+const regex2 = / src='\/[^']+/g;
+
+export function replaceRelativeURLsWithAbsolute(renderedHtml: string, baseApiUrl: string): string | undefined {
+    if (!renderedHtml || !baseApiUrl) {
+        return renderedHtml;
+    }
+
+    // substring(7) because we need to get the relative url without the first '/'
```
bwieger-atlassian-com, line 9: Can you explain this a bit more? 

7 feels like a magic and possibly brittle number.

marcomura: The regex is searching for anything starting with ` src="/` which is 7 chars long, and we need to get the URL without including its first `/`, so anything after those 7 characters.
Do you want me to update the comment?

bwieger-atlassian-com (dismissed):
Mostly LGTM. A little concerned about the magic number, but not enough to block this fix from going out. 

That said, this change feels risky enough to warrant a [release test plan](https://hello.atlassian.net/wiki/spaces/AIDO/pages/4777772018/Release+Test+Plan) for the next release. Just to ensure we didn't break an image loading else where.

src/util/html.ts
bwieger-atlassian-com: Got it! Thanks

bwieger-atlassian-com: No need to update.

marcomura:
> Mostly LGTM. A little concerned about the magic number, but not enough to block this fix from going out.
> 
> That said, this change feels risky enough to warrant a [release test plan](https://hello.atlassian.net/wiki/spaces/AIDO/pages/4777772018/Release+Test+Plan) for the next release. Just to ensure we didn't break an image loading else where.

Sorry, I thought that the existing comment was enough to explain why `7`.
I'm going to update it to make even clearer.

I'll prepare the release test plan before merging.
Thanks.

sdzh-atlassian (approved):
🚢 🇮🇹 #️⃣ 2️⃣

bwieger-atlassian-com (approved):

PR #106, opened 2025-02-12
AXON-113: Fix Bitbucket DC 8.0 and later bugs with 404s 

bwieger-atlassian-com:
Many of the bitbucket APIs that we used around tasks are now deprecated in favor of the "comments" APIs. 

Using the "tasks" API on version of BitBucket DC 8.0 and later result in 404s. 

Since we don't have a clear pattern for fetching the version of the backend we are attaching to, the solution here uses a try catch pattern looking for 404s. 

I chose to prioritize the new APIs because I !assume! that most of our future (and perhaps current) users are on 8.0 and later. 

8.0 was release around May 2022 (3 years old) 

**[BEFORE](https://www.loom.com/share/de22e2c02842496ea52b057e27e4b0f6)**

**[AFTER](https://www.loom.com/share/171712ebfeeb47c195160efb5a3f4d49?sid=278594d6-e6e6-4d3a-8100-96646b9165cf)**

### DIFF
--- a/src/bitbucket/bitbucket-server/pullRequests.ts
+++ b/src/bitbucket/bitbucket-server/pullRequests.ts
@@ -246,7 +246,7 @@ export class ServerPullRequestApi implements PullRequestApi {
     private async postTask_v0(site: BitbucketSite, prId: string, content: string, commentId?: string) {
         const bbApi = await clientForSite(site);
         const repo = await bbApi.repositories.get(site);
-        const { data } = await this.client.post(`/rest/api/1.0/tasks`, {
+        const { data } = await this.client.post(`/rest/api/latest/tasks`, {
             anchor: {
                 id: commentId,
                 type: 'COMMENT',
@@ -258,6 +258,7 @@ export class ServerPullRequestApi implements PullRequestApi {
             state: 'OPEN',
             text: content,
         });
+
         return this.convertDataToTask(data, site);
     }
 
@@ -935,6 +936,7 @@ export class ServerPullRequestApi implements PullRequestApi {
         const { data } = await this.client.get(
             `/rest/api/1.0/projects/${ownerSlug}/repos/${repoSlug}/pull-requests/${prId}/tasks/count`,
         );
+
         return data;
     }
 
--- a/src/bitbucket/bitbucket-server/pullRequests.ts
+++ b/src/bitbucket/bitbucket-server/pullRequests.ts
@@ -182,9 +182,22 @@ export class ServerPullRequestApi implements PullRequestApi {
     async getTasks(pr: PullRequest): Promise<Task[]> {
         const { ownerSlug, repoSlug } = pr.site;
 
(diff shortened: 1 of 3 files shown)


bwieger-atlassian-com:
Still some bugs around this, but massive improvement

src/bitbucket/bitbucket-server/pullRequests.ts
```diff
@@ -215,10 +258,60 @@ export class ServerPullRequestApi implements PullRequestApi {
...
+                severity: 'BLOCKER',
+                text: content,
+                properties: {},
+            },
+        );
+        return this.convertDataToTask_v8(data.data, site);
     }
+
     async editTask(site: BitbucketSite, prId: string, task: Task): Promise<Task> {
+        try {
```
sdzh-atlassian, line 283: I'm very on board with this approach of trying things with fallback - but just out of curiosity - would we be able to figure out the right version beforehand, and store it somewhere like in the `DetailedSiteInfo`?

sdzh-atlassian (approved):
Couple small things caught my eye, but overall I'd say ship it! :)

src/bitbucket/bitbucket-server/pullRequests.ts
marcomura: Similar to my comment before, definitely it would be great

src/bitbucket/bitbucket-server/pullRequests.ts
```diff
@@ -215,10 +258,60 @@ export class ServerPullRequestApi implements PullRequestApi {
...
+            } else {
+                throw error;
+            }
+        }
+    }
+
+    private async editTask_v8(site: BitbucketSite, prId: string, task: Task) {
+        const projectKey = site.ownerSlug;
+        const repositorySlug = site.repoSlug;
+        const pullRequestId = prId;
```
marcomura, line 297: this looks unnecessary :) you can just rename the function arg

marcomura: same in other places

marcomura:
The file `PullRequests.ts` would be best paired with from some UTs

src/bitbucket/bitbucket-server/pullRequests.ts
```diff
@@ -180,8 +180,38 @@ export class ServerPullRequestApi implements PullRequestApi {
     }
 
     async getTasks(pr: PullRequest): Promise<Task[]> {
+        try {
+            return this.getTasks_v8(pr);
```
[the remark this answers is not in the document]
bwieger-atlassian-com: Good idea!

src/bitbucket/bitbucket-server/pullRequests.ts
bwieger-atlassian-com: Agreed! Focused on delivering value to the customer quickly, and then we can follow up on optimization.

PR #110, opened 2025-02-13
AXON-113: First test for server pull requests

bwieger-atlassian-com:
A first pass at make a test for this file. 

Let me know if you like the structure. Then I'll go do some more.

### DIFF
--- a/src/bitbucket/bitbucket-server/pullRequests.test.ts
+++ b/src/bitbucket/bitbucket-server/pullRequests.test.ts
@@ -1,27 +1,34 @@
 import { ServerPullRequestApi } from './pullRequests';
 import { BitbucketSite } from '../model';
-import { IHTTPClient } from '../httpClient';
+
+const mockGet = jest.fn();
+const mockPost = jest.fn();
+const mockPut = jest.fn();
+const mockDelete = jest.fn();
+jest.mock('../httpClient', () => ({
+    HTTPClient: jest.fn().mockImplementation(() => ({
+        get: mockGet,
+        post: mockPost,
+        put: mockPut,
+        delete: mockDelete,
+    })),
+}));
+import { HTTPClient } from '../httpClient';
+import { AxiosResponse } from 'axios';
 
 describe('ServerPullRequestApi', () => {
     let api: ServerPullRequestApi;
-    let mockClient: jest.Mocked<IHTTPClient>;
+    let mockClient: HTTPClient = new HTTPClient('', '', '', async (errJson: AxiosResponse) => Error('some error'));
     let site: BitbucketSite;
 
     beforeEach(() => {
-        mockClient = {
-            get: jest.fn(),
-            post: jest.fn(),
-            put: jest.fn(),
-            delete: jest.fn(),
-            getRaw: jest.fn(),
-        } as unknown as jest.Mocked<IHTTPClient>;
-
+        jest.clearAllMocks();
         api = new ServerPullRequestApi(mockClient);
         site = { ownerSlug: 'owner', repoSlug: 'repo', details: { userId: 'user' } } as BitbucketSite;
     });
 
     it('should get with a v8 if no 404s', async () => {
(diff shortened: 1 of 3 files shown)


src/bitbucket/bitbucket-server/pullRequests.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+        await api.get(site, 'PR-1');
+
+        expect(mockGet).lastCalledWith(
+            '/rest/api/1.0/projects/owner/repos/repo/pull-requests/PR-1/blocker-comments?count=true',
+        );
+    });
+});
+
+const getTaskCountDataV8 = {
```
bwieger-atlassian-com, line 55: Test data. Could be placed in a supporting file.

src/bitbucket/bitbucket-server/pullRequests.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    data: {
+        OPEN: 5,
+    },
+    headers: {},
+};
+
+const getPullRequestData = {
```
sdzh-atlassian, line 62: Do we actually need the realistic PR response here? I'd normally suggest that we minimize stuff like this and only keep the necessary fields for tests - but I'm not sure if it might be needed for other tests or something

src/bitbucket/bitbucket-server/pullRequests.test.ts
sdzh-atlassian: For something this small I'd say it might be better to keep it here, but no strong opinion :)

src/bitbucket/bitbucket-server/pullRequests.test.ts
```diff
@@ -0,0 +1,206 @@
...
+        api = new ServerPullRequestApi(mockClient);
+        site = { ownerSlug: 'owner', repoSlug: 'repo', details: { userId: 'user' } } as BitbucketSite;
+    });
+
+    it('should get with a v8 if no 404s', async () => {
+        mockGet.mockImplementation((url, queryParams?) => {
+            if (
+                url.includes('/rest/api/1.0/projects/owner/repos/repo/pull-requests/PR-1/blocker-comments?count=true')
+            ) {
+                return Promise.resolve(getTaskCountDataV8);
```
sdzh-atlassian, line 35: NIT: (and I might be wrong) any reason not to just do `return getTaskCountDataV8`, `return {data: getPullRequestData }` and `throw new Error('Not Found')` here? Basically, do we actually need the `Promise.*` if the closure is already `async`?

src/bitbucket/bitbucket-server/pullRequests.test.ts
sdzh-atlassian: also, SUPER NITPICK (sorry): `getTaskCountDataV8` and `getPullRequestData` sound like function names to me, so I was actually a bit confused when reading the usage above :P 

Again, this is like new level of pettiness as far as nitpicks go, feel free to ignore 😁

sdzh-atlassian (approved):
Feel free to ignore the nitpicks and ship it! ⛵

src/bitbucket/bitbucket-server/pullRequests.test.ts
cabella-dot: Maybe we can create a mock values file to hold all of these to keep the tests readable

src/bitbucket/bitbucket-server/pullRequests.test.ts
cabella-dot: I agree, I think it would be better if they were named like constants! i.e. `TASK_COUNT_DATA_V8`

src/bitbucket/bitbucket-server/pullRequests.test.ts
cabella-dot: ^^ only if we plan on adding more cases to this test

PR #125, opened 2025-02-15
[AXON-110] BitBucket DC PAT Auth Bug

bwieger-atlassian-com:
Users were reporting issues authenticating with PAT on BitBucket DC. 

This fixes it. 

I looked into adding unit tests for this file. There are many things to mock out here to make this efficient. 

I'm punting on unit tests here.
Instead, I will look at UI testing.

| Before | After |
|--------|--------|
| Screenshot 2025-02-14 at 4 04 22 PM | image |

### DIFF
--- a/src/atlclients/clientManager.ts
+++ b/src/atlclients/clientManager.ts
@@ -134,7 +134,7 @@ export class ClientManager implements Disposable {
             } else {
                 result = {
                     repositories:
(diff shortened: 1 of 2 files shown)


sdzh-atlassian (approved):
Oh yeah _that_ file would definitely be really hard to write tests for, lol :)
Ship it! ⛵

marcomura (approved):

PR #130, opened 2025-02-18
[AXON-117] Fix 'Create issue webview' stuck on the loader

marcomura:
The issue repros when the first server is an unreachable server, like a shut down Jira DC.

The page was stuck, and looping on a circle of error/refresh, because of a wrong condition on the react page.
The fix is in `CreateIssuePage.tsx`, the rest of the changes are minor random improvements.

### DIFF
--- a/src/webviews/createIssueWebview.ts
+++ b/src/webviews/createIssueWebview.ts
@@ -6,6 +6,7 @@ import {
     isScreensForSite,
     isSetIssueType,
 } from '../ipc/issueActions';
+import { WebViewID } from '../lib/ipc/models/common';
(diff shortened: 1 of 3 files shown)


src/webviews/createIssueWebview.ts
```diff
@@ -266,16 +267,14 @@ export class CreateIssueWebview
                 };
             }
 
-            if (this._screenData) {
```
sdzh-atlassian, line 269: Is this actually guaranteed to be set somehow?

sdzh-atlassian (approved):

PR #131, opened 2025-02-19
Adding a check if commit message already starts with the issue key

sambenskin:
Adding a check if commit message already starts with the issue key


sambenskin:
I tried to fill in the CLA but got a 500 error so I've emailed atlassian to investigate.

PR #135, opened 2025-02-20
[AXON-46] Add remote auth flow using atlascode-backend

sdzh-atlassian:
### What is this?

The next PR in the chain to enable a smooth 3LO-from-anywhere experience in the extension ;)

With this, we can now use an external service to bounce authentication from - in this PR it's referred to as the `JiraRemote` OAuth strategy

Short recap of the changes:
 * Fork the existing monolithic OAuth process, split it in 2 parts (start via user action / finish via deeplink)
 * Add a button (completely invisible to everyone, FF pending) to trigger the new oauth
 * Add an ability to configure OAuth client from the environment on the fly, for dev purposes
 * Move OAuth strategy data into their own file, now with a bunch of docs
 * Implement URIHandler path for `/auth`
 * A whooooole lot of IPC/messaging boilerplate 😞 
 
What's notably missing:
 * OAuth attempt/session management
 * The way to trigger new auth outside of dev environment is effectively hidden
 * PKCE for the new Jira auth
 * Bitbucket remote auth

There are also some known issues - for instance, the credentials seemingly get bundled between the strategies and somehow invalidate themselves - rather inconsistently, at that 😞

image

I'm leaving the follow-up to follow-up PRs. Also, some of the issues might be addressed by #79  

### How was this tested

* Unit tests, especially the new one on `Strategy`/`StrategyData` to prevent regressions
* E2E tested the new flow
* E2E tested the old flow

### DIFF
--- a/src/atlclients/strategyData.ts
+++ b/src/atlclients/strategyData.ts
@@ -1,0 +1,188 @@
+import { OAuthProvider, Product, ProductBitbucket, ProductJira } from './authInfo';
+
+export type StrategyProps = {
+    /**  What does this strategy refer to? Essentially, strategy ID */
+    provider: OAuthProvider;
+    /** Is this for JIRA or Bitbucket? */
+    product: Product;
+    /**
+     * Client ID of the OAuth app. Docs:
+     *  - Jira: https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/
+     *  - Bitbucket: https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/
+     */
+    clientID: string;
+    /**
+     * Base URL for the initial authorization request
+     */
+    authorizationURL: string;
+    /**
+     * Base URL for getting access tokens
+     */
+    tokenURL: string;
+    /**
+     * Base URL for getting user profile information
+     */
+    profileURL: string;
+    /**
+     * The callback URL this strategy will supply to the OAuth provider.
+     * Must be exactly as configured in the OAuth app.
+     */
+    callbackURL: string;
+    /**
+     * Base URL for the API calls. Used to
+     */
+    apiURL: string;
+
+    /**
+     * Present only in non-PKCE environments
+     */
+    clientSecret?: string;
+
+    /**
+     * Jira-only
+     * Base URL for getting accessible resources
+     */
+    accessibleResourcesURL?: string;
+    /**
+     * Jira-only
+     * Scope for the OAuth request, as seen in a classic Jira platform REST API authorization URL
+     * (e.g. "manage:jira-project offline_access")
+     * See developer.atlassian.com 3LO docs
+     */
+    scope?: string;
+    /**
+     * Jira-only
+     * Additional parameters for the OAuth request
+     * See developer.atlassian.com 3LO docs
+     */
+    authParams?: {
+        /**
+         * The audience parameter for the OAuth request
+         */
+        audience: string;
+        /**
+         * The prompt parameter for the OAuth request
+         */
+        prompt: string;
+    };
+
+    /**
+     * Bitbucket-only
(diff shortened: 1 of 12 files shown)


src/atlclients/strategyData.ts
```diff
@@ -0,0 +1,188 @@
...
+            audience: 'api.stg.atlassian.com',
+            prompt: 'consent',
+        },
+    };
+
+    static readonly BitbucketProd: StrategyProps = {
+        provider: OAuthProvider.BitbucketCloud,
+        product: ProductBitbucket,
+        clientID: '3hasX42a7Ugka2FJja',
+        clientSecret: 'st7a4WtBYVh7L2mZMU8V5ehDtvQcWs9S',
```
cabella-dot, line 167: are these client secrets ok to be made public

src/lib/webview/controller/config/configWebviewController.ts
```diff
@@ -138,6 +145,14 @@ export class ConfigWebviewController implements WebviewController<SectionChangeM
                 this._analytics.fireAuthenticateButtonEvent(id, msg.siteInfo, isCloud);
                 break;
             }
+            case ConfigActionType.RemoteLogin: {
+                const uri = vscode.Uri.parse('vscode://atlassian.atlascode/auth');
```
cabella-dot, line 149: /nit have this as a constant?

src/atlclients/strategyData.ts
sdzh-atlassian: Hahah, excellent catch, that is the million dollar question! :)

The short answer is `no`. However, they've been in the repo for 5 years 💀
In this PR, I'm just moving them from `strategy.ts` to `strategyData.ts`

There's _sort_ of a plan for how to address this going forward - but it would be quite out of scope for this specific PR

src/lib/webview/controller/config/configWebviewController.ts
sdzh-atlassian: Good call! :)

sdzh-atlassian: Fixed in the next push!

src/atlclients/strategyData.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const getRemoteAuthConfig = () => {
+    const DEFAULT_REMOTE_AUTH_CONFIG = {
+        clientID: '',
+        clientSecret: '',
+        callbackURL: '',
+    };
+
+    try {
+        const config = JSON.parse(process.env.ATLASCODE_REMOTE_AUTH_CONFIG || '') as RemoteAuthConfig;
```
marcomura, line 94: JSON.parse fails for `''`, better to fallback to `'{}'`

src/lib/ipc/fromUI/config.ts
```diff
@@ -22,6 +23,7 @@ export enum ConfigActionType {
 
 export type ConfigAction =
     | ReducerAction<ConfigActionType.Login, LoginAuthAction>
+    | ReducerAction<ConfigActionType.RemoteLogin, RemoteLoginAuthAction>
```
marcomura, line 26: I believe you can just write `ReducerAction<ConfigActionType.RemoteLogin>` and avoid creating an empty interface

marcomura (dismissed):
two minor comments, LTGM

src/lib/ipc/fromUI/config.ts
sdzh-atlassian: You're right! Thanks, TIL it defaults to `void` :) Removed the empty interface

src/atlclients/strategyData.ts
sdzh-atlassian: Thanks! It's in a `try`/`catch` block anyway - but changed to a better default

cabella-dot (approved):

Issue #90, opened 2025-02-04
Add settings to not push instantly to remote


SteveH-K:
Hello,
I would like to suggest a new settings for the atlascode.
When I use the "Start work on Jira issue" feature, I would like a setting to disable immediate pushing to the remote.
Background: This immediately triggers a pipeline in our Git Server which is not needed as there is no difference from the source branch (usually the main branch).
Could you please consider such a setting?
See this line: https://github.com/atlassian/atlascode/blob/b58c55477001bffbe83c40513cc19a156510e204/src/webview/startwork/vscStartWorkActionApi.ts#L77
Best regads,
Steve

bwieger-atlassian-com:
Thanks for raising this @SteveH-K ! 

Added to the backlog -> https://softwareteams.atlassian.net/browse/AXON-109 (Atlassian private link) 

It might take some time to get to this. We have some pretty big challenges with authentication and UI.

SteveH-K:
Hello @bwieger-atlassian-com,
thank you for your quick response and for adding this suggestion to your backlog!
I'm happy to see this new feature in the plugin.
I wish you luck in tackling your current challenges!
Best regards,
Steve

marcomura:
Hello @SteveH-K ,

Thanks for your suggestion!
This change has been implemented, and it's now released under the new version 3.4.8.

Preview:
![Image](https://github.com/user-attachments/assets/cdbcb89a-a307-493e-8897-5f53ec7124db)

This new toggle is defaulted to "true" for the moment, but we are considering remembering the last setting so it doesn't have to be adjusted every time.

Thanks
-- Marco

SteveH-K:
Hello Marco,
thank you for the quick implementation of this feature.
However, a version 3.4.9 of the Atlascode plugin was released without your changes from version 3.4.8. See: https://github.com/atlassian/atlascode/compare/v3.4.9...main
So unfortunately I can't see your changes right now. But I suspect that with the next version push your changes will also be published.

bwieger-atlassian-com:
@SteveH-K Sorry for the revert in functionality in 3.4.9

There was a breaking bug that needed to be reversed.

We should have this fixed later today and have 3.4.10 ready for you.

SteveH-K:
Hello @marcomura,
no problem at all! I updated to 3.4.10 today and your new toggle works like a charm. Thank you very much!

ian-h-chamberlain:
[...]

PR #137, opened 2025-02-20
[AXON-109] Expose toggle switch to push branch to remote (Start work)

marcomura:
With this PR, the Start Work view exposes a toggle switch to control if the new branch should be automatically pushed to remote or not.
The toggle switch is defaulted to `true`, to keep parity with today's behavior.

The status of the new toggle is being instrumented within the `workStarted` event.

This addresses the following open issues:
[Add settings to not push instantly to remote](https://github.com/atlassian/atlascode/issues/90) #90
[Create branch as a local branch during "Start work"](https://bitbucket.org/atlassianlabs/atlascode/issues/538/create-branch-as-a-local-branch-during)

Here how it looks like:
image

### DIFF
--- a/src/analytics.test.ts
+++ b/src/analytics.test.ts
@@ -20,11 +20,8 @@ function setProcessPlatform(platform: NodeJS.Platform) {
 describe('viewScreenEvent', () => {
     const originalPlatform = process.platform;
 
-    beforeAll(() => {
-        setProcessPlatform('win32');
-    });
-
     beforeEach(() => {
+        setProcessPlatform('win32');
         mockedData.getFirstAAID_value = true;
     });
 
--- a/src/analytics.test.ts
+++ b/src/analytics.test.ts
@@ -1,0 +1,106 @@
+import { viewScreenEvent } from './analytics';
+
+interface MockedData {
+    getFirstAAID_value?: boolean;
+}
+
+const mockedData: MockedData = {};
+
+jest.mock('./container', () => ({
+    Container: { siteManager: { getFirstAAID: () => mockedData.getFirstAAID_value } },
+}));
+
+function setProcessPlatform(platform: NodeJS.Platform) {
+    Object.defineProperty(process, 'platform', {
+        value: platform,
+        writable: false,
+    });
+}
+
+describe('viewScreenEvent', () => {
+    const originalPlatform = process.platform;
+
+    beforeAll(() => {
+        setProcessPlatform('win32');
+    });
+
+    beforeEach(() => {
+        mockedData.getFirstAAID_value = true;
+    });
+
+    afterAll(() => {
+        setProcessPlatform(originalPlatform);
+    });
+
+    it('should create a screen event with the correct screen name', async () => {
+        const screenName = 'testScreen';
+        const event = await viewScreenEvent(screenName);
+        expect(event.name).toEqual(screenName);
+        expect(event.screenEvent.attributes).toBeUndefined();
+    });
+
+    it('should exclude from activity if screen name is atlascodeWelcomeScreen', async () => {
+        const screenName = 'atlascodeWelcomeScreen';
+        const event = await viewScreenEvent(screenName);
+        expect(event.screenEvent.attributes.excludeFromActivity).toBeTruthy();
+    });
+
+    it('should include site information if provided (cloud)', async () => {
+        const screenName = 'testScreen';
+        const site: any = {
+            id: 'siteId',
+            product: { name: 'Jira', key: 'jira' },
+            isCloud: true,
+        };
+        const event = await viewScreenEvent(screenName, site);
+        expect(event.screenEvent.attributes.instanceType).toEqual('cloud');
+        expect(event.screenEvent.attributes.hostProduct).toEqual('Jira');
+    });
+
+    it('should include site information if provided (server)', async () => {
+        const screenName = 'testScreen';
+        const site: any = {
+            id: 'siteId',
+            product: { name: 'Jira', key: 'jira' },
+            isCloud: false,
+        };
+        const event = await viewScreenEvent(screenName, site);
+        expect(event.screenEvent.attributes.instanceType).toEqual('server');
+        expect(event.screenEvent.attributes.hostProduct).toEqual('Jira');
+    });
+
+    it('should include product information if provided', async () => {
+        const screenName = 'testScreen';
+        const product = { name: 'Bitbucket', key: 'bitbucket' };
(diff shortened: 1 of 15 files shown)


sdzh-atlassian (dismissed):
Ship it!⛵

Still can't believe such a straightforward change is 10 files of boilerplate because of IPC lol

marcomura:
> Ship it!⛵
> 
> Still can't believe such a straightforward change is 10 files of boilerplate because of IPC lol

It's also pretty weird that we have 3 different copies of `createOrCheckoutBranch` when at most we should need 2 (one for Jira and one for BitBucket)...

src/analytics.test.ts
```diff
@@ -0,0 +1,106 @@
+import { viewScreenEvent } from './analytics';
```
sdzh-atlassian, line 1: Testing against `analytics.ts` is kind of a heroic effort, nice :)

src/analytics.ts
```diff
@@ -21,7 +21,7 @@ export const Registry = {
 };
 
 class AnalyticsPlatform {
-    private static nodeJsPlatformMapping = {
+    private static nodeJsPlatformMapping: Record<string, any> = {
```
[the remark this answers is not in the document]
marcomura: I'm not sure if our telemetry library supports instrumenting different value types... e.g., the switch I'm instrumenting is a boolean.

src/analytics.ts
```diff
@@ -107,8 +107,13 @@ export async function issueCommentEvent(site: DetailedSiteInfo): Promise<TrackEv
...
 }
 
-export async function issueWorkStartedEvent(site: DetailedSiteInfo): Promise<TrackEvent> {
-    return instanceTrackEvent(site, 'workStarted', 'issue');
+export async function issueWorkStartedEvent(
+    site: DetailedSiteInfo,
+    pushBranchToRemoteChecked: boolean,
+): Promise<TrackEvent> {
+    const attributesObject = instanceType({}, site);
+    attributesObject.attributes.pushBranchToRemoteChecked = pushBranchToRemoteChecked;
```
sdzh-atlassian, line 115: Oh man, I can't name the amount of hours I lost on the fact that `attributes` here expects an object like `{ attributes: {a: 'b'}}` and not `{a: 'b'}` 💀

marcomura: yeah it's totally unintuitive... maybe one day we'll refactor it :D

src/analytics.ts
```diff
@@ -657,33 +662,27 @@ function tenantOrNull<T>(e: Object, tenantId?: string): T {
...
+        eventProps.attributes.instanceType = site.isCloud ? 'cloud' : 'server';
+        eventProps.attributes.hostProduct = site.product.name;
     }
 
-    if (attrs) {
-        newObj['attributes'] = { ...newObj['attributes'], ...attrs };
-    }
-
-    return newObj;
+    return eventProps;
```
sdzh-atlassian, line 681: Sorry, this is kind of sus to me :( Why modify the parameter instead of returning a new object?

src/analytics.ts
```diff
@@ -657,33 +662,27 @@ function tenantOrNull<T>(e: Object, tenantId?: string): T {
...
-    if (newObj['attributes']) {
-        newObj['attributes'] = { ...newObj['attributes'], ...{ excludeFromActivity: true } };
-    } else {
-        Object.assign(newObj, { attributes: { excludeFromActivity: true } });
-    }
-    return newObj;
+function excludeFromActivity(eventProps: Record<string, any>): Record<string, any> {
+    eventProps.attributes = eventProps.attributes || {};
+    eventProps.attributes.excludeFromActivity = true;
+    return eventProps;
```
sdzh-atlassian, line 687: Same here

src/react/atlascode/startwork/StartWorkPage.tsx
```diff
@@ -95,7 +95,7 @@ const StartWorkPage: React.FunctionComponent = () => {
 
     const [transitionIssueEnabled, setTransitionIssueEnabled] = useState(true);
     const [branchSetupEnabled, setbranchSetupEnabled] = useState(true);
-    const [pushBranchEnabled, setPushBranchEnabled] = useState(false);
+    const [pushBranchEnabled, setPushBranchEnabled] = useState(true);
```
sdzh-atlassian, line 98: I thought we were going to default this to `false`, were we not?

src/analytics.ts
marcomura: `newObj` was never a new object, I changed this part because it was confusing.
But it was creating a new object for `attributes` which seemed expensive and unnecessary - I did change that.

src/react/atlascode/startwork/StartWorkPage.tsx
marcomura: no, changed my mind.
I'm keeping parity with today's behavior to avoid taking people by surprise, and in a future PR I'm going to make the toggle value sticky (e.g., if you turn it off, next time it'll be off by default)

src/analytics.ts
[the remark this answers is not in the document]
sdzh-atlassian: Pretty sure it does! But isn't this specifically for the `platform` value?

src/analytics.ts
sdzh-atlassian: Oh my god you're right! Sorry, I didn't look to the left, and was assuming there's an `Object.assign` or something

This kind of _really_ doesn't sit right with me ngl - but it's definitely an upgrade to the original 🤔 I was going to suggest that we at least don't return anything then - but looks like we're using the return value in a bunch of places, too :/

src/analytics.ts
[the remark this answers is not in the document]
marcomura: oh sorry, for that one you are correct. Let me fix it :) thanks

src/analytics.ts
sdzh-atlassian: Would you mind documenting the side effect on the function?

src/react/atlascode/startwork/StartWorkPage.tsx
sdzh-atlassian: Oh, my bad - I didn't realize we push the branch by default today, I thought it's the other way around. Good call with the lack of disruption!

src/analytics.ts
sdzh-atlassian: Had a chat about this, agreed that the implementation could use an improvement, but decided to limit the refactoring here to keep the scope small :)

sdzh-atlassian (approved):

bwieger-atlassian-com (approved):

PR #143, opened 2025-02-22
Axon 144 experiment auth UI flow statsig set up

cabella-dot:
What: 
* added support for Statsig experiment values to be sent to FE for experimentation
* set up AA experiment implementation in Onboarding flow

Demo loom: 

https://www.loom.com/share/7888dc16753649fc9d6cc16e6e36f2da

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -71,7 +71,6 @@ import { AtlascodeUriHandler } from './uriHandler';
 import { CheckoutHelper } from './bitbucket/interfaces';
 import { ProductJira } from './atlclients/authInfo';
 import { ATLASCODE_TEST_USER_EMAIL, ATLASCODE_TEST_HOST } from './constants';
-import { ExperimentGateValues, FeatureGates } from './util/featureFlags/features';
 
 const isDebuggingRegex = /^--(debug|inspect)\b(-brk\b|(?!-))=?/;
 const ConfigTargetKey = 'configurationTarget';
@@ -205,13 +204,9 @@ export class Container {
                 analyticsAnonymousId: env.machineId,
             },
             eventBuilder: new EventBuilder(),
-        })
-            .then(() => {
-                this.initializeUriHandler(context, this._analyticsApi, this._bitbucketHelper);
-            })
-            .then(() => {
-                this.checkGates();
-            });
+        }).then(() => {
+            this.initializeUriHandler(context, this._analyticsApi, this._bitbucketHelper);
+        });
     }
 
     static getAnalyticsEnable(): boolean {
@@ -224,7 +219,7 @@ export class Container {
         analyticsApi: VSCAnalyticsApi,
         bitbucketHelper: CheckoutHelper,
     ) {
-        if (FeatureFlagClient.checkGate(Features.EnableNewUriHandler)) {
+        if (FeatureFlagClient.featureGates[Features.EnableNewUriHandler]) {
             console.log('Using new URI handler');
             context.subscriptions.push(AtlascodeUriHandler.create(analyticsApi, bitbucketHelper));
         } else {
@@ -389,10 +384,6 @@ export class Container {
         await Container.loginManager.userInitiatedServerLogin(site, authInfo);
     }
 
-    static async checkGates() {
-        this._featureGates = await FeatureFlagClient.evaluateFeatures();
-        this._experimentGates = await FeatureFlagClient.evaluateExperiments();
-    }
     private static _pullRequestDetailsWebviewFactory: MultiWebview<any, PullRequestDetailsAction>;
     static get pullRequestDetailsWebviewFactory() {
         return this._pullRequestDetailsWebviewFactory;
@@ -536,15 +527,5 @@ export class Container {
     private static _bitbucketHelper: CheckoutHelper;
     static get bitbucketHelper() {
         return this._bitbucketHelper;
-    }
-
-    private static _featureGates: FeatureGates;
-    static get featureGates() {
-        return this._featureGates;
-    }
-
-    private static _experimentGates: ExperimentGateValues;
-    static get experimentGates() {
-        return this._experimentGates;
     }
 }
--- a/src/container.ts
+++ b/src/container.ts
@@ -71,6 +71,7 @@ import { AtlascodeUriHandler } from './uriHandler';
 import { CheckoutHelper } from './bitbucket/interfaces';
 import { ProductJira } from './atlclients/authInfo';
 import { ATLASCODE_TEST_USER_EMAIL, ATLASCODE_TEST_HOST } from './constants';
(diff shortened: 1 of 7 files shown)


src/container.ts
```diff
@@ -204,9 +205,13 @@ export class Container {
                 analyticsAnonymousId: env.machineId,
             },
             eventBuilder: new EventBuilder(),
-        }).then(() => {
-            this.initializeUriHandler(context, this._analyticsApi, this._bitbucketHelper);
-        });
+        })
+            .then(() => {
```
marcomura, line 209: nitpick: merge the two then()

src/util/featureFlags/client.ts
```diff
@@ -66,6 +66,25 @@ export class FeatureFlagClient {
         return gateValue;
     }
 
+    public static checkExperimentValue(experiment: string): any {
+        let gateValue: any;
+        const experimentGate = ExperimentGates[experiment];
```
marcomura, line 71: you should null check `experimentGate` here, or this function will throw an exception if the experiment isn't not known

src/util/featureFlags/features.ts
```diff
@@ -2,3 +2,21 @@ export enum Features {
...
+    [Experiments.NewAuthUI]: {
+        gate: 'atlascode_new_auth_ui',
+        parameter: 'isEnabled',
+        defaultValue: false,
+    },
+};
+
+type ExperimentGate = { [key: string]: { gate: string; parameter: string; defaultValue: any } };
+
+export type FeatureGates = { [key: string]: boolean };
```
marcomura, line 20: nitpick: there is a built in type for this: `Record<K, V>`. This type will be `Record<string, boolean>`.
Same for other part of the code.

src/util/featureFlags/features.ts
```diff
@@ -2,3 +2,21 @@ export enum Features {
...
+
+export const ExperimentGates: ExperimentGate = {
+    [Experiments.NewAuthUI]: {
+        gate: 'atlascode_new_auth_ui',
+        parameter: 'isEnabled',
+        defaultValue: false,
+    },
+};
+
+type ExperimentGate = { [key: string]: { gate: string; parameter: string; defaultValue: any } };
```
marcomura, line 18: nitpick: is it worth having an interface for it?

src/container.ts
```diff
@@ -384,6 +389,10 @@ export class Container {
         await Container.loginManager.userInitiatedServerLogin(site, authInfo);
     }
 
+    static async checkGates() {
```
bwieger-atlassian-com, line 392: For my learning: 

Does this mean that we have solved the initialization problem with the FF Client?

src/util/featureFlags/client.ts
```diff
@@ -68,18 +81,22 @@ export class FeatureFlagClient {
...
-                `FeatureGates: FeatureGates is not initialized. Returning default value: ${experimentGate.defaultValue}`,
-            );
-        } else {
-            console.log(ExperimentGates[experiment]);
-            gateValue = FeatureGates.getExperimentValue(
-                experimentGate.gate,
-                experimentGate.parameter,
-                experimentGate.defaultValue,
-            );
+        try {
```
marcomura, line 84: mmh no, using try/catch instead of a null check is not a good pattern. It obfuscates what it's there for, and performance wise exceptions handling is worse than a simple if.

In general, try catch should be used to catch exceptions that functions you are calling may throw, for which you don't have much control.
When you have control of the code, you should simply check the correctness of the values.

In this case it should be as simple as
```
const experimentGate = ExperimentGates[experiment];
if (!experimentGate) {
  return undefined;
}
...
```

src/util/featureFlags/client.ts
```diff
@@ -68,18 +81,22 @@ export class FeatureFlagClient {
...
+            const experimentGate = ExperimentGates[experiment];
+            if (FeatureGates === null) {
+                console.warn(
+                    `FeatureGates: FeatureGates is not initialized. Returning default value: ${experimentGate.defaultValue}`,
```
marcomura, line 88: I just noticed this warning... it doesn't seem correct. If `FeatureGates == null` it's going to return `gateValue` which is uninitialized (undefined), no?

src/container.ts
cabella-dot: Yes, now it will not fetch from client every time page is rendered. We will store it during a session

src/util/featureFlags/client.ts
```diff
@@ -66,6 +79,28 @@ export class FeatureFlagClient {
...
     }
 
+    public static checkExperimentValue(experiment: string): any {
+        let gateValue: any;
+        const experimentGate = ExperimentGates[experiment];
+        if (!experimentGate) {
+            return undefined;
+        }
+        if (FeatureGates === null) {
+            console.warn(
```
bwieger-atlassian-com, line 89: Not sure if console works here, you might want to use Logger.

src/util/featureFlags/client.ts
```diff
@@ -66,6 +79,28 @@ export class FeatureFlagClient {
         return gateValue;
     }
 
+    public static checkExperimentValue(experiment: string): any {
+        let gateValue: any;
+        const experimentGate = ExperimentGates[experiment];
+        if (!experimentGate) {
+            return undefined;
```
bwieger-atlassian-com, line 86: Do we need a log here?

src/util/featureFlags/client.ts
```diff
@@ -66,6 +79,28 @@ export class FeatureFlagClient {
...
+            );
+            gateValue = experimentGate.defaultValue;
+        } else {
+            gateValue = FeatureGates.getExperimentValue(
+                experimentGate.gate,
+                experimentGate.parameter,
+                experimentGate.defaultValue,
+            );
+        }
+        console.log(`ExperimentGateValue: ${experiment} -> ${gateValue}`);
```
bwieger-atlassian-com, line 100: same here about console vs Logger

bwieger-atlassian-com (approved):
I left some comments, but none are blockers to this PR.

src/util/featureFlags/client.ts
cabella-dot: yes, I will implement

src/util/featureFlags/client.ts
cabella-dot: this is just for debugging purposes, I don't believe we want to be logging this every time FG gets checked? If we do, I can implement

cabella-dot:
@bwieger-atlassian-com the way the logger is currently implemented, it cannot be used within the FE portion of the codebase bc the dependency 'fs' (filesystem) throws an error if needed in ./src/webviews/ directory. Bc `FeatureFlagClient` is used in this directory, the extension fails. I will create a follow up task to find a better way to log errors or warnings in the FE.

PR #144, opened 2025-02-24
[AXON-161] Update default issue creation keywords

sdzh-atlassian:
### What is this?

Context - currently, we have a pretty strict syntax that requires `TODO:` for the `Create JIRA Issue` annotation to pop up:
image

Update to the default keywords for ticket creation, per suggestion in Slack:
 * Removed trailing `:`
 * Added a bunch of extra options

Please note that this is just the default values. The docs aren't very clear - but I'm pretty sure this wouldn't affect users who already made changes to these settings

### How was this tested?

It was edited right in GH UI, so it kind of wasn't :) TBD in next nightly or manual build

### DIFF
--- a/package.json
+++ b/package.json
@@ -1178,14 +1178,22 @@
                     },
                     "default": [
                         "TODO",
+                        "todo",
+                        "Todo",
+                        "ToDo",
                         "BUG",
+                        "bug",
+                        "Bug",
                         "FIXME",
+                        "fixme",
+                        "Fixme",
+                        "FixMe",
                         "ISSUE",
-                        "NOTE",
-                        "OPTIMIZE",
+                        "issue",
+                        "Issue",
                         "HACK",
-                        "XXX",
-                        "FIX"
+                        "hack",
+                        "Hack"
                     ],
                     "description": "Strings that will trigger the Jira issue create Code Lens",
                     "scope": "window"
--- a/package.json
+++ b/package.json
@@ -1177,10 +1177,15 @@
                         "type": "string"
                     },
                     "default": [
-                        "TODO:",
-                        "BUG:",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1177,10 +1177,15 @@
...
                     "default": [
-                        "TODO:",
-                        "BUG:",
-                        "FIXME:",
-                        "ISSUE:"
+                        "TODO",
+                        "BUG",
+                        "FIXME",
+                        "ISSUE",
+                        "NOTE",
```
marcomura, line 1184: How did you come up with this new entries?
Personally, I often use "NOTE" to emphasize a comment in the code that doesn't require any follow up.
And "XXX" ? 🤣

package.json
```diff
@@ -1177,10 +1177,15 @@
                         "type": "string"
                     },
                     "default": [
-                        "TODO:",
-                        "BUG:",
-                        "FIXME:",
-                        "ISSUE:"
+                        "TODO",
```
bwieger-atlassian-com, line 1180: LGTM, and can we add the case variants here: 

1. CASE 
2. case 
3. Case

package.json
marcomura, line 1184: I think "NOTE" would trigger too many false positives, and I would add "FIXME"

package.json
bwieger-atlassian-com: And let's constrain this to follow what GitHub does until we have more data 

image

bwieger-atlassian-com (changes requested):

bwieger-atlassian-com (approved):

cabella-dot (approved):
looks good to me

marcomura (approved):

PR #146, opened 2025-02-24
[AXON-149] (WIP) re-enable auth UI for bitbucket in RDE

sdzh-atlassian:
### What is this?

> ⚠️ Note: found some issues while submitting the PR - hence draft
>   * ~~url parsing might be borked after rebase, somehow~~ - tested, works now
>   * ~~would be good to disable the `save` button in BB cloud~~ - maybe in a follow-up
> * Please note that this specifically DOESN'T address BB cloud auth, and targets BasicAuth/PAT on BB DC/server instead

This PR enables the custom auth flow for Bitbucket in RDE, in the same way as JIRA - except using BB [app password](https://support.atlassian.com/bitbucket-cloud/docs/app-passwords/) turned out to be a _really_ poor substitute for JIRA API token, so that's not available ;)

The rationale here is to enable folks who use BB Server/DC to authenticate remotely from RDEs via PAT/basic auth

| | Old | New |
|-|-|-|
| Local<br/> (no change) | image | image |
| Remote | image | image |

UI after this:
image

### How was this tested?

 * Via dev-container, managed to authenticate against an available instance of BB DC

### DIFF
--- a/src/react/atlascode/config/auth/SiteAuthenticator.tsx
+++ b/src/react/atlascode/config/auth/SiteAuthenticator.tsx
@@ -1,10 +1,9 @@
-import { Box, Button, Grid, Typography } from '@material-ui/core';
+import { Box, Button, Grid } from '@material-ui/core';
 import React, { memo, useCallback, useContext } from 'react';
 
 import { AuthDialogControllerContext } from './useAuthDialog';
 import { CloudAuthButton } from './CloudAuthButton';
-import DomainIcon from '@material-ui/icons/Domain';
-import { Product, ProductJira } from '../../../../atlclients/authInfo';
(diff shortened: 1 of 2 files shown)


livingtonds:
Hi!What about plans for merging this pr?)

sdzh-atlassian:
Hi folks!

> url parsing might be borked after rebase, somehow
Sorry about the delay with this, the error above was caused by an incident we had last week - but back then we weren't sure and put this on pause :)

I'm updating and re-submitting it for approval

sdzh-atlassian:
Just tested the urtl parser - everything seems to be in order.

Please note that this PR _DOES NOT_ address the cloud auth not working in RDE - but should at least allow basic/PAT auth against BB DC/server

PR #154, opened 2025-02-26
Axon 137 sidebar UI new sidebar view for jq ls

cabella-dot:
WIP

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -72,6 +72,7 @@ import { CheckoutHelper } from './bitbucket/interfaces';
 import { ProductJira } from './atlclients/authInfo';
 import { ATLASCODE_TEST_USER_EMAIL, ATLASCODE_TEST_HOST } from './constants';
 import { CustomJQLViewProvider } from './views/jira/customJqlExplorer/customJqlViewProvider';
+import { Logger } from './logger';
 
 const isDebuggingRegex = /^--(debug|inspect)\b(-brk\b|(?!-))=?/;
 const ConfigTargetKey = 'configurationTarget';
@@ -185,7 +186,16 @@ export class Container {
 
         this._loginManager = new LoginManager(this._credentialManager, this._siteManager, this._analyticsClient);
         this._bitbucketHelper = new BitbucketCheckoutHelper(context.globalState);
-
+        FeatureFlagClient.initialize({
+            analyticsClient: this._analyticsClient,
+            identifiers: {
+                analyticsAnonymousId: env.machineId,
+            },
+            eventBuilder: new EventBuilder(),
+        }).then(() => {
+            this.initializeUriHandler(context, this._analyticsApi, this._bitbucketHelper);
+            this.initializeCustomJqlView(context);
+        });
         if (config.jira.explorer.enabled) {
             context.subscriptions.push((this._jiraExplorer = new JiraContext()));
         } else {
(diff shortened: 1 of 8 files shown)


cabella-dot:
TODO: tests

src/container.ts
```diff
@@ -220,13 +217,34 @@ export class Container {
...
         if (FeatureFlagClient.featureGates[Features.EnableNewUriHandler]) {
-            console.log('Using new URI handler');
+            Logger.debug('Using new URI handler');
             context.subscriptions.push(AtlascodeUriHandler.create(analyticsApi, bitbucketHelper));
         } else {
             context.subscriptions.push(new LegacyAtlascodeUriHandler(analyticsApi, bitbucketHelper));
         }
     }
+    static initializeNewSidebarView(context: ExtensionContext, config: IConfig) {
+        if (FeatureFlagClient.featureGates[Features.NewSidebarTreeView]) {
```
cabella-dot, line 227: add null check fir FeatureFlagClient.featuregates if failed to connect

src/util/featureFlags/features.ts
```diff
@@ -13,6 +15,11 @@ export const ExperimentGates: ExperimentGate = {
         parameter: 'isEnabled',
         defaultValue: false,
     },
+    [Experiments.AA]: {
```
marcomura, line 18: Can you do this in a separate PR?
Don't want to have this change hiding here in case we are searching for it in the future

src/container.ts
```diff
@@ -184,29 +187,23 @@ export class Container {
...
             analyticsClient: this._analyticsClient,
             identifiers: {
                 analyticsAnonymousId: env.machineId,
             },
             eventBuilder: new EventBuilder(),
-        }).then(() => {
-            this.initializeUriHandler(context, this._analyticsApi, this._bitbucketHelper);
-        });
+        })
+            .then(() => {
```
marcomura, line 197: how about removing then(), catch the error and log it, and launch all the initialization in a finally() ?
nothing should be conditional to the success or failure of the feature flag initializer because we have default values in case of failure

marcomura: the downside of the current implementation is that there are two different entrypoint for `initializeLegacySidebarView` and it's harder to follow the logic

marcomura (approved):
looks good! thanks!

marcomura (approved):

marcomura (dismissed):

sdzh-atlassian (dismissed):
Nice! Good to see some tests going into place, too! 🚢

bwieger-atlassian-com (changes requested):
- Enhancement / Bug: If I log out of all sites, I see JQL filters with no results. Instead I should see a call to action to sign in. 
- Bug: After logging out, and then logging back in, JQL filters go entirely blank.

cabella-dot:
@bwieger-atlassian-com 

https://www.loom.com/share/87d16c3fe69543ef9f14a8fcc957f0c8

bwieger-atlassian-com:
@cabella-dot Reviewed! LGTM 


Here's my review video: https://www.loom.com/share/7045725bcdfd4ffc8b5e78cbf7d677f6?sid=bb5ed99c-8a20-418d-9e5a-cd654025e69c 

It looks like you resolved one of the JQL bugs with this PR. Good work! 🚀 

There is still an outstanding pre-existing bug. But not an expectation to solve in this PR.

bwieger-atlassian-com (approved):

bwieger-atlassian-com (approved):

PR #164, opened 2025-03-03
Fix exceptions due to failures in FeatureFlagClient initialization

marcomura:
The code was throwing exceptions because when failing to retrieve the data from the server (e.g., missing `.env` file), the fields `FeatureFlagClient._featureGates` and `FeatureFlagClient._experimentValues` were left uninitialized.

### DIFF
--- a/src/util/featureFlags/client.ts
+++ b/src/util/featureFlags/client.ts
@@ -24,20 +24,27 @@ export class FeatureFlagClient {
     }
 
     public static async initialize(options: FeatureFlagClientOptions): Promise<void> {
-        const targetApp = process.env.ATLASCODE_FX3_TARGET_APP || '';
-        const environment =
-            (process.env.ATLASCODE_FX3_ENVIRONMENT as FeatureGateEnvironment) || FeatureGateEnvironment.Production;
+        const targetApp = process.env.ATLASCODE_FX3_TARGET_APP;
+        const environment = process.env.ATLASCODE_FX3_ENVIRONMENT as FeatureGateEnvironment;
+        const apiKey = process.env.ATLASCODE_FX3_API_KEY;
+        const timeout = process.env.ATLASCODE_FX3_TIMEOUT;
+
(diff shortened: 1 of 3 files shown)


src/util/featureFlags/client.ts
```diff
@@ -47,29 +54,27 @@ export class FeatureFlagClient {
...
-                this._experimentValues = await this.evaluateExperiments();
-            })
             .catch((err) => {
                 console.warn(`FeatureGates: Failed to initialize client. ${err}`);
-                console.warn('FeatureGates: Disabling feature flags');
                 this.eventBuilder
                     .featureFlagClientInitializationFailedEvent()
                     .then((e) => options.analyticsClient.sendTrackEvent(e));
+            })
+            .finally(async () => {
```
sdzh-atlassian, line 69: Possibly a dumb question - but why the `async` here?

marcomura: you are right, it's not supposed to be async. It's a leftover after removing the await.
Thanks for catching it!

sdzh-atlassian (approved):

src/util/featureFlags/client.ts
```diff
@@ -101,29 +104,25 @@ export class FeatureFlagClient {
...
-                    [feature]: await this.checkGate(feature),
-                };
-            }),
-        );
-
-        return featureFlags.reduce((acc, val) => ({ ...acc, ...val }), {});
+    private static evaluateFeatures(): FeatureGateValues {
+        const featureFlags = Object.values(Features).map(async (feature) => {
+            return {
+                // eslint-disable-next-line @typescript-eslint/await-thenable
```
cabella-dot, line 110: you can probably now delete this lint ignore if you are not awaiting

cabella-dot (approved):
LGTM

src/util/featureFlags/client.ts
marcomura: right, thanks!

cabella-dot (approved):

PR #165, opened 2025-03-04
[AXON-194] Allow easy override of feature flags and experiments

marcomura:
The override is done through environment variables.

It can be applied in environment machines via the `.env` file, or mocking the environment in tests.


.env.example
```diff
@@ -3,3 +3,8 @@ ATLASCODE_FX3_API_KEY="totally-real-api-key"
 ATLASCODE_FX3_ENVIRONMENT="production"
 ATLASCODE_FX3_TARGET_APP="atlascode_web"
 ATLASCODE_FX3_TIMEOUT="2000"
+
+ATLASCODE_FF_OVERRIDES="feature_flag=true"
+
+ATLASCODE_EXP_OVERRIDES_BOOL="exp1_name=true"
+ATLASCODE_EXP_OVERRIDES_STRING="exp2_name=something"
```
sdzh-atlassian, line 10: Missing newline 😛

.env.example
```diff
@@ -3,3 +3,8 @@ ATLASCODE_FX3_API_KEY="totally-real-api-key"
 ATLASCODE_FX3_ENVIRONMENT="production"
 ATLASCODE_FX3_TARGET_APP="atlascode_web"
 ATLASCODE_FX3_TIMEOUT="2000"
+
+ATLASCODE_FF_OVERRIDES="feature_flag=true"
```
sdzh-atlassian, line 7: I'd really appreciate a comment here about comma being the expected delimiter. Or, maybe, something like `f1=true,f2=false` in the example value?

webpack.react.webview.js
```diff
@@ -69,6 +69,9 @@ module.exports = {
             'process.env.ATLASCODE_FX3_TARGET_APP': JSON.stringify(process.env.ATLASCODE_FX3_TARGET_APP),
             'process.env.ATLASCODE_FX3_TIMEOUT': JSON.stringify(process.env.ATLASCODE_FX3_TIMEOUT),
             'process.env.ATLASCODE_TEST_USER_API_TOKEN': JSON.stringify(process.env.ATLASCODE_TEST_USER_API_TOKEN),
+            'process.env.ATLASCODE_FF_OVERRIDES': JSON.stringify(process.env.ATLASCODE_FF_OVERRIDES),
+            'process.env.ATLASCODE_EXP_OVERRIDES_BOOL': JSON.stringify(process.env.ATLASCODE_EXP_OVERRIDES_BOOL),
+            'process.env.ATLASCODE_EXP_OVERRIDES_STRING': JSON.stringify(process.env.ATLASCODE_EXP_OVERRIDES_STRING),
```
sdzh-atlassian, line 74: Random thought, maybe for one of those Friday engineering excellence things - we might be at a point where it could be really good to extract some common webpack logic :D Not in scope of this PR though

sdzh-atlassian: Or, I guess, we've been at that point for quite a while now and just failed to act xD My bad :D

sdzh-atlassian (approved):
Approved with some nits ;)

.env.example
marcomura: you are right, the syntax should be properly documented, thanks

webpack.react.webview.js
marcomura: yeah, it would be nice to refactor the webpack configs... I'm not even sure why we have so many

sdzh-atlassian (approved):

PR #171, opened 2025-03-05
Decoupling ChangedFiles & Conflicts API

sinhaviraj:
We found that for larger repos the conflicts API would always time out within 30 seconds and the UI wouldn't load. In order to fix that I have decoupled the changedFiles & conflicts API calls.

Now isConflicted is nomore the property on diffstats and is instead stored as an array of conflictedFiles in state. The rendering component would then compare this array with the fileName and if fileName exists in conflictedArray then mark this as a conflicted file. 

Loom video explaining the changes - https://www.loom.com/share/097ef30c0301409d95e0b6e12d9e5a63

Loom Video trying the changes on larger repos - https://www.loom.com/share/8681ec2bf2cf4026b3869afdff78c802

### DIFF
--- a/src/bitbucket/bitbucket-cloud/pullRequests.ts
+++ b/src/bitbucket/bitbucket-cloud/pullRequests.ts
@@ -217,25 +217,25 @@ export class CloudPullRequestApi implements PullRequestApi {
             Logger.error(error);
         }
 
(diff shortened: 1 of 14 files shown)


bwieger-atlassian-com:
Tested https://www.loom.com/share/a3e29250c0194f648dbfe78dd0e0b16f?sid=b189be1e-8e8a-4467-9542-0686c62118df

LGTM

bwieger-atlassian-com (approved):

PR #174, opened 2025-03-05
Axon 201 implement experiment flow split

cabella-dot:
Loom: https://www.loom.com/share/19e0e4b8bfc347d58404b0bb4c75cfa3

Due to the nature of this experiment (only first time installers will be put into cohorts):
* changed experiment fetching to be outside of initialization
* implemented exp flow in `extension.ts` 

Now, only users whose `previousVersion === undefined` will be placed into exp cohorts:
* A - control
* B - jump to settings page auth
* C - jump to legacy auth flow
* D - jump to new auth flow

### DIFF
--- a/src/react/atlascode/onboarding/OnboardingPage.tsx
+++ b/src/react/atlascode/onboarding/OnboardingPage.tsx
@@ -60,7 +60,6 @@ export const OnboardingPage: React.FunctionComponent = () => {
     const [state, controller] = useOnboardingController();
     const { authDialogController, authDialogOpen, authDialogProduct, authDialogEntry } = useAuthDialog();
     const [activeStep, setActiveStep] = React.useState(0);
-    // const [authExpEnabled, setAuthExpEnabled] = React.useState(false);
     const [useAuthUI, setUseAuthUI] = React.useState(false);
     const [jiraSignInText, setJiraSignInText] = useState('Sign In to Jira Cloud');
     const [bitbucketSignInText, setBitbucketSignInText] = useState('Sign In to Bitbucket Cloud');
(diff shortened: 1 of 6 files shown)


src/react/atlascode/onboarding/OnboardingPage.tsx
```diff
@@ -60,6 +60,7 @@ export const OnboardingPage: React.FunctionComponent = () => {
     const [state, controller] = useOnboardingController();
     const { authDialogController, authDialogOpen, authDialogProduct, authDialogEntry } = useAuthDialog();
     const [activeStep, setActiveStep] = React.useState(0);
+    // const [authExpEnabled, setAuthExpEnabled] = React.useState(false);
```
sdzh-atlassian, line 63: commented out code?

src/util/featureFlags/client.ts
```diff
@@ -153,6 +153,7 @@ export class FeatureFlagClient {
             );
         }
         console.log(`ExperimentGateValue: ${experiment} -> ${gateValue}`);
+        this._experimentValues[experiment] = gateValue;
```
sdzh-atlassian, line 156: Why are we doing this here? IMO side effects typically need a good reason, right?

src/react/atlascode/onboarding/OnboardingPage.tsx
cabella-dot: oops haha thanks for catching this 🥇

sdzh-atlassian (changes requested):
Left a couple small questions

src/util/featureFlags/client.ts
cabella-dot: This is so we can still send the experimentValues to the front end in file `singleViewFactory.ts`.

sdzh-atlassian: Sorry, I should have been more clear - why is this no longer done in bulk at initialization, like before?

cabella-dot: Because this gets fetched conditionally now. For example, now we only fetch the experiment when it is a new user. So only when they fetch that experiment is fetch should it be saved. If it was fetched and saved at init, people who are not new users will get put into cohorts and will give us wonky data in Statsig. We need to focus the fetching of experiments to the target audience

sdzh-atlassian: Ah! Well, TBH this generally feels a bit sus to me - but I don't really have a good alternative suggestion, so LFG :D

sdzh-atlassian (approved):

PR #184, opened 2025-03-07
AXON-189: adding status commands + new issueNode testing + put behind fg

cabella-dot:
### What Is This Change?

To follow new designs for the sidebar tree view, this PR:
* adds `statusCategory` (to do, in prog, done) tags onto the side of each IssueNode on hover
* when new tag is clicked => open issue
* added testing for issueNode

Loom: 
https://www.loom.com/share/721aeda9fef64617a99a476289f5f1e0?sid=e4b1450f-bfe9-4c98-9b47-b8fdc30790e2

### How Has This Been Tested?

Added UT for IssueNode in `issueNode.test.ts`

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,18 @@
     "contributes": {
         "commands": [
             {
+                "command": "atlascode.jira.todoIssue",
+                "title": "To Do"
+            },
+            {
+                "command": "atlascode.jira.inProgressIssue",
+                "title": "In Progress"
+            },
+            {
+                "command": "atlascode.jira.doneIssue",
+                "title": "Done"
+            },
+            {
                 "command": "atlascode.jira.createIssue",
                 "title": "Create Jira Issue",
                 "icon": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -98,6 +98,18 @@
     },
     "contributes": {
         "commands": [
+            {
+                "command": "atlascode.jira.todoIssue",
+                "title": "To Do"
```
sdzh-atlassian, line 103: probably should be one word, right?

package.json
```diff
@@ -642,24 +654,39 @@
...
+                },
+                {
+                    "command": "atlascode.jira.doneIssue",
+                    "when": "viewItem == doneJiraIssue && atlascode:isJiraAuthenticated",
+                    "group": "inline"
+                },
                 {
                     "command": "atlascode.jira.assignIssueToMe",
-                    "when": "viewItem == jiraIssue && atlascode:isJiraAuthenticated",
+                    "when": "viewItem =~ /(jiraIssue|todoJiraIssue|inProgressJiraIssue|doneJiraIssue)/ && atlascode:isJiraAuthenticated",
```
sdzh-atlassian, line 674: Maaaan, just as an observation - this stuff is really tricky to properly review IMO. Or debug, probably, right?

src/views/nodes/issueNode.ts
```diff
@@ -3,41 +3,68 @@ import * as vscode from 'vscode';
...
+        default:
+            return ISSUE_NODE_CONTEXT_VALUE;
+    }
 };
 
-const ISSUE_NODE_CONTEXT_VALUE = 'jiraIssue';
 export class IssueNode extends AbstractBaseNode {
     public issue: MinimalORIssueLink<DetailedSiteInfo>;
-
+    private _newPanelFG: boolean = false;
```
sdzh-atlassian, line 30: nit: might want to name this with boolean conventions, like `isUsingNewPanelFG`? Took me a double-take to realize this is a feature flag. Might be just me though

src/views/nodes/issueNode.ts
```diff
@@ -3,41 +3,68 @@ import * as vscode from 'vscode';
...
     }
 
     getTreeItem(): vscode.TreeItem {
-        const title = isMinimalIssue(this.issue) && this.issue.isEpic ? this.issue.epicName : this.issue.summary;
+        let title: string;
+        let description: string | undefined;
+        let contextValue: string;
+        const summary = isMinimalIssue(this.issue) && this.issue.isEpic ? this.issue.epicName : this.issue.summary;
+
+        if (this._newPanelFG) {
```
sdzh-atlassian, line 45: super nit and very debatable - I'd try to avoid having a bunch of `let`s unless absolutely necessary ;)
In this case, maybe by doing something like

```typescript
    const { title, description, contextValue } = this.prepareTreeItem(); // or some such
  // ...
  prepareTreeItem() {
    const summary = // ...
    return this._newPanelFG ? {
      // ...
    } : {
      // ...
    };
  }
```

Again, pure matter of preference, please feel free to ignore :D If anything, I'd appreciate feedback on if this particular preference is unreasonable :D

sdzh-atlassian (approved):
Added a couple nits - but generally looks good :D

package.json
cabella-dot: based on the designs, i think it should be 2: https://www.figma.com/design/oPqs5rKNZB2YeUDzqqv30z/Atlassian-VSCode-extension-M1?node-id=364-22738&p=f&m=dev

but I do have to make it capitalized!

package.json
cabella-dot: definitely.

Luckily, this is determined through `contextValue` of the treeItem so I an test the implementation, just not if the button _actually shows up

src/views/nodes/issueNode.ts
cabella-dot: will do!

src/views/nodes/issueNode.ts
cabella-dot: This makes sense, will fix!

src/util/featureFlags/client.ts
```diff
@@ -163,6 +163,19 @@ export class FeatureFlagClient {
         return featureFlags;
     }
 
+    static areFeaturesInitialized(): boolean {
+        try {
```
marcomura, line 167: let's not try/catch here, I've fixed the client to prevent `featureGates` to ever be undefined exactly to make checks like this safe :)

src/util/featureFlags/client.ts
```diff
@@ -163,6 +163,19 @@ export class FeatureFlagClient {
         return featureFlags;
     }
 
+    static areFeaturesInitialized(): boolean {
```
marcomura, line 166: what is this for?

src/views/nodes/issueNode.ts
```diff
@@ -3,41 +3,68 @@ import * as vscode from 'vscode';
...
 
-const ISSUE_NODE_CONTEXT_VALUE = 'jiraIssue';
 export class IssueNode extends AbstractBaseNode {
     public issue: MinimalORIssueLink<DetailedSiteInfo>;
-
+    private _newPanelFG: boolean = false;
     constructor(_issue: MinimalORIssueLink<DetailedSiteInfo>, parent: AbstractBaseNode | undefined) {
         super(parent);
         this.issue = _issue;
+        this._newPanelFG = FeatureFlagClient.areFeaturesInitialized()
```
marcomura, line 34: it's going to be a big issue if we don't initialize the feature in time for whatever comes next... we shouldn't have to check this

cabella-dot: good point, will remove this

marcomura (approved):

PR #190, opened 2025-03-10
AXON-136 [SidebarUI] Restyled sidebar view for Assigned Jira issues

marcomura:
### What Is This Change?

Implementation of the new "Assigned to me" side panel showing all the Jira issues assigned to me across all Jira sites.

image

Following items still needs to be done:
- polish integration with the new Custom JQL panel
- adjust icons such as "new jira item", "search", etc.
- review telemetry
- finish Jira item's UI

### How Has This Been Tested?

- [X] Lots of manual testing
- [X] new UTs

### DIFF
--- a/package.json
+++ b/package.json
@@ -228,6 +228,15 @@
                 "category": "Atlassian"
             },
             {
+                "command": "atlascode.jira.refreshAssignedWorkItemsExplorer",
+                "title": "Refresh Assigned Work Items",
+                "icon": {
+                    "dark": "resources/dark/refresh.svg",
+                    "light": "resources/light/refresh.svg"
+                },
+                "category": "Atlassian"
+            },
+            {
                 "command": "atlascode.jira.refreshCustomJqlExplorer",
                 "title": "Refresh Custom JQL Filters",
                 "icon": {
@@ -476,6 +485,11 @@
                     "when": "atlascode:jiraExplorerEnabled && config.atlascode.jira.enabled"
                 },
                 {
+                    "id": "atlascode.views.jira.assignedWorkItemsTreeView",
+                    "name": "Assigned Jira Work Items",
+                    "when": "config.atlascode.jira.enabled && atlascode:customJQLExplorerEnabled"
+                },
+                {
                     "id": "atlascode.views.jira.customJqlTreeView",
                     "name": "Custom Jql Filters",
                     "when": "config.atlascode.jira.enabled && atlascode:customJQLExplorerEnabled"
@@ -579,6 +593,11 @@
                     "command": "atlascode.jira.searchIssues",
                     "when": "view == atlascode.views.jira.customJqlTreeView && atlascode:isJiraAuthenticated",
                     "group": "navigation@3"
+                },
+                {
+                    "command": "atlascode.jira.refreshAssignedWorkItemsExplorer",
+                    "when": "view == atlascode.views.jira.assignedWorkItemsTreeView && atlascode:isJiraAuthenticated",
+                    "group": "navigation@2"
                 },
                 {
                     "command": "atlascode.jira.refreshCustomJqlExplorer",
--- a/src/container.ts
+++ b/src/container.ts
@@ -72,6 +72,7 @@ import { CheckoutHelper } from './bitbucket/interfaces';
 import { ProductJira } from './atlclients/authInfo';
 import { ATLASCODE_TEST_USER_EMAIL, ATLASCODE_TEST_HOST } from './constants';
 import { CustomJQLViewProvider } from './views/jira/treeViews/customJqlViewProvider';
+import { AssignedWorkItemsViewProvider } from './views/jira/treeViews/jiraAssignedWorkItemsViewProvider';
 import { Logger } from './logger';
 import { SearchJiraHelper } from './views/jira/searchJiraHelper';
 
@@ -88,7 +89,7 @@ export class Container {
             product: 'externalProductIntegrations',
             subproduct: 'atlascode',
             version: version,
-            deviceId: env.machineId,
+            deviceId: this.machineId,
             enable: this.getAnalyticsEnable(),
         });
 
@@ -187,10 +188,11 @@ export class Container {
 
         this._loginManager = new LoginManager(this._credentialManager, this._siteManager, this._analyticsClient);
         this._bitbucketHelper = new BitbucketCheckoutHelper(context.globalState);
+
         await FeatureFlagClient.initialize({
             analyticsClient: this._analyticsClient,
             identifiers: {
-                analyticsAnonymousId: env.machineId,
+                analyticsAnonymousId: this.machineId,
             },
             eventBuilder: new EventBuilder(),
         })
@@ -225,11 +227,13 @@ export class Container {
             context.subscriptions.push(new LegacyAtlascodeUriHandler(analyticsApi, bitbucketHelper));
         }
     }
+
     static initializeNewSidebarView(context: ExtensionContext, config: IConfig) {
         if (FeatureFlagClient.featureGates[Features.NewSidebarTreeView]) {
             Logger.debug('Using new custom JQL view');
             SearchJiraHelper.initialize();
-            context.subscriptions.push((this._customJqlViewProvider = new CustomJQLViewProvider()));
+            context.subscriptions.push(new CustomJQLViewProvider());
+            context.subscriptions.push(new AssignedWorkItemsViewProvider());
         } else {
             this.initializeLegacySidebarView(context, config);
         }
@@ -247,6 +251,7 @@ export class Container {
             });
         }
     }
+
     static initializeBitbucket(bbCtx: BitbucketContext) {
         this._bitbucketContext = bbCtx;
         this._pipelinesExplorer = new PipelinesExplorer(bbCtx);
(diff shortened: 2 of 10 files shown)


bwieger-atlassian-com:
@marcomura There's a 2 bugs that I could find: 

**On Activation Bug**
1. Be on the latest version of the extension eg: 3.4.12
2. Install the extension based on this branch 
3. Navigate quickly to Atlassian Extension 
4. Click on one of the Jira's 
5. Note the error shown 

**On Refresh Bug**
1. Install the extension based on this branch 
2. Wait for everything to load up properly. 
3. Click on refresh in the Assigned Work Items Section 
4. Immediately click on a JIRA 
5. Note the error shown 

https://www.loom.com/share/2dc932ffa41f468fb86f4ac4f98cb7eb?sid=7f5af4b2-8dc3-439f-9fc9-3f6dfc19534c


image

bwieger-atlassian-com (changes requested):
See bug comment

src/container.ts
```diff
@@ -88,7 +89,7 @@ export class Container {
             product: 'externalProductIntegrations',
             subproduct: 'atlascode',
             version: version,
-            deviceId: env.machineId,
+            deviceId: this.machineId,
```
bwieger-atlassian-com, line 92: Can you explain why you made this change?

src/container.ts
```diff
@@ -548,9 +553,4 @@ export class Container {
     static get bitbucketHelper() {
         return this._bitbucketHelper;
     }
-
-    private static _customJqlViewProvider: CustomJQLViewProvider;
```
bwieger-atlassian-com, line 552: Can you share why you removed this?

marcomura: it was unreferenced - aka, dead code

src/container.ts
marcomura: this method should be the source of truth for `machineId`, it's not good practice to spread the implementation:
```    static get machineId() {
        return env.machineId;
    }```
    
If in the future machineId() needs to do something extra to retrieve the machineId, we'll need to change only one place instead of many

**Review by bwieger-atlassian-com — APPROVED** [review]
LGTM! 

It's looking amazing. Excited for the finishing touches 💯

File: src/views/jira/searchJiraHelper.ts [hunk 3]
```diff
@@ -27,22 +27,26 @@ export class SearchJiraHelper {
...
         this._searchableIssueMap[dataProviderId] = flattenedIssueList;
     }
 
+    /***
+     * Append the issues for a specific data provider in the SearchJiraHelper.
+     * This is where `Commands.JiraSearchIssues` will look for issues to display
+     * @param flattenedIssueList - List of fetched issues
+     * @param dataProviderId - Id of the data provider (i.e. CustomJqlViewProvider.viewId())
+     */
+    static appendIssues(flattenedIssueList: MinimalORIssueLink<DetailedSiteInfo>[], dataProviderId: string) {
```
bwieger-atlassian-com, line 39: I believe we need to de-dup here

bwieger-atlassian-com:
Different bug found: 

1. Right-clicking on the JIRA in the Assigned Issues gives a context menu
2. Clicking on some of the options in the context menu results in an error (Assign to me, start work)

package.json
```diff
@@ -580,6 +594,11 @@
                     "when": "view == atlascode.views.jira.customJqlTreeView && atlascode:isJiraAuthenticated",
                     "group": "navigation@3"
                 },
+                {
+                    "command": "atlascode.jira.refreshAssignedWorkItemsExplorer",
+                    "when": "view == atlascode.views.jira.assignedWorkItemsTreeView && atlascode:isJiraAuthenticated",
+                    "group": "navigation@2"
+                },
```
cabella-dot, line 601: make sure to add the `atlascode.jira.searchIssues` and `atlascode.jira.createIssue` to this view 

i.e.
Screenshot 2025-03-10 at 5 01 02 PM

src/views/jira/searchJiraHelper.ts
cabella-dot: yeah, it seems that my previous logic is creating duplicates when I test this 



in `createIssueQuickPick` below, and easy fix that I found to work on this branch is to replace the derivation of `quickPickIssues` with 

```
const issueSet = new Set<string>();
const quickPickIssues: QuickPickIssue[] = [];

Object.values(this._searchableIssueMap).forEach((issueArray) => {
    issueArray.forEach((issue) => {
        if (!issueSet.has(issue.key)) {
            issueSet.add(issue.key);
            quickPickIssues.push({
                label: issue.key,
                description: issue.summary,
                issue: issue,
            });
        }
    });
});
```

cabella-dot: I can put this in a different PR

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        return [];
+    }
+
+    private buildTreeItemsFromIssues(issues?: MinimalIssue<DetailedSiteInfo>[]): TreeItem[] {
+        return issues ? issues.map((issue) => new JiraIssueNode(issue)) : [];
+    }
+}
+
+class JiraIssueNode extends TreeItem {
```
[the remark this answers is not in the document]
marcomura: `IssueNode` brings too much overhead, and I already removing the dependency of IssueNode and AbstractBaseNode from the custom jql panel too.
Can we move the status tag implementation in the much simpler `JiraIssueNode` class instead?

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        this._onDidChangeTreeData.fire();
+    }
+
+    dispose() {
+        this._disposable.dispose();
+    }
+
+    private refresh(): void {
+        this._onDidChangeTreeData.fire();
```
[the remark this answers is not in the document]
marcomura: `SearchJiraHelper.setIssues` replaces the collection, so there is no need to clean it.

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    getTreeItem(element: TreeItem): TreeItem {
+        return element;
+    }
+
+    async getChildren(element?: TreeItem): Promise<TreeItem[]> {
+        // this branch should never be triggered
+        if (element) {
```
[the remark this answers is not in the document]
marcomura: when a child has the property collapsibleState set to `None`, vscode assumes it's a leaf and it doesn't call the getChildren for it - so, for this view, it never happens that getChildren is called with element != undefined.

package.json
marcomura: I'll do it in the next PR :)

src/views/jira/searchJiraHelper.ts
marcomura: @bwieger-atlassian-com to clarify, `appendIssues` specifically doesn't need de-dup, as it's called from a logic that progressively discover new issues from different sites, therefore never duplicates.

the duplication is happening due to same issues being shown by both panels "Assigned jira issues" and "Custom jql queries". There needs to be de-duplication applied when reading all issues from different sources.

marcomura:
> Different bug found:
> 
> 1. Right-clicking on the JIRA in the Assigned Issues gives a context menu
> 2. Clicking on some of the options in the context menu results in an error (Assign to me, start work)

Thanks! Fixed!

marcomura:
>yeah, it seems that my previous logic is creating duplicates when I test this
>
>in createIssueQuickPick below, and easy fix that I found to work on this branch is to replace the derivation of quickPickIssues with

Applied! Thanks!!

bwieger-atlassian-com:
Bug: Right Click on "Custom JQL Filters" Jira item, attempt "assign to me", get an error 
image

image

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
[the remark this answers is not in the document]
cabella-dot: oh ok!

cabella-dot:
> Bug: Right Click on "Custom JQL Filters" Jira item, attempt "assign to me", get an error image
> 
> image

This is caused in file `assignIssue.ts` at line 19. It tries to explicitly call `Container.jiraExploere.refresh()` but because we are using the new view and it is not initialized it throws this error. @marcomura We would need to hold the new views in the Container and explicitly call their refresh functions in this flow.

marcomura:
> Bug: Right Click on "Custom JQL Filters" Jira item, attempt "assign to me", get an error

This bug is not on my view.
Please, let's more forward with the implementation, we can't fix everything in a single PR.

marcomura:
> > Bug: Right Click on "Custom JQL Filters" Jira item, attempt "assign to me", get an error 
> 
> This is caused in file `assignIssue.ts` at line 19. It tries to explicitly call `Container.jiraExploere.refresh()` but because we are using the new view and it is not initialized it throws this error. @marcomura We would need to hold the new views in the Container and explicitly call their refresh functions in this flow.

I think we can just call
```
vscode.commands.executeCommand(Commands.RefreshCustomJqlExplorer);
```
without exposing the view in the container.

Let's tackle this issue in a separate PR.

bwieger-atlassian-com (approved):
Approving to move forward. But let's be sure to be bug free before starting the experiment (to the best of our knowledge)

PR #194, opened 2025-03-13
AXON-193: make customJqlTreeView only visible when user has custom JQLs + fix bug + add util tests

cabella-dot:
### What Is This Change?

* Made `customJqlTreeView` to only be visible when user has custom JQLs enabled
* Made `assignedToMeTreeView `and `customJqlTreeview` only visible when Jira Explorer is enabled
* fixed bug where assigning yourself in Issue view does not refresh the tree
* Add util tests for new JiraIssueNode for coverage on statusCategory tag assignment
* added relevant commands to AssignedToMeView

Loom:
https://www.loom.com/share/7b1f04900c164c08a83e2d7dd8ac85d7?sid=1546e877-1257-4ac0-a3f2-4a0c29ad4e8d

### How Has This Been Tested?

Unit testing + Manual testing

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/views/jira/treeViews/customJqlViewProvider.ts
+++ b/src/views/jira/treeViews/customJqlViewProvider.ts
@@ -118,8 +118,11 @@ class JiraIssueQueryNode extends TreeItem {
                 return [JiraIssueQueryNode._treeItemNoIssuesMessage];
             }
 
-            issues = Container.config.jira.explorer.nestSubtasks ? await this.constructIssueTree(issues) : issues;
+            // index only issues that are directly retrieved with the JQL queries, and not
+            // the extra parent items retrieved to rebuild the issues hierarchy
             SearchJiraHelper.appendIssues(issues, CustomJQLViewProviderId);
+
+            issues = Container.config.jira.explorer.nestSubtasks ? await this.constructIssueTree(issues) : issues;
 
             return issues.map((issue) => new JiraIssueNode(JiraIssueNode.NodeType.CustomJqlQueriesNode, issue));
         })();
--- a/src/views/jira/treeViews/customJqlViewProvider.ts
+++ b/src/views/jira/treeViews/customJqlViewProvider.ts
@@ -40,8 +40,6 @@ export class CustomJQLViewProvider implements TreeDataProvider<TreeItem>, Dispos
     private _onDidChangeTreeData = new EventEmitter<TreeItem | undefined | void>();
     readonly onDidChangeTreeData = this._onDidChangeTreeData.event;
 
-    private _jqlEntries: JQLEntry[] = [];
-
     constructor() {
         this._disposable = Disposable.from(
             Container.jqlManager.onDidJQLChange(this.refresh, this),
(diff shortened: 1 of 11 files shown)


src/views/jira/treeViews/customJqlViewProvider.ts
```diff
@@ -45,24 +40,36 @@ export class CustomJQLViewProvider implements TreeDataProvider<TreeItem>, Dispos
     private _onDidChangeTreeData = new EventEmitter<TreeItem | undefined | void>();
     readonly onDidChangeTreeData = this._onDidChangeTreeData.event;
 
+    private _jqlEntries: JQLEntry[] = [];
```
marcomura, line 43: this seems to be unused

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
```diff
@@ -36,14 +38,24 @@ export class AssignedWorkItemsViewProvider implements TreeDataProvider<TreeItem>
...
-
+        Container.context.subscriptions.push(configuration.onDidChange(this.onConfigurationChanged, this));
         this._onDidChangeTreeData.fire();
     }
 
+    private onConfigurationChanged(e: ConfigurationChangeEvent) {
+        if (configuration.changed(e, 'jira.explorer.enabled')) {
+            setCommandContext(CommandContext.AssignedIssueExplorer, Container.config.jira.explorer.enabled);
+            this.refresh();
+        } else if (configuration.changed(e, 'jira.explorer')) {
```
marcomura, line 54: why do we need to refresh for `jira.explorer` config change?

marcomura:
Thanks for adding the icons like 'create new issue', 'search', the 3-dots menu, etc!
And for fixing the missing refresh when editing a jira issue in tab :)

src/views/jira/treeViews/customJqlViewProvider.ts
cabella-dot: i put this in to hold the JQLs so they are accessible in the configuration listener

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
cabella-dot: This is to refresh if any jira.explorer config changes happen. This would be 
* group issues by epic
* fetch all JQL query results
* show notifications
* refresh interval

So if the user wants to fetch more results, it should refresh the tree with the new issue results. WYDT

marcomura: makes sense, thanks

src/views/jira/treeViews/customJqlViewProvider.ts
marcomura: but the class field isn't really used - the data is always re-fetched from the local function, so it can be stored in a local variable

cabella-dot: ah good point. Thanks, I'll put it to a local variable instead!

marcomura (approved):
looks good!! thanks!

PR #203, opened 2025-03-17
First attempt at configuring test coverage thresholds

marcomura:
Because we run `unit` and `react` tests separately, and collect the code coverage for them separately, it's not straightforward to configure test coverage thresholds for our code.

With this PR, the jest configuration is refactored to centralize the configuration for both projects, and introduces the `coverageThreshold` setting that accounts for `unit` and `react` coverage separately.


bwieger-atlassian-com (approved):
Let's try it

PR #204, opened 2025-03-18
tsconfig: added alwaysStrict, noFallthroughCasesInSwitch, strictPropertyInitialization

marcomura:
Added the following typescript compiler options:

### Always Strict - [alwaysStrict](https://www.typescriptlang.org/tsconfig/#alwaysStrict)
Ensures that your files are parsed in the ECMAScript strict mode, and emit “use strict” for each source file.

### No Fallthrough Cases In Switch - [noFallthroughCasesInSwitch](https://www.typescriptlang.org/tsconfig/#noFallthroughCasesInSwitch)
Report errors for fallthrough cases in switch statements. Ensures that any non-empty case inside a switch statement includes either break, return, or throw. This means you won’t accidentally ship a case fallthrough bug.

### Strict Property Initialization - [strictPropertyInitialization](https://www.typescriptlang.org/tsconfig/#strictPropertyInitialization)
When set to true, TypeScript will raise an error when a class property was declared but not set in the constructor.


bwieger-atlassian-com (dismissed):
✅

bwieger-atlassian-com (approved):

PR #208, opened 2025-03-19
WIP: Move from npm to yarn

pastelsky:
Usually this sort of thing wouldn't matter — I thought npm's caught up a bit with yarn.
But turns out its still pretty bad when installing fat packages. I was trying to experiment getting @atlassian/editor into the extension,
and at the moment  npm takes a good 10+ minutes to npm install @atlassian/editor-core .
It takes several more minutes to run things like npm dedupe (been running for > 20 mins now).

Moving from `npm` to `yarn` essentially seems like a blocker for any reasonable devloop to work with editor core. 

This is still in draft because it likely requires the extension to be sanity tested once.


marcomura:
Closing as stale

PR #211, opened 2025-03-19
[AXON-235] Don't cache feature flags, and better telemetry around it

marcomura:
### What Is This Change?

We have run an AA experiment to exercise the experimentation pipeline and verify everything is going according with our expectations.
Unfortunately, it is not.

This PR tries to address what's going on with the following changes:
- Updated the `@atlaskit/feature-gate-js-client` package
- Remove caching of feature flags and experiments, so stat-sig can detect users' interaction correctly
- Improved telemetry about feature initialization, covering branches that were not covered before, and adding the following attributes:
  - `success` = `true`|`false`
  - `errorType` = `None`|`Failed`|`Skipped`|`IdMissing`

One change that is worth highlighting is how the `singleViewFactory` passes the feature flags and experiment gates to the view.
Before, it was passing the entire cache of the feature flags and experiment gates to every instantiated views.
After this change, every view's controller needs to declare which feature flags and/or experiments it needs, and the singleViewFactory only sends those to them, evaluating them on demand.

Another change to specifically approve is that I have consolidated two `featureFlagClient` telemetry events: `initialized` and `initializationFailed` into a single `initialized` event that carries an attribute indicating if it's successful or not.
I believe having a single event will help with aggregate metrics, and counting total attempts to retrieve the feature flags.
Also, the `initializationFailed` event was never registered in DataPortal anyway.

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 


src/util/featureFlags/features.ts
```diff
@@ -1,10 +1,9 @@
-export enum Features {
+export const enum Features {
     EnableNewUriHandler = 'atlascode-enable-new-uri-handler',
-    EnableAuthUI = 'atlascode-enable-auth-ui',
```
bwieger-atlassian-com, line 3: What made you remove this?

marcomura: We already have an experiment configured for it, so we don't need also a feature flag.
This was discussed before, that we should use either one or the other.

marcomura: Also, it was unreferenced.

bwieger-atlassian-com (approved):

PR #217, opened 2025-03-20
[AXON-239] Fix TreeItem id to allow same issue to appear twice

marcomura:
### What Is This Change?

The bug is the following error when configuring two distinct custom JQL queries that return the same Jira issue:
image

The reason for this error is because we build the TreeItem id as `{issueKey}_{siteId}`, which is the same for the same issue across different JQL queries.

The solution is building the TreeItem id as `{issueKey}_{siteId}_{jqlEntryId}`.

To accomodate this solution, some other changes are required:
- in jqlManager, the jqlEntry for retrieving the default jql queries for the assigned jira work items needs to have a stable id, so depending for the reason it's generated (site initialization vs retrieve default queries) the id is generated differently
- a new TreeViewIssue interface is introduced to extend the fields of a MinimalIssue, and carry the jql source for it

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

### DIFF
--- a/src/views/jira/treeViews/utils.ts
+++ b/src/views/jira/treeViews/utils.ts
@@ -13,18 +13,22 @@ export function createLabelItem(label: string, command?: Command): TreeItem {
     return item;
 }
 
+export interface TreeViewIssue extends MinimalIssue<DetailedSiteInfo> {
+    jqlSource: JQLEntry;
+    children: TreeViewIssue[];
+}
+
 /** This function returns a Promise that never rejects. */
-export async function executeJqlQuery(jqlEntry: JQLEntry): Promise<MinimalIssue<DetailedSiteInfo>[]> {
+export async function executeJqlQuery(jqlEntry: JQLEntry): Promise<TreeViewIssue[]> {
     try {
         if (jqlEntry) {
             const jqlSite = Container.siteManager.getSiteForId(ProductJira, jqlEntry.siteId);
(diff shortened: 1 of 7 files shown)


src/views/jira/treeViews/utils.ts
```diff
@@ -48,14 +52,14 @@ export class JiraIssueNode extends TreeItem {
...
+        public issue: TreeViewIssue,
     ) {
-        const collapsibleState = issue.subtasks.length
+        const collapsibleState = issue.children.length
             ? TreeItemCollapsibleState.Expanded
             : TreeItemCollapsibleState.None;
         super(issue.key, collapsibleState);
 
-        this.id = `${issue.key}_${issue.siteDetails.id}`;
+        this.id = `${issue.key}_${issue.siteDetails.id}_${issue.jqlSource.id}`;
```
bwieger-atlassian-com, line 62: Checking if this messes up the state...

marcomura: No it doesn't, these items are never stored. This id is used by vscode to know if I'm re-rendering something that is already rendered.

bwieger-atlassian-com: Cool, I did my check to be confident. Auth is still a fresh wound 😢

PR #239, opened 2025-03-22
Removed yaml as required extension

marcomura:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,7 @@
         "onStartupFinished"
     ],
     "extensionDependencies": [
-        "vscode.git",
-        "redhat.vscode-yaml"
+        "vscode.git"
     ],
     "extensionKind": [
         "workspace"


sdzh-atlassian:
We might need to also remove the activation of this extension on start-up. Not sure if there are any effects on the functionality elsewhere
image

marcomura:
> We might need to also remove the activation of this extension on start-up. Not sure if there are any effects on the functionality elsewhere

Aren't those stuff used for running e2e tests in the container?

marcomura:
Oh, no, it's something called at bootstrap in `extension.ts`:

```
    // Following are async functions called without await so that they are run
    // in the background and do not slow down the time taken for the extension
    // icon to appear in the activity bar
    activateBitbucketFeatures();
    activateYamlFeatures(context);
```

bwieger-atlassian-com (changes requested):
Marking as need changes based on internal discussion.

marcomura:
Postponed to better days: AXON-246

PR #244, opened 2025-03-25
AXON-182-enable-new-auth-ui

cabella-dot:
### What Is This Change?

* enable new auth UI for all new users on first time install with no auth
* fix bug where searching for resolution field when saving details was creating lag, which slowed down the `auto-skip` to next section after authentication

### How Has This Been Tested?

UT
manual testing

loom: 
https://www.loom.com/share/a0a3d869e5b94cb2b3b86c3d7dbf8b07?sid=3b37b7df-c58a-4bf1-9a57-c46f3d0183b6

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/atlclients/loginManager.ts
```diff
@@ -93,11 +94,10 @@ export class LoginManager {
                     await this._credentialManager.saveAuthInfo(siteInfo, oauthInfo);
 
                     if (site.product.key === ProductJira.key) {
-                        await this.updateHasResolutionField(siteInfo);
+                        this.updateHasResolutionField(siteInfo).then(() => this._siteManager.addSites([siteInfo]));
+                    } else {
+                        this._siteManager.addSites([siteInfo]);
```
bwieger-atlassian-com, line 99: Can you tell me more about this change?

cabella-dot: Sure! This `updateHasResoultionField` function calls the atlClient which adds latency. This occurs after authentication, so the user would be stuck at the Jira Page after authenticating until that process finishes. Only after that process finishes does the page switch to Bitbucket. With this change, I async call the `updateHasResolutionField` so the user can continue authenticating while the fields + issues are being fetched in the background

bwieger-atlassian-com: Sorry, I was referring to the `else` not the `await` 

Screenshot 2025-03-25 at 2 01 36 PM

cabella-dot: `siteManager.addSites` needs to be called after the `updateHasResolutionField` so the site data is correct. So if it is not Jira and does not call `updateResolutionField`, `siteManager.addSites` gets called on its own. I put that in there to make sure it is called after the `updateHasResolution` is resolved.

bwieger-atlassian-com (approved):

Issue #219, opened 2025-03-21
Settings ui doesn't open


flamein:
**Describe the bug**
Trying to open UI settings to login causes: 
```
command 'atlascode.extensionContext.showConfigPage' not found
```

**To Reproduce**
try to open ui settings

**Expected behavior**
I expected the UI settings to open

**Screenshots**

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (3.4.14 and 3.5.56) 
 - Visual Studio Code Version (1.98.2) 
 - Is this a local or cloud environment? (Local) 
 - If Local, OS (macOS)

**Additional context**

robert-steinhoefel:
I do have the same issue. But it's not only that. I also cannot see any JIRA issues nor add them.
When startis VSC, I get the warning that it could not connect to JIRA. Anything related to `atlascode.` is not found. It seems as if the extension is not installed properly. Switching to nightly path did not help as well as uninstalling and re-installing the extension.

![Image](https://github.com/user-attachments/assets/2ec1b4dc-a7d1-4218-b998-d465bb4e5c30)

bwieger-atlassian-com:
Looking

bwieger-atlassian-com:
@flamein @robert-steinhoefel  I'm attempting to reproduce. 

Can you tell me more about what you are doing? 
- Which button/command are you using to get this error? 
- Are you using a remote environment (WSL, ProjectIDX, some other container-ized or web-based IDE)?
- Are you using a variant of VSCode? (windsurf, cursor, VSCodium, etc)
- Have you changed any settings that might have a material impact?

robert-steinhoefel:
As for my part, I'm also using VS Code 1.98.2, no variant, locally installed (Win 11 Pro, 24H2, Build 26100.3476).
I'm trying to connec to JIRA Cloud.
The error occurs with *any* `Atlassian:`-command from the command palette (Ctrl+Shift+P) in this workspace. The error from the screenshot occurs when i click the gear-button in the JIRA Issues extension pane.
I have deactivated the VSC extension for some workspaces that are not using JIRA.

pcurren-atlassian:
Hi there.
I'm also unable to open the settings UI.
I'm Atlassian staff if you want to connect directly for further debugging.

aitriago:
Hi there:
No one of the Atlassian Jira features are running.
Installed locally in my Mac M1. 
- OS Sequoia 15.3.2 (24D81)
[...]

PR #248, opened 2025-03-26
better error handling

bwieger-atlassian-com:
### What Is This Change?

- Users are reporting issues with commands not being registers. This appears to be due to poor error handling. #219 
- Although we haven't pinpointed were the errors are coming from, we have improved the error handling allow users to manually reset themselves by deleting all of their auths in the Settings > Jira > Authentication section and then re-authenticating 

### How Has This Been Tested?
- Manually tested with a user who was experiencing this. Unfortunately, we didn't grab the log before hand to understand the cause of the error. But the solution worked. 

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -42,8 +42,6 @@ export async function activate(context: ExtensionContext) {
     try {
         await Container.initialize(context, configuration.get<IConfig>(), atlascodeVersion);
 
-        await JQLManager.backFillOldDetailedSiteInfo();
-
         registerCommands(context);
         activateCodebucket(context);
 
@@ -55,6 +53,8 @@ export async function activate(context: ExtensionContext) {
             CommandContext.IsBBAuthenticated,
             Container.siteManager.productHasAtLeastOneSite(ProductBitbucket),
         );
+
+        await JQLManager.backFillOldDetailedSiteInfos();
     } catch (e) {
         Logger.error(e, 'Error initializing atlascode!');
     }
--- a/src/jira/jqlManager.ts
+++ b/src/jira/jqlManager.ts
@@ -24,14 +24,12 @@ export class JQLManager extends Disposable {
     // we have introduced a new field in DetailedSiteInfo that is populated at auth time.
     // For those who already have this data saved before the introduction of the new logic,
     // we need to backfill this field to avoid constructing a wrong default JQL query.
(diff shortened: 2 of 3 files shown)


src/extension.ts
```diff
@@ -55,6 +53,8 @@ export async function activate(context: ExtensionContext) {
             CommandContext.IsBBAuthenticated,
             Container.siteManager.productHasAtLeastOneSite(ProductBitbucket),
         );
+
+        await JQLManager.backFillOldDetailedSiteInfos();
```
bwieger-atlassian-com, line 57: Moved down to allow commands to be registered even if some exception slips through

src/jira/jqlManager.ts
```diff
@@ -24,14 +24,12 @@ export class JQLManager extends Disposable {
...
+    public static async backFillOldDetailedSiteInfos(): Promise<void> {
         for (const site of Container.siteManager.getSitesAvailable(ProductJira)) {
-            if (site.hasResolutionField === undefined) {
-                const client = await Container.clientManager.jiraClient(site);
-                const fields = await client.getFields();
-                site.hasResolutionField = fields.some((f) => f.id === 'resolution');
-
-                Container.siteManager.addOrUpdateSite(site);
+            try {
+                await JQLManager.backFillOldDetailedSiteInfo(site);
```
bwieger-atlassian-com, line 30: Try catch on each item instead of the whole loop to allow for processing even if there are some failures.

src/jira/jqlManager.ts
```diff
@@ -42,6 +40,16 @@ export class JQLManager extends Disposable {
         this._disposable = Disposable.from(configuration.onDidChange(this.onConfigurationChanged, this));
     }
 
+    private static async backFillOldDetailedSiteInfo(site: DetailedSiteInfo) {
```
bwieger-atlassian-com, line 43: Refactored into a function for ease of reading

src/jira/jqlManager.ts
marcomura: Good catch!
Just one nitpick: I think the try/catch would look better inside the `backFillOldDetailedSiteInfo` function.

marcomura (approved):
lgtm

PR #255, opened 2025-03-27
Refactor common test utilities into their folder

marcomura:
### What Is This Change?

Created a root folder `testsutil` where we can implement UT utility functions.

For now, we only have `forceCastTo<>`.
More to come 😄

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-#registry=https://packages.atlassian.com/api/npm/npm-remote
-unsafe-perm = true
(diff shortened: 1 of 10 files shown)


.npmrc
marcomura, line 1: Deleted this to solve these npm warnings:
```
npm warn Unknown env config "unsafe-perm". This will stop working in the next major version of npm.
npm warn Unknown project config "unsafe-perm". This will stop working in the next major version of npm.
```

sdzh-atlassian (approved):

PR #265, opened 2025-03-29
[AXON-267] Added telemetry for errors and exceptions

marcomura:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/errorReporting.ts
+++ b/src/errorReporting.ts
@@ -1,0 +1,94 @@
+import { TrackEvent } from './analytics-node-client/src/types';
+import { errorEvent } from './analytics';
+import { Logger } from './logger';
+import { AnalyticsClient } from './analytics-node-client/src/client.min';
+
+let nodeJsErrorReportingRegistered = false;
+let analyticsClientRegistered = false;
+
+let analyticsClient: AnalyticsClient | undefined;
+let eventQueue: Promise<TrackEvent>[] | undefined = [];
+
+function errorHandler(error: Error): void {
+    try {
+        if (error.hasOwnProperty('__atlascode_logged')) {
+            return;
+        }
+        (error as any).__atlascode_logged = true;
+
+        Logger.debug('[ERROR]', error);
+
+        const event = errorEvent(error);
+
+        if (analyticsClient) {
+            event.then((e) => analyticsClient!.sendTrackEvent(e));
+        } else {
+            eventQueue!.push(event);
+        }
+    } catch {}
+}
+
+function replaceJavascriptErrorObject() {
+    // if we are calling this twice, probably it's because the extension got upgraded and it's re-initializing.
+    // we need to retrieve the original Error and re-create another override
+    const originalError = Error.hasOwnProperty('__atlascode_originalError')
+        ? ((Error as any).__atlascode_originalError as ErrorConstructor)
+        : Error;
+
+    Error = class Error extends originalError {
+        constructor(message?: string) {
+            super(message);
+
+            // for Errors created in the wild, we filter for the string `atlascode` to avoid noise
+            if (message && message.includes('atlascode')) {
+                errorHandler(this);
+            }
+        }
+    } as ErrorConstructor;
+
+    (Error as any).__atlascode_originalError = originalError;
+}
+
+function restoreJavascriptErrorObject() {
+    if (Error.hasOwnProperty('__atlascode_originalError')) {
+        Error = (Error as any).__atlascode_originalError as ErrorConstructor;
+    }
+}
+
+export function registerErrorReporting(): void {
+    if (!nodeJsErrorReportingRegistered) {
+        nodeJsErrorReportingRegistered = true;
+
+        try {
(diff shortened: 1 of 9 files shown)


src/errorReporting.ts
```diff
@@ -0,0 +1,69 @@
...
+let analyticsClientRegistered = false;
+
+let _logger_onError_eventRegistration: Disposable | undefined = undefined;
+
+let analyticsClient: AnalyticsClient | undefined;
+let eventQueue: Promise<TrackEvent>[] | undefined = [];
+
+function errorHandler(error: Error | string): void {
+    try {
+        Logger.debug('[LOGGED ERROR]', error);
```
bwieger-atlassian-com, line 17: Why is this at Debug level?

src/errorReporting.ts
```diff
@@ -0,0 +1,69 @@
...
+            process.addListener('unhandledRejection', errorHandler);
+
+            _logger_onError_eventRegistration = Logger.onError((data) => errorHandler(data.error), undefined);
+        } catch {}
+    }
+}
+
+export function unregisterErrorReporting(): void {
+    try {
+        process.removeListener('uncaughtException', errorHandler);
```
bwieger-atlassian-com, line 45: Same here, all-or-nothing, or best-effort-based?

src/errorReporting.ts
```diff
@@ -0,0 +1,69 @@
...
+            eventQueue!.push(event);
+        }
+    } catch {}
+}
+
+export function registerErrorReporting(): void {
+    if (!nodeJsErrorReportingRegistered) {
+        nodeJsErrorReportingRegistered = true;
+
+        try {
```
bwieger-atlassian-com, line 33: try catch should be around the full function body, not partial.

src/errorReporting.ts
```diff
@@ -0,0 +1,69 @@
...
+            process.addListener('uncaughtException', errorHandler);
```
bwieger-atlassian-com, line 34: I assume this is some javascript magic that listens to certain events, is that right?

src/views/jira/treeViews/utils.ts
```diff
@@ -47,6 +47,15 @@ export const loginToJiraMessageNode = createLabelItem('Please login to Jira', {
     arguments: [ProductJira],
 });
 
+/** Takes only the last group of the UUID if it's a real UUID. Otherwise, it returns the input as is. */
+function anonymizeUUID(uuid: string) {
```
bwieger-atlassian-com, line 51: Why does this exist?

src/views/jira/treeViews/utils.ts
```diff
@@ -61,7 +70,9 @@ export class JiraIssueNode extends TreeItem {
 
         // this id is constructed to ensure unique values for the same jira issue across multiple jql queries.
         // therefore, multiple jql entries must have a unique id for the same site.
-        this.id = `${issue.key}_${issue.siteDetails.id}_${issue.jqlSource.id}`;
+        //
+        // we anonymize the UUID taking just a segment of it, so it can't be considered PII.
+        this.id = `atlascode_${issue.key}_${anonymizeUUID(issue.siteDetails.id)}_${anonymizeUUID(issue.jqlSource.id)}`;
```
bwieger-atlassian-com, line 75: Why are we modifying this?

src/errorReporting.ts
marcomura: this is just an debugging echo to highlight what we are sending to telemetry

src/views/jira/treeViews/utils.ts
marcomura: oops leftover, thanks for spotting it

src/views/jira/treeViews/utils.ts
marcomura: it's a leftover of the Error override part, removing it - thanks

src/errorReporting.ts
marcomura: while every scenario has different requirements, in general try catch should be around what you want to protect from unhandled failures.

in this scenario, the if() and the assignment can't fail, so I didn't include them in the try.

personally, I think the code is more readable if it's written surgically to do exactly what you intend it to do.
e.g., a try / catch on an entire function would make you wonder "what could really fail here?"

src/errorReporting.ts
marcomura: same as other comment

src/errorReporting.ts
bwieger-atlassian-com: Okay, my though is around the likelihood of someone adding new code around this. 

Maybe change to a guard statement instead? 

```
if (nodeJsErrorReportingRegistered) return; 
... rest of function 
```

marcomura: done :)

bwieger-atlassian-com (approved):
Yay!

PR #266, opened 2025-03-30
Updated tsconfig files for esnext lib + polish content

marcomura:
### What Is This Change?

Upgraded the lib version to 'esnext' to include some most recent stuff.
Since lib != target, something will get polyfilled.

Polished the content of the tsconfig files to contain the same compilation flags in the same order.


tsconfig.notest.json
```diff
@@ -4,11 +4,12 @@
         "outDir": "build/extension",
         "module": "esnext",
         "target": "es2018",
-        "lib": ["es2018", "dom"],
+        "lib": ["ESNext", "dom"],
```
bwieger-atlassian-com, line 7: What's your thinking behind using ESNext? 

I am asking from a curious point of view. I don't have a formed opinion here.

marcomura: The typescript compilers offers the possibility to use the latest libraries (most recent functions, classes, etc) with an older target, and it automatically polyfill what's missing.
I think it would be nice to be able to use whatever the latest ecmascript has to offer.

bwieger-atlassian-com (approved):
LGTM! 

A little existential fear about ESNext picking up something not backwards compatible, but that's a larger problem than we need to deal with.

PR #288, opened 2025-04-08
Revamp top header for PR Details page

akalia-atlassian:
### What Is This Change?

This PR does two major things in two different commits:

1. Adds better loaders across various parts of the page
2. Completely revamps how the top section works by using atlaskit components like forntbucket uses

### How Has This Been Tested?

- [x] This was tested thoroughly earlier but not after the rebase.
- [x] Moreover, this needs to have a conversation with @cabella-dot because of some CSS issues while bundling

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:


cabella-dot (dismissed):

src/react/atlascode/common/AtlGlobalStyles.tsx
```diff
@@ -7,6 +7,90 @@ const useStyles = makeStyles(
     (theme: Theme) =>
         ({
             '@global': {
+                // Using Atlassian specific tokens
```
pastelsky, line 10: Consider using token values from `@atlaskit/tokens` instead of hardcoding them here.

Also would recommend seeing if we can directly use values from the package instead so its more maintainable in the future.

Can be done as a follow up.

pastelsky (dismissed):
LGTM, have a few comments on how we do token mapping — but that can be discussed and fixed separately.

src/react/atlascode/common/AtlGlobalStyles.tsx
akalia-atlassian: Okay, created a ticket for the same. Resolving this for now.

pastelsky (approved):

PR #291, opened 2025-04-08
fix: mutex problems in auth code

akalia-atlassian:
### What Is This Change?

This PR majorly exists to fix a very glaring problem in the auth info flow.

In `getAuthInfo()`, we have a read from secret storage and then delete from secret storage.

For certain flows in hardcoded site, we saw parallel get and save auth info results in:

1. read in get operation
2. save in save operation
3. delete in get operation

This interleaving means that the save operation is incorrectly undone. And this is because `get/delete` should be an atomic operation. This PR exists to ensure the same. 

This PR also has some semi-related refactoring:
1. Moving logger to the very top of activate function so that initial logs in certain functions are not missed.
2. `getKeyForSecrets` is introduced to ensure that the key for secretstorage doesn't mistakenly diverges over time. 
3. `refreshAccessToken` returns boolean now: this is quite unrelate but needed in future PRs. Its existing return value is not used anywhere. 

### How Has This Been Tested?

- [x] This was found and fixed while testing a nasty bug in devsphere where authentication used to disappear for new devboxes. This was the reason for that.

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


Aline1973tech (approved):

src/atlclients/authStore.ts
```diff
@@ -59,21 +65,40 @@ export class CredentialManager implements Disposable {
      * it's available, otherwise will return the value in the secretstorage.
      */
     public async getAuthInfo(site: DetailedSiteInfo, allowCache = true): Promise<AuthInfo | undefined> {
-        return this.getAuthInfoForProductAndCredentialId(site, allowCache);
+        return this.mutex.runExclusive(() => {
```
marcomura, line 68: I don't have clear how this is going to solve the race condition.
runExclusive should be done on the caller to avoid A+B+C to be broken up, no?

akalia-atlassian: @marcomura Within `getAuthInfoForProductAndCredentialId`, we call:

1. `this.getAuthInfoFromKeychain`
2. `this.removeSiteInformationFromKeychain`

This is what we are trying to make atomic: I found that in certain cases, it was happening that:

1. This function's `this.getAuthInfoFromKeychain` got called.
2. Another function's `this.saveAuthInfo` was called.
3. This function's `this.removeSiteInformationFromKeychain` was called.

In this race condition, we have no site auth info in the end, despite a saveAuthInfo.

PR #292, opened 2025-04-08
Auth Info centralisation

akalia-atlassian:
### What Is This Change?

This PR is a much-needed (according to me) refactoring that we faced: we could not use PAT tokens easily because many places within the code just assumes that we will necessarily be following the Oauth / bearer tokens. This PR does away with that assumption and creates auth headers always on the basis of auth info.

Moreover, this reimagines the auth info as tagged unions: this ensures that type checking is very easy and we can differentiate empty auth info from actual ones. This is still backward compatible i.e. we do not rely on `.type` field introduced in this PR because stored auth infos won't have those fields. But eventually, in some time after the release, all auth infos will have a direct type field instead of relying on fragile logic of other fields's presence.

### How Has This Been Tested?

- [x] Tested locally.
 
Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/atlclients/authInfo.ts
```diff
@@ -189,18 +198,21 @@ export const emptyAccessibleResourceV1: AccessibleResourceV1 = {
 };
 
 export const emptyAuthInfo: AuthInfo = {
```
pastelsky, line 200: NoAuthInfo?

src/atlclients/authInfo.ts
```diff
@@ -240,19 +252,25 @@ export function isEmptySiteInfo(a: any): boolean {
     );
 }
 
-export function isOAuthInfo(a: any): a is OAuthInfo {
-    return a && (<OAuthInfo>a).access !== undefined && (<OAuthInfo>a).refresh !== undefined;
+export function isOAuthInfo(a: AuthInfo | undefined): a is OAuthInfo {
+    // This check should be retired over time when auth info is updated to use the new type
+    const oldCheck = a && (<OAuthInfo>a).access !== undefined && (<OAuthInfo>a).refresh !== undefined;
```
pastelsky, line 257: Make it clear in comments that this has not been changed for backward compatibility reasons

src/atlclients/authInfo.ts
```diff
@@ -240,19 +252,25 @@ export function isEmptySiteInfo(a: any): boolean {
...
+    return oldCheck || (!!a && a.type === 'oauth');
```
pastelsky, line 258: Need to upgrade all existing tokens to use the new types.

PR #300, opened 2025-04-09
AXON-319: add option to explicitly fire exposure event

cabella-dot:
### What Is This Change?
Problem: Our Statsig client is unable to send exposure events to Statsig even though we are connected

Possible solution:

Add option to explicitly fire exposure/check event through Statsig client. Safe to do so in this context as there is no possibility of false exposure for AA exp

### How Has This Been Tested?

Unfortunately, Statsig does not get our staging events/exposures so need to release to production to see impact.

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/util/featureFlags/client.ts
+++ b/src/util/featureFlags/client.ts
@@ -136,7 +136,7 @@ export abstract class FeatureFlagClient {
         let gateValue = false;
         if (FeatureGates.initializeCompleted()) {
             // FeatureGates.checkGate returns false if any errors
-            gateValue = FeatureGates.checkGate(gate);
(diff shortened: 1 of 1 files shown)


bwieger-atlassian-com (dismissed):
LGTM, do as you need.

marcomura (approved):
According to the doc, `fireGateExposure` should default to true, so in theory this shouldn't be necessary.

Removing the `analyticsWebClient` is a long shot as well, since according to Yoni it shouldn't be used to send the gate check beacon back.

Either way, in absence of other ideas, let's give it a try! :)

PR #112, opened 2025-02-13
Fixes 111 Corrupted diff when a file path includes unusual characters

marcellourbani:
fixes #111

Issue #111, opened 2025-02-13
Corrupted diff when a file path includes unusual characters


marcellourbani:
**Describe the bug**
My files have names like src/#bti#te_cl_check_authorisation.clas.abap. The extension doesn’t handle them properly, which leads to bogus results.

**To Reproduce**

Create a pull request in a project with files who start by #foo#
click on the fiff 

**Expected behavior**
want to see the diff 

**Screenshots**

This looks like this in Bitbucket cloud:

![Image](https://github.com/user-attachments/assets/c2455319-0166-452a-abd4-afdaab63c73c)

and like this in vscode:

![Image](https://github.com/user-attachments/assets/1ab8a715-0c0c-45af-bedd-bf0542bc94e3)

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (3.5.15) 
 - Visual Studio Code Version (1.97.0) 
 - Is this a local or cloud environment? (Local) 
 - If Local, OS ( linux)

**Additional context**

Was https://bitbucket.org/atlassianlabs/atlascode/issues/5434

marcomura:
Hello @marcellourbani ,

I'm struggling to get a repo for this issue with atlascode v3.4.25.

I tried creating a file with a unicode in it, or with a '#' in it, but the diff for both look good:

Bitbucket:
![Image](https://github.com/user-attachments/assets/01df46ad-c67f-4fd4-b787-d0c4bbf81f63)

Atlascode:
![Image](https://github.com/user-attachments/assets/12b01602-f079-4686-af53-6b24f6040e09)
![Image](https://github.com/user-attachments/assets/0e3ce51f-7573-4d5f-bc7a-5ce5e0798bf0)

What am I missing?

marcellourbani:
There's a race condition: the plugin tries to get a file from the API AND frokm the filesystem, whoever replies first wins.
Usually the API wins when the file version is not there (i.e. I pushed it, you didn't pull it yet)
But sometimes API wins anyway, leading to intermittent errors

I don' have a public test repo on bitbucket, but [I have this in github who is suitable](https://github.com/marcellourbani/btidummy) with files who would reproduce the issue
Also have a dozen private ones, like [this](https://bitbucket.org/basistech/activecontrol_backend_new/src/main/) but I assume you don't have access

To reproduce the error I usually delay or comment out the file bit, in gitContentProvider.ts lines 33-49 ()in my fix:

```typescript
[...]

PR #303, opened 2025-04-10
Marcellourbani/issue111 corrupted diff when a file path includes unusual characters

marcellourbani:
### What Is This Change?

Addresses issue #111 .
Escapes file names in URLs to support odd ones (like /namespace/file)

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Got a bunch of errors in test, but no new ones (they fail in main too)

This is a repeat of #112 who was stuch for a dumb bot issue


src/bitbucket/bbUtils.ts
```diff
@@ -150,3 +150,7 @@ export function workspaceRepoFor(repository: Repository): WorkspaceRepo {
         siteRemotes: siteRemotes,
     };
 }
+
+export function encodePathParts(path: string) {
+    return path.split('/').map(encodeURIComponent).join('/');
```
marcomura, line 155: just to be extra safe, can you null check path?
`return path?.split...`

marcomura:
Thanks for this PR @marcellourbani !
Nice catch and fix for it.

I'll take some time tomorrow to run a test pass on it - meanwhile can you address the small comment I left?

src/bitbucket/bbUtils.ts
marcellourbani: shouldn't be needed: path is string, not string|undefined
Anyway I did it

marcomura: if only that'd be enough to prevent a null/undefined to be passed... :)
it's better to be safe than crashing

Issue #308, opened 2025-04-10
Issue view is blank


internetlee:
**Describe the bug**
Experiencing something very similar to #46 . Issue detail view doesn't render anything. I see the loading indicator and Atlassian logo while the editor tab reads "Jira Issue", then loading state goes away, tab title is replaced with the issue key, and editor tab is blank.

**To Reproduce**
1. Go to default assigned to me filter or a custom JQL filter in the Activity Menu sidebar
2. Click on any issue
3. Wait for loading state to clear
4. Editor pane for the issue view is blank

**Expected behavior**
Opening an issue from the filter list should allow me to see the issue view in an editor tab

**Screenshots**

Webview DevTools when trying to open an issue view (end of the output many more `processing field` log lines before this`)
![Image](https://github.com/user-attachments/assets/7717f3b4-1ca5-4476-8f8a-8878d1b28def)

Editor Tab
![Image](https://github.com/user-attachments/assets/b3d9fecc-8048-4848-9d39-2d16c233a141)

ExtensionHost output (not certain when this is happening)
![Image](https://github.com/user-attachments/assets/9e399de6-c8b1-4165-8779-ddc087a19d19)

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version `3.4.24` (also attempted with pre-release version on Apr 10, 2025)
 - Visual Studio Code Version
```
Version: 1.99.2
Commit: 4949701c880d4bdb949e3c0e6b400288da7f474b
Date: 2025-04-10T01:21:25.295Z
Electron: 34.3.2
ElectronBuildId: 11161073
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Darwin arm64 23.2.0
```
 - Is this a local or cloud environment? local, connected to Jira Cloud
 - If Cloud, Browser (chrome, safari) n/a
 - If Local, OS (macOS, windows, linux) macos

**Additional context**
Only authed with Jira Cloud, we don't use BitBucket.

First attempt to use this again in a few months as I previously had persistent authentication errors before. I can now consistently get authed, stay authed, update settings, update JQL custom queries and get accurate filter results (further than I've ever had luck with before), but now I can't see the issue view.

bwieger-atlassian-com:
@internetlee I will look into this immediately. Thank you!

bwieger-atlassian-com:
@internetlee, can you tell me more about the Jira item you are trying to open? 

Is this a story/bug/epic? 

[...]

PR #310, opened 2025-04-10
Fix Issue #308 - Some Jira Issues Cannot load - Make sure we are only rendering fields that exist

bwieger-atlassian-com:
### What Is This Change?
- Issue #308 reported an issue around some Jiras not loading because some fields aren't present and we didn't handle this correctly. 

### How Has This Been Tested?
- Manual testing via making a project on a my personal jira site that didn't have one of the key fields. 

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
+++ b/src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
@@ -556,7 +556,7 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
     }
     commonSidebar(): any {
         const commonItems: SidebarItem[] = ['assignee', 'reporter', 'labels', 'priority', 'components', 'fixVersions']
-            .filter((field) => this.state.fields[field] !== undefined)
+            .filter((field) => !!this.state.fields[field])
             .map((field) => {
                 return {
                     itemLabel: this.state.fields[field].name,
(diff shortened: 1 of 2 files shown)


cabella-dot (dismissed):

src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
```diff
@@ -555,19 +555,14 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
...
-            'components',
-            'fixVersions',
-        ].map((field) => {
-            return {
-                itemLabel: this.state.fields[field].name,
-                itemComponent: this.getInputMarkup(this.state.fields[field], true, field),
-            };
-        });
+        const commonItems: SidebarItem[] = ['assignee', 'reporter', 'labels', 'priority', 'components', 'fixVersions']
+            .filter((field) => this.state.fields[field] !== undefined)
```
marcomura, line 559: nitpick: I would not explicitly check for 'undefined' as it would still break for 'null' - `!!this.state.fields[field]` would work better

marcomura (dismissed):
left one comment but can go as is

marcomura (approved):
lgtm - thanks for addressing the comment!

PR #314, opened 2025-04-11
fix: Attempt a third merge of the reverted PR

akalia-atlassian:
### What Is This Change?

So, I was trying to figure out why is it working for dev but not prod: and then I realised - we do not need very intelligent chunking solutions given we package and ship our extension in one go unlike websites.

So, I have removed a chunking section for CSS from the extension and that seems to have fixed it for the production package.

But the same change is not working for the existing main: so, both deduplication and removal of chunking fixed it for me. But deduplicating is its own thorny issue:

1. I have fixed some small CSS problems that me and @cabella-dot  found.
2. I have removed the @atlaskit/panel altogether and copied it over as is: this is a hack but this is so that changes are from existing setup are minimal. We can iterate over this.
3. Similarly, I copied over the colour mapping from atlaskit CSS variables to VS Code CSS variables: I couldn't find a single place where CSS for both BitBucket and Jira pages would apply so this is the solution for now. Again, to be fixed eventually.

### How Has This Been Tested?

Ran through all the views and user actions: followed the script that @cabella-dot  did. I am facing two issues right now:

- [ ] Not able to open the create issue page
- [ ] Not able to change assignees or labels - any select boxes

I spent some time looking into if select box might have changed its schema but it has not. And then I figured the same issues exist in `main` too. So, I am not sure anymore how to seprate testing for just this PR so that it successfully gets merged this time.

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/webviews/components/App.css
```diff
@@ -2093,3 +2088,83 @@ pre {
 .commit-files-summary .commit-files-summary--filename {
     margin-left: 5px;
 }
+
+:root {
```
bwieger-atlassian-com, line 2092: @marcomura @cabella-dot is this something we are comfortable with?

bwieger-atlassian-com:
Screenshot 2025-04-14 at 11 23 28 AM

bwieger-atlassian-com:
Otherwise LGTM

src/webviews/components/App.css
marcomura: why wasn't it required before?

PR #321, opened 2025-04-12
Make Webpack fail on warnings

marcomura:
### What Is This Change?

With this change, Webpack is going to fail the build if there are warnings.


bwieger-atlassian-com (approved):
Great idea

PR #334, opened 2025-04-14
AXON-344: add window object

cabella-dot:
### What Is This Change?

Based on [this thread](https://atlassian.slack.com/archives/C04PR2YE4UC/p1742429892425979), it was discoved that the exposure events were being blocked because the Statsig client was recognizing our extension as a server environment rather than a client. This check is shown in this [code here](https://github.com/statsig-io/js-client-monorepo/blob/05021e3b0ed79e7ae09ded914dff6089561853ec/packages/client-core/src/SafeJs.ts#L13-L25)

As a work around, I defined a window object so the Stastig client can declare us as a client side app and let us send exposure events back to statsig

Test exposures on `atlascode-noop` flag: https://console.statsig.com/LqivKg6ADZZaGczRfBKfX/gates/atlascode-noop/diagnostics

Test exposures in `atlascode_aa_experiment`: https://console.statsig.com/LqivKg6ADZZaGczRfBKfX/experiments/atlascode_aa_experiment/diagnostics

I have opened this issue [here](https://github.com/statsig-io/js-client-monorepo/issues/23) so Statsig is aware of this issue for vscode developers
 
### How Has This Been Tested?

Test exposures for AA experiment and FG

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,10 +27,12 @@ import { GitExtension } from './typings/git';
 import { FeatureFlagClient } from './util/featureFlags';
 
 const AnalyticDelay = 5000;
+
 export async function activate(context: ExtensionContext) {
     const start = process.hrtime();
 
     registerErrorReporting();
+
     const atlascode = extensions.getExtension('atlassian.atlascode')!;
     const atlascodeVersion = atlascode.packageJSON.version;
     const previousVersion = context.globalState.get<string>(GlobalStateVersionKey);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -1,6 +1,6 @@
 import { pid } from 'process';
 import * as semver from 'semver';
-import { commands, env, ExtensionContext, extensions, languages, Memento, window } from 'vscode';
+import * as vscode from 'vscode';
```
marcomura, line 3: this is not needed, and importing individual components is better

cabella-dot: good call!

bwieger-atlassian-com (dismissed):

marcomura (approved):
yay, looking forward to roll this out!

src/extension.ts
```diff
@@ -90,14 +93,12 @@ export async function activate(context: ExtensionContext) {
     // icon to appear in the activity bar
     activateBitbucketFeatures();
     activateYamlFeatures(context);
-
```
marcomura, line 93: not to be annoying, but these new lines were making the code easier to read 😆

cabella-dot: oh oops I guess when I was changing the imports it removed the some whitespace 😅

marcomura (approved):
thanks!!

PR #335, opened 2025-04-14
Gate the error telemetry behind a feature flag to limit the exposure

marcomura:
### What Is This Change?

The existing error telemetry, if unconstrained, seems to be sending between 2M to 3M events for 150k users, which is too much.

This change gates it under a feature flag so we can control the % of users who will be sending error telemetry events, and hopefully maintain the same ratio of failures by getting a fraction of them.

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:


cabella-dot (approved):

cabella-dot (approved):

src/extension.ts
```diff
@@ -103,6 +103,14 @@ export async function activate(context: ExtensionContext) {
     );
 }
 
+function activateErrorReporting(): void {
```
bwieger-atlassian-com, line 106: can you wrap this with safeCall block?

marcomura: mmh, nothing can really fail there... all those functions are already well protected

bwieger-atlassian-com (approved):

PR #344, opened 2025-04-15
Remove JQL initialization on Jira site add

marcomura:
### What Is This Change?

With the new Jira panels, we don't need the JQL initialization for every site anymore.

The default query is built ad-hoc for the assigned jira work items panel, and we only need custom JQLs manually configured by the user.

Side effect: users who are upgrading from previous versions will see the default queries appearing in the Custom JQL panel, and they have to delete them manually.
Unfortunately, I think there aren't safest alternatives to this.

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:


cabella-dot (approved):
lgtm

bwieger-atlassian-com (approved):

PR #364, opened 2025-04-21
Enrich launched event

bwieger-atlassian-com:
### What Is This Change?

The current launched event doesn't track Authentication State. So, adding it in. 

Why is the necessary? 

We need to approximate the number of calls we will make to the notification service. Without knowing the number of auth'ed users who are launching VSCode in a given day, this is hard. 

Added in dataPortal: 
https://data-portal.internal.atlassian.com/analytics/registry/18825?tabIndex=Attributes

### How Has This Been Tested?
Manual debug to confirm that the is<Product>Authenticate works as expected. 

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -51,19 +51,19 @@ export async function upgradedEvent(version: string, previousVersion: string): P
 
 export async function launchedEvent(
     location: string,
-    isJiraCloudAuthenticated: boolean,
-    isJiraServerAuthenticated: boolean,
-    isBitbucketCloudAuthenticated: boolean,
-    isBitbucketServerAuthenticated: boolean,
+    numJiraCloudAuthed: number,
+    numJiraDcAuthed: number,
+    numBitbucketCloudAuthed: number,
+    numBitbucketDcAuthed: number,
 ): Promise<TrackEvent> {
     return trackEvent('launched', 'atlascode', {
         attributes: {
             machineId: Container.machineId,
             extensionLocation: location,
-            jiraCloudAuthenticated: isJiraCloudAuthenticated,
-            jiraServerAuthenticated: isJiraServerAuthenticated,
-            bitbucketCloudAuthenticated: isBitbucketCloudAuthenticated,
-            bitbucketServerAuthenticated: isBitbucketServerAuthenticated,
+            numJiraCloudAuthed: numJiraCloudAuthed,
+            numJiraDcAuthed: numJiraDcAuthed,
+            numBitbucketCloudAuthed: numBitbucketCloudAuthed,
+            numBitbucketDcAuthed: numBitbucketDcAuthed,
         },
     });
 }
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -51,15 +51,19 @@ export async function upgradedEvent(version: string, previousVersion: string): P
 
 export async function launchedEvent(
     location: string,
-    isJiraAuthenticated: boolean,
-    isBitbucketAuthenticated: boolean,
+    isJiraCloudAuthenticated: boolean,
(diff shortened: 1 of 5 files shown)


src/analytics.ts
```diff
@@ -49,9 +49,18 @@ export async function upgradedEvent(version: string, previousVersion: string): P
...
+    isJiraAuthenticated: boolean,
+    isBitbucketAuthenticated: boolean,
+): Promise<TrackEvent> {
     return trackEvent('launched', 'atlascode', {
-        attributes: { machineId: Container.machineId, extensionLocation: location },
+        attributes: {
+            machineId: Container.machineId,
+            extensionLocation: location,
+            jiraAuthenticated: isJiraAuthenticated,
+            bitbucketAuthenticated: isBitbucketAuthenticated,
```
bwieger-atlassian-com, line 62: Aside from adding this to the data portal, is there any other steps necessary? 

https://data-portal.internal.atlassian.com/analytics/registry/18825?tabIndex=Attributes

marcomura: just if you want it in Amplitude, you have to enable it explicitely

src/analytics.ts
```diff
@@ -49,9 +49,18 @@ export async function upgradedEvent(version: string, previousVersion: string): P
...
+    location: string,
+    isJiraAuthenticated: boolean,
+    isBitbucketAuthenticated: boolean,
+): Promise<TrackEvent> {
     return trackEvent('launched', 'atlascode', {
-        attributes: { machineId: Container.machineId, extensionLocation: location },
+        attributes: {
+            machineId: Container.machineId,
+            extensionLocation: location,
+            jiraAuthenticated: isJiraAuthenticated,
```
marcomura, line 61: would it make sense to split jira into jiraCloud and jiraDC ?
same for bitbucket

bwieger-atlassian-com: oh, I like this idea!

src/vscAnalyticsApi.ts
```diff
@@ -72,8 +72,20 @@ export class VSCAnalyticsApi implements AnalyticsApi {
...
+    public async fireLaunchedEvent(
+        location: string,
+        isJiraCloudAuthenticated: boolean,
+        isJiraServerAuthenticated: boolean,
+        isBitbucketCloudAuthenticated: boolean,
+        isBitbucketServerAuthenticated: boolean,
+    ): Promise<void> {
+        return launchedEvent(
+            location,
+            isJiraCloudAuthenticated,
```
bwieger-atlassian-com, line 84: Modifying to track the number

src/extension.ts
```diff
@@ -182,7 +182,13 @@ async function sendAnalytics(version: string, globalState: Memento) {
         });
     }
 
-    launchedEvent(env.remoteName ? env.remoteName : 'local').then((e) => {
+    launchedEvent(
```
sdzh-atlassian, line 185: Random thought for when we rework this eventually - in cases like this we could benefit from a single-object-parameter, along the lines of
```js
launchedEvent({
    env: env.remoteName || 'local',
    siteCount: {
        jiraCloud: Container.siteManager.numberOfSites(ProductJira, true),
        jiraServer: Container.siteManager.numberOfSites(ProductJira, false),
        // ...
    },
});
```

bwieger-atlassian-com: Yeah, the dataPortal supports type Map, but I don't know what that means for when it reaches Amplitude

Presumably, it does type Map<String, ???> 

But what's in the ???

sdzh-atlassian (approved):
ship it! ⛵

PR #375, opened 2025-04-29
Fixed Logger.error calls, and added a description field to error telemetry

marcomura:
### What Is This Change?

Lots of `Logger.error` calls pass the exception as the first argument, and a description of the error as the second argument.
The second argument was not captured by the `errorTelemetry` logic, so this PR is using the second argument (if present) as an override of the Error message. The stack preserves the original error message.

Moreover, there were several snippets of code like this:

```
catch (e) {
    Logger.error(new Error(`New error message: ${e}`));
    ...
}
```

Not only this is pretty useless, since it can just call `Logger.error(e, 'New error message')`, but it was also destroying the original stack trace and creating a new one.
Fixed all of them.

Additionally, this PR:
- prevents the minifier to rename function names, so they are preserved in the stack trace
- added some extra errors related to authentication

- added a logic to auto-detect the function name where the error got logged:
image

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`


src/logger.ts
```diff
@@ -85,23 +86,23 @@ export class Logger {
         }
     }
 
-    public static error(ex: Error, classOrMethod?: string, ...params: any[]): void {
-        this.Instance.error(ex, classOrMethod, params);
+    public static error(ex: Error, classOrMethod?: string): void {
```
sdzh-atlassian, line 89: Why is this called `classOrMethod` I wonder 🤔

sdzh-atlassian: Nevermind, found the answer in some usage, like this:
```js
Logger.error(e, 'StartWorkOnIssueWebview.forceUpdateIssue');
```

We use it in a different way in most places though, so maybe let's rename this to `description`? Not that it matters much, anyway :D

sdzh-atlassian (dismissed):
Looks good to me! ⛵

src/logger.ts
marcomura: changed :)

webpack.react.prod.js
```diff
@@ -101,8 +102,8 @@ module.exports = {
         }),
     ],
     performance: {
-        maxEntrypointSize: 262144,  // overridden to 256KiB, reccomended is 244KiB
-        maxAssetSize: 524288,       // overridden to 512KiB, reccomended is 244KiB
+        maxEntrypointSize: 307200,  // overridden to 300KiB, reccomended is 244KiB
```
sdzh-atlassian, line 105: Does this max value have any margin, or is it the exact current value? :)

src/atlclients/authStore.ts
```diff
@@ -139,18 +139,23 @@ export class CredentialManager implements Disposable {
                             await this.removeSiteInformationFromKeychain(site.product.key, site.credentialId);
                         } else if (Container.siteManager.getSiteForId(site.product, site.id)) {
                             // if keychain does not have any auth info for the current site but the site has been saved, we need to remove it
-                            Logger.debug(
-                                `removing dead site for product ${site.product.key} credentialID: ${site.credentialId}`,
+                            Logger.error(
+                                new Error(
```
sdzh-atlassian, line 143: I might have misunderstood the description - but didn't we want to remove cases like this with `new error)?
If there's no thrown error here, should it really go through the same `logger.error` + analytics?

webpack.react.prod.js
marcomura: It has a good margin :)
new entrypoint is 257 KiB, and new asset is 548 KiB

src/atlclients/authStore.ts
marcomura: In this case it's legit to create a new Error() instance because we are "producing" an error ad-hoc, opposed to capturing an error via try/catch where we should preserve the original one.

sdzh-atlassian (approved):
Once again, looks good to me! ;) ⛵

bwieger-atlassian-com (approved):

PR #376, opened 2025-04-30
[AXON-374] chore: add format-on-save to the workspace settings

sdzh-atlassian:
### What Is This Change?

Small ways-of-working improvement - we now format file automatically on save after editing. Returning from the other project, this was bugging me quite a bit ;D

The configuration for this is added to workspace settings, which are now no longer `.gitignore`-d

### How Has This Been Tested?

Edited some files, then saved them ;) Observed the expected behavior

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 
- [x] ~If Atlassian employee & Bitbucket changes: did you test with DC in mind? [See Instructions](https://www.loom.com/share/71e5d17734a547f68fd6128be6cd760e?sid=835e58a7-1240-498d-b2d7-fa7fdf8ffa36)~ N/A

Recommendations:
- [x] ~Update the CHANGELOG if making a user facing change~ N/A

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ build
 node_modules
 .vscode-test/
 .vscode/*
+!.vscode/settings.json
 launch.json
 !.vscode/launch.json.example
 *.vsix


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
marcomura, line 3: is this the same formatter that our lint is using today?

sdzh-atlassian: Good catch! This is `prettier`, which is part of our `lint-fix` process (and a personal preference :D)
There _is_ a way to use `eslint` here - but when I tried it, the eslint extension had some issues picking up our configs, and I decided to not dig into it further for now

In theory, `prettier` should cover the absolute majority of cases, but I'm happy to iterate on this if it ever proves insufficient

marcomura (approved):
lgtm

PR #381, opened 2025-05-04
Added more util UTs

marcomura:
### What Is This Change?

Self explanatory 😄 
Check comments for some extra details.

### How Has This Been Tested?

- [x] manual tested
- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/util/files.ts
+++ b/src/util/files.ts
@@ -5,21 +5,20 @@ export interface FileWithContent extends File {
     fileContent: string | undefined;
 }
 
-export function readFilesContentAsync(files: File[]): Promise<FileWithContent[]> {
-    const promise = new Promise<any>((resolve) => {
-        let doneCount = 0;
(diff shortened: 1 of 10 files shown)


src/util/files.ts
```diff
@@ -5,21 +5,20 @@ export interface FileWithContent extends File {
...
     });
+}
 
-    return promise;
+async function getFileWithContent(file: File): Promise<FileWithContent> {
+    (file as FileWithContent).fileContent = await readFile(file);
+    return file as FileWithContent;
+}
+
+export function readFilesContentAsync(files: File[]): Promise<FileWithContent[]> {
```
marcomura, line 22: I wrote the original code for this, and I decided to rewrite it to make it clearer.
Tested with attachments, there are no regressions.

Compared to the previous implementation, it now handles errors while before it was pending on them forever.

bwieger-atlassian-com (approved):

PR #382, opened 2025-05-05
Clean up unused mui and webview configs

marcomura:
### What Is This Change?

Removed a bunch of unused `mui` and `webview` scripts from package.json, and related webpack configuration files.
Also, removed some leftovers from the bitbucket issues cleanup.

### How Has This Been Tested?

- [X] running and debugging the extension still works
- [X] `extension:install` still works


src/analyticsTypes.ts
```diff
@@ -31,7 +31,6 @@ export enum AnalyticsView {
     SettingsPage = 'page:v2:settings',
 
     BitbucketIssuePage = 'page:v2:bitbucket:issue',
-    CreateBitbucketIssuePage = 'page:v2:bitbucket:createIssue',
```
sdzh-atlassian, line 34: I guess this is the bit that we deprecated?

src/react/index-dev.tsx
```diff
@@ -51,7 +51,6 @@ const ports: Record<string, number> = {
     atlascodeSettingsV2: UIWSPort.Settings,
     atlascodeOnboardingV2: UIWSPort.Onboarding,
     bitbucketIssuePageV2: UIWSPort.BitbucketIssuePage,
-    createBitbucketIssuePageV2: UIWSPort.CreateBitbucketIssuePage,
```
sdzh-atlassian, line 54: IIRC this `UIWSPort` stuff is what I've been talking about @marcomura - regarding the old method of live updates in the browser

package.json
```diff
@@ -70,32 +70,14 @@
...
         "test:e2e:docker:rerun": "e2e/scripts/in-docker npm run test:e2e:rerun",
-        "devcompile": "npm-run-all --parallel devcompile:react devcompile:extension",
+        "devcompile": "npm run clean && npm-run-all --parallel devcompile:react devcompile:extension",
         "devcompile:react": "webpack --mode development --config webpack.react.dev.js --fail-on-warnings",
         "devcompile:extension": "webpack --mode development --config webpack.extension.dev.js --fail-on-warnings",
         "release:stable": "scripts/release.sh",
         "watch": "concurrently -r \"npm:watch:*\"",
         "watch:extension": "WATCH=true webpack --mode development --config webpack.react.dev.js --progress --watch",
         "watch:react": "WATCH=true webpack --mode development --config webpack.extension.dev.js --progress --watch",
-        "dev": "concurrently -r npm:watch:extension npm:watch:react",
```
sdzh-atlassian, line 80: Thanks for the clean-up! For a bit of context from our chat with @marcomura - this whole thing is sort of replaced by `npm run dev` now

sdzh-atlassian (approved):
<3

src/analyticsTypes.ts
marcomura: yep

src/react/index-dev.tsx
marcomura: oh I see - I'll take a look at that separately to see if we can remove that part too then.
Thanks!

PR #407, opened 2025-05-15
AXON-405: Remove FF for badges and banners

bwieger-atlassian-com:
### What Is This Change?

Badges are out in production now. 

Data indicates that it increases the rate of a no-auth launch -> Authentication by a 2x factor. 

Removing FF so we can accelerate the change faster. 

Note: The FF strategy is enforcing bake times that are too slow. 

### Tests
- [x] Unit tests passed. 
- [x] Manual testing to ensure nothing broke

### DIFF
--- a/src/views/notifications/bannerDelegate.ts
+++ b/src/views/notifications/bannerDelegate.ts
@@ -70,11 +70,8 @@ export class BannerDelegate implements NotificationDelegate {
             let count = 0;
             if (event.action === NotificationAction.Added) {
                 event.notifications.forEach((notification) => {
-                    this.showNotification(
-                        notification,
-                        this.makeActionText(notification),
-                        this.makeActionFunction(notification),
-                    );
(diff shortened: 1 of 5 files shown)


src/views/notifications/bannerDelegate.ts
```diff
@@ -70,11 +70,8 @@ export class BannerDelegate implements NotificationDelegate {
             let count = 0;
             if (event.action === NotificationAction.Added) {
                 event.notifications.forEach((notification) => {
-                    this.showNotification(
-                        notification,
-                        this.makeActionText(notification),
-                        this.makeActionFunction(notification),
-                    );
+                    const { text, action } = this.makeAction(notification);
```
bwieger-atlassian-com, line 73: Simplifying logic here per @sdzh-atlassian request in the last PR. 

Technically, this has nothing to do with the FF removal

marcomura (approved):

PR #413, opened 2025-05-17
Updated ts-loader package and others

marcomura:
### What Is This Change?

Updated `ts-loader` package, and more others.

### Benchmarks

```
npm dedupe  36.57s user 0.84s system 69% cpu 53.548 total
```

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [x] `npm run package:install`
- [x] manual testing


webpack.extension.prod.js
```diff
@@ -144,6 +144,7 @@ module.exports = [
         resolve: {
             extensions: ['.tsx', '.ts', '.js'],
             alias: {
+                axios: path.resolve(__dirname, 'node_modules/axios'),
```
sdzh-atlassian, line 147: What's the background on this?

Small general call-out - this is one place where our `e2e` tests are helping already - I once ran into a not-very-obvious error here that made `.vsix` artifact borked in production, but not debug. Good that we would now catch this in the build! :D

tsconfig.json
```diff
@@ -11,7 +11,7 @@
         "sourceMap": true,
         "allowJs": true,
         "jsx": "react",
-        "moduleResolution": "node",
+        "moduleResolution": "bundler",
```
sdzh-atlassian, line 14: What's the context of this change?

webpack.extension.prod.js
marcomura: mmh, I think this was a change that copilot did and was totally unecessary... I'll double check.
Thanks for pointint it out!

tsconfig.json
marcomura: Some updated packages require updating the module resolution:

```
TS2307: Cannot find module 'merge-anything' or its corresponding type declarations.
  There are types at '/Users/.../Repos/atlascode/node_modules/merge-anything/dist/index.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
```

PR #419, opened 2025-05-24
Added AI-generated unit tests for src files

marcomura:


src/logger.test.ts
```diff
@@ -0,0 +1,309 @@
+import { expansionCastTo } from 'testsutil/miscFunctions';
+import { ConfigurationChangeEvent, Disposable, ExtensionContext, LogOutputChannel, window } from 'vscode';
+
+import { configuration, OutputLevel } from './config/configuration';
+import { extensionOutputChannelName } from './constants';
+import { Container } from './container';
+import { ErrorEvent, Logger } from './logger';
```
sdzh-atlassian, line 7: Don't we have error analytics hooked into `Logger`? Feels a bit sus to use it in tests IMO

sdzh-atlassian (approved):

src/logger.test.ts
sdzh-atlassian: bruh 👁️  👄  👁️
My bad 🤦

PR #425, opened 2025-05-28
Use npm ci in pipeline

marcomura:
### What Is This Change?

`npm ci` is generally faster than `npm install` because it installs packages directly from the `package-lock.json` file without checking for updates or modifying it. This makes `npm ci` particularly efficient for clean installations in automated environments.


cabella-dot (approved):
cool!

PR #427, opened 2025-05-29
[AXON-432] Transition issue from the context menu

sdzh-atlassian:
### What Is This Change?

A small PR to add a way to transition issues right from the `ASSIGNED` list on the left 😉 

There's a new dropdown menu button:
image

Which does this when you click it:
image

If you have fancy transitions, you get a fancy picker! 😁 
image

### How Has This Been Tested?

` M A N U A L L Y ` (over a rather short period of time 😢 )

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 
- [ ] ~If Atlassian employee & Bitbucket changes: did you test with DC in mind? [See Instructions](https://www.loom.com/share/71e5d17734a547f68fd6128be6cd760e?sid=835e58a7-1240-498d-b2d7-fa7fdf8ffa36)~ - Nope, this is very much a quick fix only tested for cloud

Recommendations:
- [ ] Update the CHANGELOG if making a user facing change - TODO


sdzh-atlassian:
Remaining stuff:
 * Analytics
 * Changelog
 * More cleanup?
 *

src/commands.ts
```diff
@@ -193,6 +195,33 @@ export function registerCommands(vscodeContext: ExtensionContext) {
             commands.executeCommand(Commands.ShowIssue, issueNode.issue),
         ),
         commands.registerCommand(Commands.AssignIssueToMe, (issueNode: IssueNode) => assignIssue(issueNode)),
+        commands.registerCommand(Commands.TransitionIssue, async (issueNode: IssueNode) => {
+            const issue = issueNode.issue as MinimalIssue<DetailedSiteInfo>;
```
marcomura, line 199: shouldn't it check if issue is a MinimalIssue? `isMinimalIssue(...)`
even inside the IssueNode class it's checked

src/analytics.ts
```diff
@@ -189,8 +189,12 @@ export async function issueCreatedEvent(site: DetailedSiteInfo, issueKey: string
...
 }
 
-export async function issueTransitionedEvent(site: DetailedSiteInfo, issueKey: string): Promise<TrackEvent> {
-    return instanceTrackEvent(site, 'transitioned', 'issue', { actionSubjectId: issueKey });
+export async function issueTransitionedEvent(
+    site: DetailedSiteInfo,
+    issueKey: string,
+    method?: string,
+): Promise<TrackEvent> {
+    return instanceTrackEvent(site, 'transitioned', 'issue', { actionSubjectId: issueKey, method });
```
marcomura, line 197: nitpick: `source` instead of `method` ?

src/commands.ts
```diff
@@ -193,6 +195,33 @@ export function registerCommands(vscodeContext: ExtensionContext) {
...
+                    },
+                )
+                .then(async (transition) => {
+                    if (!transition) {
+                        return;
+                    }
+
+                    const target = issue.transitions.find((x) => x.name === transition.label);
+                    if (!target) {
+                        window.showErrorMessage(`Transition ${transition.label} not found`);
```
marcomura, line 218: I believe this is never supposed to happen since the labels are being built from the same transitions data 😆 but anyway, can you add a `Logger.error` so we can catch it in telemetry?

src/analytics.ts
sdzh-atlassian: @cabella-dot and I actually pondered for a bit on what to call this, with `source` being among the options ;D
Happy to change it to that

FWIW, the rationale I had behind `method` was because this is describing a `transition`, and `source` could possibly refer to the source state of the transition

src/commands.ts
sdzh-atlassian: Added the check, but I'm pretty sure it should only ever be a `MinimalIssue` here 🤔

src/commands.ts
marcomura: you can create one ad-hoc:
```
Logger.error(new Error(`Transition ${transition.label} not found`), 'Transition not found', transition.label);
```

Note the second string simplified is not a typo - it's for aggregating multiple instances of this error even if the labels are different 😄

sdzh-atlassian:
Rebased to main, double-checked that everything works

cabella-dot (approved):
Looks great!

src/commands.ts
sdzh-atlassian: Added a `Logger.error` in line with usage elsewhere

PR #428, opened 2025-05-29
[AXON-434] Disable the pipeline notifications from within

marcomura:
### What Is This Change?

Notifications for pipelines are very useful, but sometimes can be distracting if they trigger too frequently in very large teams.
With this change, the notification will provide a button to dismiss and disable this type of notifications.

The notifications for piplines can be re-enabled from the corresponding config setting.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [x] manual tests

Recommendations:


src/bitbucket/terminal-link/createPrLinkProvider.ts
```diff
@@ -149,6 +148,7 @@ export class BitbucketCloudPullRequestLinkProvider extends Disposable implements
         return env.openExternal(Uri.parse(url));
     }
 
-    private disable = () =>
-        configuration.update('bitbucket.showTerminalLinkPanel', false, ConfigurationTarget.Workspace);
+    private disable() {
+        configuration.updateEffective('bitbucket.showTerminalLinkPanel', false);
```
cabella-dot, line 152: does `updateEffective` disable for all config targets?

marcomura: no, it only updates a single target, but it auto-detects which target is the correct one.
it's not what we want for the terminal link provider?

cabella-dot: No, that seems like a better approach! I was just curious. Thanks for the explanation

cabella-dot (dismissed):

cabella-dot (dismissed):

src/views/notifications/PipelinesMonitor.ts
```diff
@@ -56,21 +64,20 @@ export class PipelinesMonitor implements BitbucketActivityMonitor {
         return !target.ref_name || shouldDisplay(target);
     }
 
+    // TODO check this function: it doesn't seem correct, it skips results
```
bwieger-atlassian-com, line 88: Do you want to address this now or later?

marcomura: later

src/analytics.ts
```diff
@@ -249,13 +249,15 @@ export async function searchIssuesEvent(product: Product): Promise<TrackEvent> {
...
+    source: string | undefined,
+    uri: Uri | undefined,
     notificationSurface: NotificationSurface,
     delta: number,
 ): Promise<TrackEvent> {
     return trackEvent('changed', 'notification', {
         attributes: {
-            uri: uri.toString(),
+            source,
+            uri: uri?.toString(),
```
bwieger-atlassian-com, line 237: I made URI required here as each notification should be identified by a URI. In some case the URI can be "made up"

src/views/notifications/PipelinesMonitor.ts
```diff
@@ -8,6 +9,9 @@ import { Container } from '../../container';
 import { Pipeline, PipelineTarget } from '../../pipelines/model';
 import { BitbucketActivityMonitor } from '../BitbucketActivityMonitor';
 import { descriptionForState, generatePipelineTitle, shouldDisplay } from '../pipelines/Helpers';
+import { NotificationSurface } from './notificationManager';
+
+const NotificationBannerId = '';
```
bwieger-atlassian-com: Was this meant to be blank & const?

src/analytics.ts
bwieger-atlassian-com: For your use case, probably something like: 

```diffs[i].site.details.id + diffs[i].repository.url + diffs[i].build_number.toString()```

File: src/analytics.ts [hunk 5]
```diff
@@ -249,13 +249,15 @@ export async function searchIssuesEvent(product: Product): Promise<TrackEvent> {
 }
 
 export async function notificationChangeEvent(
-    uri: Uri,
+    source: string | undefined,
```
bwieger-atlassian-com: I believe the source should be captured in the URI.

bwieger-atlassian-com:
@marcomura For the analytic part, it looks like you wanted to support the following use case: 

1. I don't have a specific URI, but I want to show where this notification came from

If that's the case, the URI field is meant to capture that. (ex: the existing Jira Login button). All that's need is some way to construct the URI. But I

src/views/notifications/PipelinesMonitor.ts
marcomura: it wasn't, thanks for catching it

src/analytics.ts
marcomura: let's chat about this

src/analytics.ts
marcomura: discussed this on 1:1 - Implemented what we have agreed on

src/analytics.ts
marcomura: discussed this on 1:1 - Implemented what we have agreed on

bwieger-atlassian-com (approved):

PR #444, opened 2025-06-04
Remove double dashing in branch names when name uses square brackets

avuy:
### What Is This Change?

This is the change that fixes the following issue:
During starting of a task, which will lead me to create and checkout branch, the branch name will always have double dash if my task make use of square bracket:

task name: [module] create function xxx
generated branch name module--create-function-xxx

the fix will change for the name to be module-create-function-xxx

### How Has This Been Tested?

There is a test written for that in the PR

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/react/atlascode/startwork/StartWorkPage.test.tsx
+++ b/src/react/atlascode/startwork/StartWorkPage.test.tsx
@@ -1,50 +1,15 @@
-// buildBranchName logic for testing
-function buildBranchName(branchType: { prefix: string }, state: { issue: { key: string; summary: string } }) {
-    return {
-        prefix: branchType.prefix.replace(/ /g, '-').toLowerCase(),
-        Prefix: branchType.prefix.replace(/ /g, '-'),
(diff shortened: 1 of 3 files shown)


src/react/atlascode/startwork/StartWorkPage.test.tsx
```diff
@@ -0,0 +1,81 @@
+// buildBranchName logic for testing
+function buildBranchName(branchType: { prefix: string }, state: { issue: { key: string; summary: string } }) {
```
marcomura: copying the production code in the test file nullify the purpose of unit testing as the production code can change without this code changing.

one way to solve it is exporting the function in a utility file, or something like that.

@cabella-dot , do you have another suggestion regarding how to test this logic from the view?

avuy: Thank you @marcomura , just added a fix

marcomura:
@avuy , can you please sign the CLA which allows you to contribute to the code? thanks!

PR #451, opened 2025-06-06
[AXON-441] Add telemetry to record which IDE Atlascode is running under (VS Code, Cursor, etc)

marcomura:
### What Is This Change?

This changes adds an attribute for the `launchedEvent` event that records which IDE the extension is running under, using a new attribute called `ideUriScheme`.

The value is taken from the custom uri scheme the editor registers to in the operating system.
Examples:
- In VS Code, the uri scheme is `vscode`
- In Cursor, the uri scheme is `cursor`
- In WindSurf, the uri scheme is `windsurf`

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- Tested with VS Code, Cursor, and WindSurf

### DIFF
--- a/src/lib/webview/controller/config/configWebviewController.ts
+++ b/src/lib/webview/controller/config/configWebviewController.ts
@@ -1,9 +1,10 @@
 import { defaultActionGuard } from '@atlassianlabs/guipi-core-controller';
 import Axios from 'axios';
 import { v4 } from 'uuid';
-import vscode from 'vscode';
+import vscode, { env } from 'vscode';
 
 import { isBasicAuthInfo, isEmptySiteInfo, isPATAuthInfo } from '../../../../atlclients/authInfo';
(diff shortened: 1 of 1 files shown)


src/lib/webview/controller/config/configWebviewController.ts
```diff
@@ -17,6 +18,8 @@ import { CommonActionMessageHandler } from '../common/commonActionMessageHandler
 import { MessagePoster, WebviewController } from '../webviewController';
 import { ConfigActionApi } from './configActionApi';
 
+const AUTH_URI = `${env.uriScheme || 'vscode'}://${ExtensionId}/auth`;
```
sdzh-atlassian, line 21: This is a solid fix, thanks - didn't realize we still had the `vscode` scheme hardcoded in places 😱

sdzh-atlassian (approved):

PR #466, opened 2025-06-13
Fixed the E2E docker build for playwright

marcomura:
### What Is This Change?

Thanks so much @sdzh-atlassian for basically doing all this change yourself 😆 

This changes is fixing the E2E docker image build to prevent docker & local host from having two different playwright versions, which will prevent the tests to run successfully (because of playwright deps).

1. Pin the playwright in package.json
2. Pin the same playwright version from package.json in Dockerfile when installing playwright
3. Fixed another issue regarding code-server sometimes listening to ipv4 instead of ipv6, changing from `[::1]` to `localhost` in the playwright test file.

### DIFF
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -3,7 +3,7 @@ FROM node:23-slim
 # Install dependencies and xvfb
 RUN apt-get update \
     && apt-get install -y wget gpg xauth curl git \ 
-    && npm install --global playwright \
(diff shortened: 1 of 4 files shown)


sdzh-atlassian (dismissed):
૮ ˶ᵔ ᵕ ᵔ˶ ა

sdzh-atlassian (approved):
(˶˃ ᵕ ˂˶) .ᐟ.ᐟ

PR #473, opened 2025-06-18
Exclude pipeline from experiments

marcomura:
### What Is This Change?

This change excludes the GitHub pipeline runs from any experiment or feature flag from statsig.
It also disables statsig logging to avoid unnecessary exposure.

This change is necessary to ensure E2E test runs are stable while experiments are run.


src/analytics.ts
```diff
@@ -170,30 +170,6 @@ export async function featureFlagClientInitializedEvent(
     });
 }
 
-// debugging event, meant to measure the exposure rate of a feature flag or an experiment
-export async function featureGateExposureBoolEvent(
```
marcomura, line 174: These events were created to validate statsig exposure when we weren't seeing any exposure in the statsig website.
I don't believe we need them anymore.

PR #475, opened 2025-06-18
FIX: fix issue when commit url is undefined

apeni:
### What Is This Change?
Fix of the issue when hyperlink for commit fails to open. The issue is related to Server Bitbucket. We don’t pass the url for commit. That’s why when  the url is undefined or malformed VS Code's webview tries to interpret this as a local resource path.
I’ve reproduced the issue by hardcoding the url to be undefined for Bitbucket Cloud. 

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:


marcomura (approved):
looks good, thanks!

PR #477, opened 2025-06-19
[AXON-370] Display full issues hierarchy in breadcrumbs

avuyteam:
### What Is This Change?

This change will allow us to display all hierarchy in Jira issues and not just the direct parent

### How Has This Been Tested?

This has been tested manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 ### Features
 
 - Added a `Don't show again` button in the pipeline notification to easily disable this type of notifications
+- Implemented full parent hierarchy display for Jira issues
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Update logic of selecting the default project and site to the most relevant one for the user on the Creating a JIRA issue page
 - Added filtering for options and error message in projects field on Create Jira issue page.
+- Implemented full parent hierarchy display for Jira issues
 
 ### Bug Fixes
 
@@ -25,7 +26,6 @@
 ### Features
 
 - Added a `Don't show again` button in the pipeline notification to easily disable this type of notifications
-- Implemented full parent hierarchy display for Jira issues
 
 ### Bug Fixes
 
--- a/src/lib/jira/constants.ts
+++ b/src/lib/jira/constants.ts
@@ -1,0 +1,5 @@
+// Top-level issue types in Jira that can be parent issues
+export const TOP_LEVEL_ISSUE_TYPES = ['Epic', 'Initiative', 'Project'] as const;
+
+// Type for top-level issue types
+export type TopLevelIssueType = (typeof TOP_LEVEL_ISSUE_TYPES)[number];
--- a/src/lib/jira/constants.ts
+++ b/src/lib/jira/constants.ts
@@ -1,5 +1,1 @@
-// Top-level issue types in Jira that can be parent issues
 export const TOP_LEVEL_ISSUE_TYPES = ['Epic', 'Initiative', 'Project'] as const;
-
-// Type for top-level issue types
-export type TopLevelIssueType = (typeof TOP_LEVEL_ISSUE_TYPES)[number];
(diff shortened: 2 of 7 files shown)


src/webviews/jiraIssueWebview.ts
```diff
@@ -141,12 +144,50 @@ export class JiraIssueWebview
...
+            // Send initial update with current issue and immediate parent
+            const msg = {
+                ...this._editUIData,
+                type: 'update',
+                hierarchy: initialHierarchy,
+            };
 
             this.postMessage(msg);
 
+            // Start loading rest of the hierarchy in the background
```
marcomura: while this is happening, there is no visual indication for it. e.g., a loading spinner, or a 3-dots filler.
It should show it in the form of `... / parent / this` while it's still fetching for more.

marcomura (changes requested):
please, introduce a visual indication that the hierarchy is still loading

marcomura:
When I tried opening multiple Jira issues, something in the code is crashing:
image

src/webviews/jiraIssueWebview.ts
bhorai-atl: Added 3-dot filler animation: 
 https://www.loom.com/share/db910177745c4b08858a618e09b40e1c?sid=c5586297-7a39-459a-a29c-c01b2fcb4e1b
 @marcomura

bhorai-atl:
@marcomura please take another look

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Features
 
 - Added a `Don't show again` button in the pipeline notification to easily disable this type of notifications
+- Implemented full parent hierarchy display for Jira issues
```
marcomura: Sorry, we release last Thursday, so this now should be added to 3.8.7 :)

src/lib/jira/constants.ts
```diff
@@ -0,0 +1 @@
+export const TOP_LEVEL_ISSUE_TYPES = ['Epic', 'Initiative', 'Project'] as const;
```
cabella-dot: Are these Static? Or can these types be modified by a project admin?

src/webviews/jiraIssueWebview.ts
```diff
@@ -993,4 +1046,86 @@ export class JiraIssueWebview
 
         return relatedPrs.filter((pr) => pr !== undefined).map((p) => p!.data);
     }
+
+    private async fetchIssueHierarchy(
```
cabella-dot: It looks like this function and the code above has lots of repeated code and will possibly post a message and call the jira client many times. If we want to recursively return the hierarchy and render the hierarchy as it loads, is it possible to do something simpler?

Here is a very rough example of what I mean that can be called after the initial render:
``` ts
private async fetchIssueHierarchy(currentIssueDetails, currentHierarchy, client, siteDetails) {
   if (!currentIssueDetials.fields.parent || TOP_LEVEL_ISSUE_TYPES.includes(currentIssueDetails.issuetype.name)) {
      return;
   }
   
   const parent = await fetchIssue(currentIssueDetials.fields.parent.key, siteDetails);
   const parentDetails = await client.getIssue(parent.key, fields);

   postMessage(...);

   await fetchIssueHierarchy(parentDetails, [parent, ...currentHierarchy], client, siteDetails)
}

```

This way it can be a lot more readable and maintainable. What do you think?

bhorai-atl: Totally agree, I have found a much simpler and more readable solution

bhorai-atl:
@cabella-dot @marcomura this PR is ready for another round of review. Please take a look when you have a moment

src/lib/jira/constants.ts
bhorai-atl: We can simply rely on the presence of a parent via parentKey prop in the _issue object. Removed

cabella-dot (approved):
looks great! thanks for looking into this

PR #487, opened 2025-06-25
[AXON-511] Issue remains visible in "Assigned JIRA work items" after removing assignee

sspanchenko:
### What Is This Change?

The sidebar's state becomes inconsistent when unassigning issues because it relies on a race condition between the UI update and JQL refresh. If the JQL refresh happens before the unassign API call completes, the issue remains in the sidebar until the next refresh. If the JQL refresh happens after - the issue disappears as expected.

Implemented a validated update approach that first checks the assignee change is possible, then manages the sidebar state directly through SearchJiraHelper instead of relying on JQL refresh timing, and verifies the server state afterwards with retries to ensure consistency.

**Before**: 
https://www.loom.com/share/adee7b24c10c4c43a4a279b58af35709

**After**:
https://www.loom.com/share/ad972b5920f44625b4c9ddbd4be98a14


src/webviews/jiraIssueWebview.ts
```diff
@@ -315,8 +339,12 @@ export class JiraIssueWebview
                             });
                         });
 
-                        commands.executeCommand(Commands.RefreshAssignedWorkItemsExplorer);
```
marcomura, line 318: I think we can just add a delay of 2000 to 5000 ms before calling these, without doing any other change

sspanchenko: Thank you. I've added refresh delay

marcomura: Thanks Serhii. The change looks good and it works!
I've just committed a change to move the delay into the command itself, so the pattern is reusable around the codebase if needed.

marcomura (approved):
Thanks you Serhii.

I just incorporated the delay into the commands themselves to the pattern can be reused across the codebase.

I'm merging this now.
Thanks!

PR #490, opened 2025-06-26
[AXON-450] When I am logged out of jira, I have a badge notification asking for authentication for Jira

sspanchenko:
### What Is This Change?
Login badge check
[video.webm](https://github.com/user-attachments/assets/fb5e7fca-dfbc-4b91-ab9d-5f864a6760c5)
Logout badge check
[video.webm](https://github.com/user-attachments/assets/bfbd4c13-d185-45d9-8ade-ac3de7223033)

### DIFF
--- a/e2e/tests/playwright.spec.ts
+++ b/e2e/tests/playwright.spec.ts
@@ -123,10 +123,6 @@ test('Authenticating with Jira works, and assigned items are displayed', async (
 
 test('When user is not authenticated in Jira, a badge notification prompts for login', async ({ page }) => {
     await page.goto('http://localhost:9988/');
-    await page.getByRole('tab', { name: 'Atlassian' }).click();
-    await page.waitForTimeout(250);
-
-    await page.getByRole('tab', { name: 'Getting Started' }).getByLabel(/close/i).click();
 
     await page.waitForTimeout(2000);
 
--- a/e2e/tests/playwright.spec.ts
+++ b/e2e/tests/playwright.spec.ts
@@ -1,4 +1,5 @@
 import { expect, test } from '@playwright/test';
+import { authenticateWithJira } from 'e2e/helpers';
 
 test("Onboarding flow's navigation among pages works", async ({ page }) => {
     await page.goto('http://localhost:9988/');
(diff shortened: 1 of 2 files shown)


e2e/tests/playwright.spec.ts
```diff
@@ -119,3 +119,66 @@ test('Authenticating with Jira works, and assigned items are displayed', async (
...
+
+    const notificationToast = page.locator('.notifications-toasts.visible');
+
+    await page.getByRole('tab', { name: 'Getting Started' }).getByLabel(/close/i).click();
+
+    // Verify that we are not logged in initially and can see the notification and sidebar item
+    await expect(page.getByRole('treeitem', { name: 'Please login to Jira' })).toBeVisible();
+    await expect(notificationToast.getByRole('button', { name: 'Log in to Jira' })).toBeVisible();
+
+    // Verify that we can login and that the notification and sidebar item are not visible after login
```
sspanchenko, line 136: If we only need to test the badge that prompts the user to log in — the one visible immediately when they open the plugin — then this part of the test is sufficient, and the rest can be removed.
However, if we also want to test the full flow where the user logs in, then logs out, and sees the "you have been logged out" message, we can keep the remaining part of the test.
Just let me know what you'd prefer.

e2e/tests/playwright.spec.ts
```diff
@@ -119,3 +119,66 @@ test('Authenticating with Jira works, and assigned items are displayed', async (
...
+test('When I Logged out of Jira, I have a badge notification asking me to login', async ({ page }) => {
+    await page.goto('http://localhost:9988/');
+    await page.getByRole('tab', { name: 'Atlassian' }).click();
+    await page.waitForTimeout(250);
+
+    const notificationToast = page.locator('.notifications-toasts.visible');
+
+    await page.getByRole('tab', { name: 'Getting Started' }).getByLabel(/close/i).click();
+
+    // Verify that we are not logged in initially and can see the notification and sidebar item
```
sspanchenko: If we only need to test the badge that prompts the user to log in — the one visible immediately when they open the plugin — then this part of the test is sufficient, and the rest can be removed (from line 136).
However, if we also want to test the full flow where the user logs in, then logs out, and sees the "you have been logged out" message, we can keep the remaining part of the test.
Just let me know what you'd prefer.

marcomura: I like to keep the message about being logged out, thanks!
Please check with @apeni who created an helper for performing the login flow that can be re-used here.

sspanchenko: @marcomura sure. I see the helpers haven't been merged yet, so I just added the helper file directly in my PR for now. Also, I split the test into two smaller tests.

PR #492, opened 2025-06-27
AXON-525: fix issue with prefix dropdown

apeni:
### What Is This Change?
✅ Fixed the issue with branch prefix always set to the first one in list ( default one ).  
❓ The reason of the issue: in `handleBranchTypeChange` we set up the branch type and call the `buildBranchNameView` func.  `buildBranchNameView` is in deps list in useEffect, where we again rewrite branch type and set it up to the first value in the list ( initial setup ).

Screenrecord with fixed behaviour: 

https://github.com/user-attachments/assets/2fbf1c72-7613-499c-831a-51c63bfa8299

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


bwieger-atlassian-com (approved):

marcomura (approved):
thanks for this quick fix!!

PR #437, opened 2025-06-03
[Hackathon] Rovo-dev demo in atlascode

marcomura:

PR #494, opened 2025-06-27
Revert "[Hackathon] Rovo-dev demo in atlascode"

marcomura:
Reverts atlassian/atlascode#437

Reverting this to not include it in a hot-fix release.


.env.example
```diff
@@ -20,6 +20,4 @@ ATLASCODE_EXP_OVERRIDES_BOOL="exp1_name=true, exp2_name=false"
 # This is for experiments that controls a string value.
 # Syntax: comma separated settings in the format <name>=<string>
 # NOTE: the string is unquoted, and a comma within the string is not supported.
-ATLASCODE_EXP_OVERRIDES_STRING="exp3_name=something, exp4_name=something else"
-
-ROVODEV_ENABLED="true"
```
sdzh-atlassian, line 25: BTW this probably doesn't need to be in here :D

PR #497, opened 2025-06-27
[RovoDev] Accept and Undo changes

marcomura:
### What Is This Change?

This PR adds the functionality to Accept or Undo changes made by Rovo Dev.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] manual tested

### DIFF
--- a/src/react/atlascode/rovo-dev/common.tsx
+++ b/src/react/atlascode/rovo-dev/common.tsx
@@ -34,6 +34,7 @@ interface OpenFileFunc {
     (filePath: string, tryShowDiff?: boolean, lineRange?: number[]): void;
 }
 
+// TODO unused - should it be cleaned up?
 export const ToolDrawer: React.FC<{
     content: ToolReturnGenericMessage[];
     openFile: OpenFileFunc;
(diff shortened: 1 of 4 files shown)


src/react/atlascode/rovo-dev/common.tsx
```diff
@@ -299,58 +296,50 @@ export const UpdatedFilesComponent: React.FC<{
...
-                                onUndo={(path: string) => onUndo([path])}
-                                onAccept={(path: string) => onAccept([path])}
-                            />
-                        );
-                    })}
+                {modifiedFiles.map((msg, index) => {
+                    return (
+                        <ModifiedFileItem
+                            key={index}
+                            msg={msg}
```
sdzh-atlassian, line 306: small nit - what happens if the filename is too long?

sdzh-atlassian (approved):
Awesome, let's ship it and iterate in follow-up PRs if needed 🥳

src/react/atlascode/rovo-dev/common.tsx
marcomura: The only problem would be the name of the tab for the diff, but I'm not sure if vscode itself handles it differently.
If it's really too long, we can think about ellipsing part of the path.
Let's figure it out once we encounter the issue 😄 

Thanks for raising this!

cabella-dot (approved):
looks great! Also, we decided to go back to the ToolDrawer so will be implementing that again after I do the plan

PR #498, opened 2025-06-27
[Rovo Dev] Auto-adjust code highlighting on changed theme

marcomura:
### What Is This Change?

This PR defines the css styling for code highlighting with the `.vscode-dark` directive to automatically adjust it to a change of light and dark themes.

It doesn't follow the same theme's color, but it simply sets two different set of colors for light and dark.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] manual testing

### DIFF
--- a/src/react/atlascode/rovo-dev/RovoDevCodeHighlighting.css
+++ b/src/react/atlascode/rovo-dev/RovoDevCodeHighlighting.css
@@ -1,0 +1,162 @@
+/* Copied from `node_modules/@speed-highlight/core/dist/themes/visual-studio-dark.css` and
+ * modified to work correctly with vscode theming
+ */
+
+[class*=shj-lang-] {
+    white-space: pre;
+    color: #112;
+    text-shadow: none;
+    box-sizing: border-box;
+    background: #fff;
+    border-radius: 10px;
+    max-width: min(100%, 100vw);
+    margin: 10px 0;
+    padding: 30px 20px;
+    font: 18px/24px Consolas, Courier New, Monaco, Andale Mono, Ubuntu Mono, monospace;
+    box-shadow: 0 0 5px #0001;
+}
+
+.shj-inline {
+    border-radius: 5px;
+    margin: 0;
+    padding: 2px 5px;
+    display: inline-block;
+}
+
+[class*=shj-lang-]::selection {
+    background: #bdf5;
+}
+
+[class*=shj-lang-] ::selection {
+    background: #bdf5;
+}
+
+[class*=shj-lang-] > div {
+    display: flex;
+    overflow: auto;
+}
+
+[class*=shj-lang-] > div :last-child {
+    outline: none;
+    flex: 1;
+}
+
+.shj-numbers {
+    counter-reset: line;
+    padding-left: 5px;
+}
+
+.shj-numbers div {
+    padding-right: 5px;
+}
+
+.shj-numbers div:before {
+    color: #999;
+    content: counter(line);
+    opacity: .5;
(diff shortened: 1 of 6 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -108,36 +107,19 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
             localResourceRoots: [
                 Uri.file(path.join(this._extensionPath, 'build')),
                 Uri.file(path.join(this._extensionPath, 'node_modules', '@vscode', 'codicons', 'dist')),
-                Uri.file(path.join(this._extensionPath, 'node_modules', '@speed-highlight', 'core', 'dist')),
```
sdzh-atlassian, line 111: Are the deletions here intentional?

sdzh-atlassian (dismissed):
Not sure if I can meaningfully review this w/o seeing how it actually looks - but I trust your testing :D
Edit: looking at the build, maybe it was a wrong call :P

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: yup, intentional :) thanks

sdzh-atlassian (approved):

src/react/atlascode/rovo-dev/RovoDevCodeHighlighting.css
```diff
@@ -0,0 +1,162 @@
...
+
+.shj-inline {
+    border-radius: 5px;
+    margin: 0;
+    padding: 2px 5px;
+    display: inline-block;
+}
+
+[class*=shj-lang-]::selection {
+    background: #bdf5;
```
cabella-dot, line 27: do we want to use vscode variables instead of direct colors to be more theme agnostic?

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -1,3 +1,5 @@
+import './RovoDevCodeHighlighting.css';
```
cabella-dot, line 1: Does this mess up the webpack builds?

cabella-dot: If not, would be more readable if we move the styles to proper css?

src/react/atlascode/rovo-dev/RovoDevCodeHighlighting.css
marcomura: unfortunately the vscode variables don't provide the colors we expect, which is why we can't follow the theme :(

cabella-dot: ah ok all good! This is me just making sure but have we tested in a number of different themes?

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: after fixing the webpack splitChunk configuration it doesn't mess it up anymore 😄 
converting the rovo dev styles to proper css is a good idea - let's add a backlog item for that? definitely not urgent

src/react/atlascode/rovo-dev/RovoDevCodeHighlighting.css
marcomura: yeah, I've observed the full set of the variables for at least 5 themes, and they simply don't match the code highlighting of the editor, not even slightly.

PR #505, opened 2025-07-01
[AXON-501] Technical planning in Rovo dev chat

cabella-dot:
### What Is This Change?

Add technical planning response in rovo dev chat.

Screenshot 2025-07-03 at 5 26 39 PM

### How Has This Been Tested?

manual testing

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/react/atlascode/rovo-dev/RovoDevCode.css
```diff
@@ -0,0 +1,299 @@
+/* Copied from `node_modules/@speed-highlight/core/dist/themes/visual-studio-dark.css` and
```
marcomura: Can you keep the style for code lighlighting isolated, just for maintainability?
So when we need to refresh it due to @speed-highlight package updates it's going to be easier.

src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -32,7 +32,7 @@ export interface ToolReturnBashMessage {
 export interface ToolReturnTechnicalPlanMessage {
     tool_name: 'create_technical_plan';
     author: 'ToolReturn';
-    content: TechnicalPlan;
```
marcomura, line 35: did something change in the response? last time I tried it the technical plan was returned already in JSON format, and not contained in a string

cabella-dot: It was breaking for me when I was using the type, it only worked if I explicitly parse it as a string

marcomura: I never managed to trigger the technical planning today, but I trust you that they changed the response :D

Which makes me happy because it means the content type is now consistent with the other responses!

marcomura (dismissed):

marcomura (approved):
it looks good!!

PR #509, opened 2025-07-03
More AI-generated unit tests

marcomura:


sdzh-atlassian (approved):
I trust in our new AI overlords 🤖 
In all seriousness, looking at how close some of these tests are to the implementation, I've sort of started questioning the usefulness of this way of adding coverage 🤔 
That said - it's probably still worth it, since the cost is (hopefully?) low and the benefits are high :D

PR #517, opened 2025-07-07
AXON-18 Creating a JIRA issue should default to the most relevant site

om-ukr:
### What Is This Change?
This PR slightly changes the logic of filling in the site and project fields on the create issue page.
After this change, the site or project where most of the assigned Jira issues originate will be set as the default if the user opens the page for the first time.

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## What's new in 3.8.7
 
+### Features
+- Update logic of selecting the default project and site to the most relevant one for the user on the Creating a JIRA issue page
+
 ### Bug Fixes
 
 - Fixed a bug when transitioning Jira issues from 'Start work', which doesn't refresh the issues panels
--- a/src/views/jira/searchJiraHelper.ts
+++ b/src/views/jira/searchJiraHelper.ts
@@ -64,6 +64,11 @@ export class SearchJiraHelper {
         return undefined;
     }
 
+    static getIssuesPerSite(siteId: string): MinimalORIssueLink<DetailedSiteInfo>[] | undefined {
+        const issues = Object.values(this._searchableIssueMap).flat();
+        return issues.filter((issue) => issue.siteDetails.id === siteId);
+    }
+
     // This method is called when the user clicks on the "Search Jira" button in the Jira Tree View
     private static createIssueQuickPick() {
         searchIssuesEvent(ProductJira).then((e) => {
--- a/src/views/jira/searchJiraHelper.ts
+++ b/src/views/jira/searchJiraHelper.ts
@@ -3,7 +3,7 @@ import { commands, QuickPickItem, window } from 'vscode';
 
 import { searchIssuesEvent } from '../../analytics';
 import { DetailedSiteInfo, ProductJira } from '../../atlclients/authInfo';
-import { Commands } from '../../constants';
+import { AssignedJiraItemsViewId, Commands } from '../../constants';
 import { Container } from '../../container';
 
 interface QuickPickIssue extends QuickPickItem {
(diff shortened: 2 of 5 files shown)


src/views/jira/searchJiraHelper.ts
```diff
@@ -64,6 +64,11 @@ export class SearchJiraHelper {
         return undefined;
     }
 
+    static getIssuesPerSite(siteId: string): MinimalORIssueLink<DetailedSiteInfo>[] | undefined {
```
marcomura: I like this approach, but it should only search for issues assigned to the user.

You can export `AssignedWorkItemsViewProviderId` from jiraAssignedWorkItemsViewProvider.ts and use it to take only the assigned issues from the map

marcomura: While I imagine you have created this method to be more general purpose, I think in this case it's better to simplify and make it `getAssignedIssuesCountPerSite(siteId: string): number`

marcomura: Then, where you consume it, you can do something like:

```
availableSites
    .map(site => [site, getAssignedIssuesCountPerSite(site)])
    .sort((a,b) => b[1]-a[1])[0][0]
```

om-ukr: @marcomura Agree, it will work for the **Site** case 
```TS
private getSiteWithMaxIssues(): DetailedSiteInfo | undefined {
        const availableSites = Container.siteManager.getSitesAvailable(ProductJira);
        const { siteWithMaxIssues } = availableSites.reduce(
            (prev: { numIssues: number; siteWithMaxIssues: DetailedSiteInfo | null }, curr) => {
                const siteIssues = SearchJiraHelper.getIssuesPerSite(curr.id);
                if (siteIssues && siteIssues.length > 0 && siteIssues.length > prev.numIssues) {
                    prev.numIssues = siteIssues.length;
                    prev.siteWithMaxIssues = curr;
                }
                return prev;
            },
            { numIssues: 0, siteWithMaxIssues: null },
        );
        if (siteWithMaxIssues) {
            return siteWithMaxIssues;
        }
        return undefined;
    }
```
For the **Project**, we need to return at least the issue key, to be able to identify the project based on it:
``` TS
private getProjectKeyWithMaxIssues(siteId: string): string | undefined {
        const siteIssues = SearchJiraHelper.getIssuesPerSite(siteId);
        if (siteIssues && siteIssues.length > 0) {
            const issuesNumberPerProjectKey = siteIssues.reduce((prev: Record<string, number>, issue) => {
                const projectKey = issue.key?.split('-')[0];
                if (!prev[projectKey]) {
                    prev[projectKey] = 1;
                    return prev;
                }
                prev[projectKey]++;
                return prev;
            }, {});
            const projectKeyWithMaxIssues = Object.keys(issuesNumberPerProjectKey).reduce(
                (prev: string, curr: string) =>
                    issuesNumberPerProjectKey[curr] > issuesNumberPerProjectKey[prev] ? curr : prev,
            );
            if (projectKeyWithMaxIssues) {
                return projectKeyWithMaxIssues;
            }
        }
        return undefined;
    }
```

om-ukr: Thank you, @marcomura. 
Somehow, I completely missed this provider logic 🙃
Fixed.

src/views/jira/searchJiraHelper.ts
```diff
@@ -64,6 +64,12 @@ export class SearchJiraHelper {
         return undefined;
     }
 
+    static getAssignedIssuesPerSite(siteId: string, dataProviderId: string): MinimalORIssueLink<DetailedSiteInfo>[] {
```
marcomura: if the method is for "assigned" issues, then it should be fixed on retrieving the data for the `dataProviderId` provider. I wouldn't take the dataProviderId as a parameter.

src/views/jira/searchJiraHelper.ts
marcomura: oh got it, makes sense. thanks!

src/views/jira/searchJiraHelper.ts
om-ukr: Oh, you are right, thank you. Fixed.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Features
 
 - Added a `Don't show again` button in the pipeline notification to easily disable this type of notifications
+- Update logic of selecting the default project and site to the most relevant one for the user on the Creating a JIRA issue page
```
marcomura: Sorry, we have released last Thursday so now this needs to be added to 3.8.7 :)

om-ukr: Not a problem at all, updated.

PR #526, opened 2025-07-08
[AXON-587] Inform Rovo Dev of reverted files

marcomura:
### What Is This Change?

When the user undoes one or more files, we should inform Rovo Dev about the reverts within the next prompt.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] manual tests

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -356,6 +356,9 @@ ${message}`;
         const webview = this._webView!;
         await this.executeApiWithErrorHandling(async (client) => {
             await client.reset();
+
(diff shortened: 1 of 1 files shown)


cabella-dot (commented):
Does this not get handled in Rovo Dev when it clears the cache?

cabella-dot (approved):
lgtm

marcomura:
@cabella-dot 
> Does this not get handled in Rovo Dev when it clears the cache?

No, Rovo Dev doesn't know if the source file has been reverted or kept.
Hopefully we are going to get an API for this in the future, but for now we agreed on just adding the context in the following prompt.

cabella-dot (approved):
ship it!

PR #528, opened 2025-07-08
[Rovo Dev] Rename 'Accept' with 'Keep' in code

marcomura:
### What Is This Change?

While the button's label has been already renamed, the name of variables and stuff in the code was not.
This PR fixes the name used in the code.


cabella-dot (approved):
lgtm

PR #533, opened 2025-07-09
[AXON-594] Make rovodev host configurable from environment

sdzh-atlassian:
### What Is This Change?

We can configure the port, but not the host for rovodev at the moment. For some launch configurations, we need that to be different too (e.g. when running in docker). This change allows for that to be done via environment

### How Has This Been Tested?

Manually, ran the extension with a bunch of different values in `ROVODEV_HOST` :)

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 
- [x] ~If Atlassian employee & Bitbucket changes: did you test with DC in mind? [See Instructions](https://www.loom.com/share/71e5d17734a547f68fd6128be6cd760e?sid=835e58a7-1240-498d-b2d7-fa7fdf8ffa36)~ N/A

Recommendations:
- [x] ~Update the CHANGELOG if making a user facing change~ N/A

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -94,6 +94,7 @@ export const rovodevInfo = {
     mappingKey: 'workspacePortMapping',
     envVars: {
         port: 'ROVODEV_PORT',
+        host: 'ROVODEV_HOST',
     },
     portRange: {
         start: 40000,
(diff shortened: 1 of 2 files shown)


sdzh-atlassian:
Pending confirmation that this actually works lol

marcomura (approved):

PR #536, opened 2025-07-10
AXON-514 investigate the ap is being called in the jira view issue page

Migelangel04:
### What Is This Change?

Please give us a brief description of what the proposed change is.
- What I did here was improve the performance of the cloud and server instances of our extension. This was done using caching, HTTP parallelization, and modifying where the api calls are taking place. I also did small tweaks to the jira-metaui-client that is present in the pi-clients latest version. This helps with editing, viewing, and creating jira issues linked to [AXON-444](https://softwareteams.atlassian.net/jira/software/projects/AXON/boards/881?selectedIssue=AXON-444) (this does not close AXON-444 since there are more improvements that can be made for cloud users). 
- No heavy architectural changes were made but I did add performance feature flags to test the performance between the proposed changes and the original code.

As reviewers, we'd really appreciate if you could elaborate on the context of the change.
* If there is an issue related to the change, please make sure to link it!
* If not - please describe the change from a user perspective.
* Is there a user concern the change is addressing that we might not be aware of?

If you're making changes to UI components, or affects UX in other ways - please include before-and-after screenshots 🖼️ or videos (e.g. loom) 🎥
https://www.loom.com/share/0afdd18c286d4467baf2460407897d2f?sid=e0cc7108-9694-4471-b596-8cb90915e0ff
Note: I forgot to show create issue but it has much better load times as well. 

### How Has This Been Tested?

🔧 Did you make sure the proposed change works, before submitting the PR?
If yes, please tell us how!

If you can, and if this is applicable to your change - please don't forget to test your changes with both Cloud and Data Center versions Jira/BB.

In particular, if you're making changes not covered by tests - please describe the manual testing you've done - this would be very helpful!
- I used unit testing and made tweaks to it to accommodate how functions are being called with the performance flag. I also use Charles Web Proxy to analyze http traffic on the new and old performance features (new vs old code).
- Tested on instEnv using Jira DC version 9.4.30 will similar levels of success between DC and cloud for perf improvements.

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

[AXON-444]: https://softwareteams.atlassian.net/browse/AXON-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AXON-274]: https://softwareteams.atlassian.net/browse/AXON-274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/jira/issuesForJql.ts
```diff
@@ -7,6 +10,50 @@ export const MAX_RESULTS = 100;
 
 export async function issuesForJQL(jql: string, site: DetailedSiteInfo): Promise<MinimalIssue<DetailedSiteInfo>[]> {
     const client = await Container.clientManager.jiraClient(site);
+    const performanceEnabled = FeatureFlagClient.checkExperimentValue(Experiments.AtlascodePerformanceExperiment);
+    const startTime = process.hrtime();
```
marcomura: It looks like `process.hrtime` is better than `performance.now` in term of precision.
Good choice. I'll adjust my PR to use `process.hrtime` as well!

src/jira/issuesForJql.ts
```diff
@@ -7,6 +10,50 @@ export const MAX_RESULTS = 100;
...
+    if (performanceEnabled) {
+        const [fields, epicFieldInfo] = await Promise.all([
+            Container.jiraSettingsManager.getMinimalIssueFieldIdsForSite(site),
+            Container.jiraSettingsManager.getEpicFieldsForSite(site),
+        ]);
+        // const epicFieldInfo = await Container.jiraSettingsManager.getEpicFieldsForSite(site);
```
marcomura: remove this - in general it's bad practice to leave commented out code

src/jira/issuesForJql.ts
```diff
@@ -7,6 +10,50 @@ export const MAX_RESULTS = 100;
...
+            .then((event) => {
+                Container.analyticsClient.sendTrackEvent(event);
+            })
+            .catch((error) => {
+                Logger.debug('Failed to send performance analytics for JQL', error);
+            });
+        return issues;
+    }
+
     const fields = await Container.jiraSettingsManager.getMinimalIssueFieldIdsForSite(site);
```
marcomura: you can put this block in an `else { }` and refactor the timing/analytics out for both conditional branches

src/jira/settingsManager.ts
```diff
@@ -42,11 +43,32 @@ const minimalDefaultIssueFields: string[] = [
...
+        if (
+            !this._projectKeyCMetaStore.has(projectKey) ||
+            this._projectKeyCMetaStore.get(projectKey)?.projects.length === 0
+        ) {
+            try {
+                const client = await Container.clientManager.jiraClient(siteInfo);
+                cMeta = await client.getCreateIssueMetadata(projectKey);
+                this._projectKeyCMetaStore.set(projectKey, cMeta);
+            } catch (error) {
+                Logger.error(error, 'cMeta Data not avaliable.');
```
marcomura:
```suggestion
                Logger.error(error, 'Create issue metadata not available.');
```

src/webviews/jiraIssueWebview.ts
```diff
@@ -121,8 +125,12 @@ export class JiraIssueWebview
             if (refetchMinimalIssue) {
                 this._issue = await fetchMinimalIssue(this._issue.key, this._issue.siteDetails);
             }
+            // First Request begins here for issue rendering
+            const startRenderUITime = process.hrtime();
             const editUI: EditIssueUI<DetailedSiteInfo> = await fetchEditIssueUI(this._issue);
-
+            const performanceEnabled = FeatureFlagClient.checkExperimentValue(
```
marcomura: this check should not be done here

src/webviews/jiraIssueWebview.ts
```diff
@@ -136,13 +144,56 @@ export class JiraIssueWebview
             msg.type = 'update';
 
             this.postMessage(msg);
-
+            const endRenderUITime = process.hrtime(startRenderUITime);
```
marcomura: I think we need some utility functions to measure time in milliseconds, and reuse them across the code

src/webviews/jiraIssueWebview.ts
```diff
@@ -136,13 +144,56 @@ export class JiraIssueWebview
...
+            editIssueUIRenderPerformanceEvent(
+                this._issue.siteDetails,
+                this._issue.key,
+                endRenderUITimeMs,
+                performanceEnabled,
+            )
+                .then((event) => {
+                    Container.analyticsClient.sendTrackEvent(event);
+                })
+                .catch((error) => {
```
marcomura: today we don't have a pattern of catching analytics errors, let's not start it now 😃  thanks for thinking about it tho!

src/webviews/jiraIssueWebview.ts
```diff
@@ -136,13 +144,56 @@ export class JiraIssueWebview
...
+            // UI component updates
+            const startIssueFieldsUpdates = process.hrtime();
+            this._componentUpdateCount = 0;
             // call async-able update functions here
-            this.updateEpicChildren();
-            this.updateCurrentUser();
-            this.updateWatchers();
-            this.updateVoters();
-            this.updateRelatedPullRequests();
+            if (performanceEnabled) {
```
marcomura: the experiment check should can be moved here

src/webviews/jiraIssueWebview.ts
```diff
@@ -136,13 +144,56 @@ export class JiraIssueWebview
...
+                this._issue.siteDetails,
+                this._issue.key,
+                endIssueFieldsUpdatesMs,
+                performanceEnabled,
+                this._componentUpdateCount,
+            )
+                .then((event) => {
+                    Container.analyticsClient.sendTrackEvent(event);
+                })
+                .catch((error) => {
```
marcomura: no error checking 😄

src/webviews/jiraIssueWebview.ts
```diff
@@ -59,12 +61,14 @@ export class JiraIssueWebview
     private _issue: MinimalIssue<DetailedSiteInfo>;
     private _editUIData: EditIssueData;
     private _currentUser: User;
+    private _componentUpdateCount: number;
```
marcomura: what's the purpose of this?

src/analytics.ts
```diff
@@ -241,6 +241,86 @@ export async function notificationChangeEvent(
     });
 }
 
+// Jira issue performance events
+export async function issuesForJqlPerformanceEvent(
```
marcomura: let's redesign the perf analytics to be more general-purpose.

something like:
```
function performanceEvent(
  type: string, // e.g., editJiraIssue.ttr
  latency: number,
 );
 ```
 
Note that we don't need to add if the experiment is enabled or not, because we will analyze the result of this metric already divided between experiment's cohorts

File: src/webviews/jiraIssueWebview.ts (hunk 10 above)
Comment (Migelangel04) [reply src/webviews/jiraIssueWebview.ts]: To count how many updates we made to the small UI components. There is a total of 5 possible updates we can make: epicChildren, currentUser, watchersInfo, VotersInfo, relatedPRs. I wanted to keep track of how many we did but it looks like it does not work when performance is disabled :(

Comment (Migelangel04) [reply src/webviews/jiraIssueWebview.ts]: I will remove it since I will be simplifying the analytics function.

File: src/analytics.ts [hunk 12]
```diff
@@ -184,6 +184,13 @@ type RovoDevPerfEvents =
     | 'rovodev.response.timeToTechPlan'
     | 'rovodev.response.timeToLastMessage';
 
+type JiraPerfEvents =
```
marcomura, line 187: @bwieger-atlassian-com what do you think about these names for the jira performance events?

src/jira/fetchIssue.ts
```diff
@@ -16,7 +18,23 @@ export async function fetchCreateIssueUI(
     projectKey: string,
 ): Promise<CreateMetaTransformerResult<DetailedSiteInfo>> {
     const client = await Container.clientManager.jiraClient(siteDetails);
-
+    const performanceEnabled = FeatureFlagClient.checkExperimentValue(Experiments.AtlascodePerformanceExperiment);
```
marcomura: nitpick: the value of the `checkExperimentValue` doesn't need to be assigned to a variable, you can just inline the check inside the if()

marcomura (approved):

src/analytics.ts
bwieger-atlassian-com: @Migelangel04

Can we add some structure to this? And let's generalize to different types of things being measured (UI, API, API Flows, etc)

Here's a proposed structure:
`<domain>.<component>.<action>.<detail>`

Also, we would want to align to metric naming that is similar to the [core web vitals](https://web.dev/articles/vitals)
I think in all of these cases we can call it **lcp**. 


So that would make one of these look like: 
ui.createJiraIssue.render.lcp
ui.createJiraIssue.refresh.lcp (if we want to track refresh)

marcomura (approved):

PR #547, opened 2025-07-14
AXON-298: x button for priority is useless

ostoliar-atlassian:
### What Is This Change?
Click the "X" button for priority is useless and causes an error. This button does not exist in the cloud version. We can only change the priority, not remove it.

### Before
axon-298-before

### After
axon-298-after

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/webviews/components/issue/AbstractIssueEditorPage.tsx
+++ b/src/webviews/components/issue/AbstractIssueEditorPage.tsx
@@ -201,6 +201,11 @@ export abstract class AbstractIssueEditorPage<
     }
 
     protected isClearableSelect = (field: SelectFieldUI): boolean => {
+        // AXON-298 Never allow clearing priority field
+        if (field.key === 'priority') {
+            return false;
+        }
+
         if (!field.required) {
             return true;
         }


ostoliar-atlassian:
Hi @marcomura ,

Could you please share your feedback regarding this issue? Here is the situation:

On the browser side, the `isClearable` property for the priority field (using the `atlaskit/select` component) is set to false by default. This means we cannot remove the priority; we can only choose one of the available options, regardless of whether the field is required or not.

On the atlascode side, we are adding a "remove priority" button if the fields is not required. This is why the error "priority: Specify the Priority (id or name) in the string format" is appearing.

Are you okay with the proposed solution of not allowing the clearing of the priority field, or should we set the default value to "medium" when someone clicks the remove button? Alternatively, do you have another solution in mind for this case?

Thank you.

marcomura:
I can see in Jira DC 9 that a new issue gets normal priority by default, and the priority can't be removed.
I'd say the same behavior should be followed from Atlascode.

Thanks!

ostoliar-atlassian:
> I can see in Jira DC 9 that a new issue gets normal priority by default, and the priority can't be removed. I'd say the same behavior should be followed from Atlascode.
> 
> Thanks!

Hi @marcomura ,

Could you please provide more details about what you expect from me regarding this issue? Alternatively, are we okay with the current fix, and can it be merged?
Regarding the DC, new issues are assigned a medium priority by default. This priority cannot be removed, and the same behavior is observed on the Atlascode side.

For the cloud as I can see, the logic is as follows:
-If the priority field is not added in the settings on the cloud side, we simply add the priority field on the Atlascode side with a default medium value. This priority cannot be removed or changed.
-If the priority field is added in the project settings on the cloud side, depending on the issue type and cloud settings, we can view and modify the priority field both during issue creation and afterward. The remove button appears for the priority field because we add it on the Atlascode side for fields that are not required.

If we need to rewrite the logic, can we address it in a separate task? This task only requires removing a button that we do not have on the cloud for the priority field.

marcomura (approved):
This fix is good as is, if needed we can address other issues in a separate change.
Thanks!

PR #548, opened 2025-07-14
[AXON-574][Rovo Dev] Detect error mid-response and provide a 'Try again' button

marcomura:
### What Is This Change?

First iteration of the error-handling scenario.

This PR implements the UI for displaying an error message in the chat, which is displayed every time the provider sends an 'error' to the view.

The error can be [retriable](https://english.stackexchange.com/questions/305273/retriable-or-retryable) or not, which will determine if the `Try again` button will be rendered or not:
image

Trying again creates a new prompt where we tell the agent that the previous response was interrupted unexpectedly due to an error, we remind the agent what the previous prompt was, and we ask the agent to continue from where it was interrupted.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/react/atlascode/rovo-dev/common.tsx
+++ b/src/react/atlascode/rovo-dev/common.tsx
@@ -14,6 +14,7 @@ import { RovoDevProviderMessageType } from 'src/rovo-dev/rovoDevWebviewProviderM
 import {
     agentMessageStyles,
     chatMessageStyles,
+    errorMessageStyles,
     inlineMofidyButtonStyles,
     messageContentStyles,
     toolCallArgsStyles,
@@ -25,6 +26,7 @@ import {
     ChatMessage,
     CodeSnippetToChange,
     DefaultMessage,
+    ErrorMessage,
     parseToolReturnMessage,
     TechnicalPlan,
     TechnicalPlanFileToChange,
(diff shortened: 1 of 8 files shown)


src/react/atlascode/rovo-dev/common.tsx
```diff
@@ -203,13 +178,55 @@ const ChatMessageItem: React.FC<{
...
+};
+
+const RetryPromptButton: React.FC<{
+    enabled: boolean;
+    retryAfterError: () => void;
+}> = ({ enabled, retryAfterError }) => {
+    return (
+        <div style={{ marginTop: '12px' }}>
+            <button disabled={!enabled} onClick={retryAfterError}>
+                Try again
```
sdzh-atlassian, line 215: Small suggestion - there's a [refresh icon](https://code.visualstudio.com/api/references/icons-in-labels) that might be appropriate here:
```html
<i className="codicon codicon-refresh"> Try again
```

src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -1,52 +1,63 @@
...
+    | DefaultMessage
+    | ErrorMessage
+    | ToolCallMessage
+    | ToolReturnGenericMessage
+    | ToolReturnGroupedMessage;
 
 export interface DefaultMessage {
     text: string;
-    author: 'User' | 'RovoDev';
+    source: 'User' | 'RovoDev';
```
sdzh-atlassian, line 11: Any reason to do this rename in the scope of this PR?
Not suggesting that we undo this, just curious about why it's here

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -332,6 +336,12 @@ ${message}`;
         }
     }
 
+    private addRetryAfterErrorContextToPrompt(message: string): string {
+        return `<context>The previous response interrupted prematurely because of an error. Continue processing the previous prompt from the point where it was interrupted.
```
sdzh-atlassian, line 340: Does this actually work as intended? I imagine an error in rovodev CLI would interrupt the session, would it not?

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -255,13 +256,15 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
         });
     }
 
-    private processError(error: Error) {
+    private processError(error: Error, retriable: boolean) {
```
sdzh-atlassian: super NIT: `isRetryable` or `canRetry` or something might be better here pls 🙏
Also, let's maybe make this nullable?

src/react/atlascode/rovo-dev/utils.tsx
marcomura: it felt wrong to add 'RovoDevError' as an author 😆

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: It depends on the error. I spoke with Tim and Rovo Dev should preserve session and context after an error thrown by the model.
But re-providing the previous prompt should help Rovo Dev recovering even if it didn't preserve the session.

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: Agree to change the name to `isRetryable`, but not on making it nullable 😆 sorry

Because it's not clear if the default should be "true" or "false", I think it's better for it to be specified every time.

src/react/atlascode/rovo-dev/common.tsx
marcomura: Discussed offline - keeping text only without icons for now, while we wait for the final design

src/react/atlascode/rovo-dev/utils.tsx
sdzh-atlassian: Hahah, understandable :D

sdzh-atlassian (approved):
Ship it! :)

PR #549, opened 2025-07-14
Addressed the bug found in jiraIssueWebview for currentUser reset

Migelangel04:
Looked to see if it affected votes, watchers, or avatarUrl on page but it did not. Now the user is not reset and the url to the image should not disappear.

### What Is This Change?
- Address the bug where the currentUser would be reset to undefined when we navigate to and from a user.

Screenshot 2025-07-14 at 3 43 00 PM
The currentUser would reset make the user undefined and have no avatarURL after navigating back to the issue. Now we do we not have this reset.

### How Has This Been Tested?
- No particular test but now the image does not reset upon reloading of the user.
- This particular value is not used to fetch any data anywhere else in jiraIssueWebview.ts

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Recommendations:

### DIFF
--- a/src/webviews/jiraIssueWebview.ts
+++ b/src/webviews/jiraIssueWebview.ts
@@ -140,7 +140,7 @@ export class JiraIssueWebview
             //     currentBranches.find(b => b.toLowerCase().indexOf(issue.key.toLowerCase()) !== -1) !== undefined;
 
             this._editUIData.recentPullRequests = [];
-            this._editUIData.currentUser = emptyUser;
+            //this._editUIData.currentUser = emptyUser;
 
             const msg = this._editUIData;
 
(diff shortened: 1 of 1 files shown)


src/webviews/jiraIssueWebview.ts
```diff
@@ -127,28 +127,16 @@ export class JiraIssueWebview
...
             this._editUIData.recentPullRequests = [];
-            this._editUIData.currentUser = emptyUser;
 
             const msg = this._editUIData;
 
             msg.type = 'update';
 
             this.postMessage(msg);
 
-            // call async-able update functions here
```
marcomura: keep this please 😆

marcomura (approved):
thanks for fixing this!

PR #557, opened 2025-07-15
[AXON-619, AXON-720][Rovo Dev] Fire 'Chat details expanded' telemetry event

marcomura:
### What Is This Change?

This change implements the telemetry event for recording that the 'Thinking' drawer was expanded by the user.

This change also fixes:
- duplicated telemetry events for the same promptId
- missing telemetry attribute to identify if the session comes from BBY
- sessionId and promptId not properly initialized for `replay`
- missing error handling for exceptions thrown inside postMessage handlers

### How Has This Been Tested?

- [x] `npm run lint`
- [x] `npm run test`
- [x] `manual tests`


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -221,19 +231,35 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
+
         const reader = response.body.getReader();
         const decoder = new TextDecoder();
         const parser = new RovoDevResponseParser();
 
         while (true) {
             const { done, value } = await reader.read();
+
+            if (!ttrStart) {
+                ttrStart = Math.trunc(performance.measure('processChatResponse_begin').duration);
```
marcomura: @Migelangel04 here's how a performance event is implemented:
1. `mark` 'processChatResponse_begin' at the beginning
2. `measure` the time lapsed from 'processChatResponse_begin' and get the duration
3. send the event via analytics

src/analytics.ts
```diff
@@ -189,83 +189,77 @@ interface RovoDevCommonParams {
...
-    sessionId: string,
-    promptId: string,
-    source: 'replay' | 'chat',
-    deepPlanEnabled: boolean,
-) {
-    return trackEvent('rovoDevPromptSent', 'atlascode', {
-        attributes: { sessionId, promptId, source, deepPlanEnabled },
+export function rovoDevNewSessionActionEvent(isBBY: boolean, sessionId: string, isManuallyCreated: boolean) {
+    return trackEvent('rovoDevNewSessionAction', 'atlascode', {
+        attributes: { isBBY, sessionId, isManuallyCreated },
```
sdzh-atlassian: I kind of object to `isBBY` as a permanent event attribute lol. Might be a bit tricky to get for someone who's not in context of this, in like a year :D

That said - whatever, we can sort this out later

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -602,6 +606,12 @@ const RovoDevView: React.FC = () => {
         [keepFiles, setChatStream, postMessage, totalModifiedFiles],
     );
 
+    const onCollapsiblePanelExpanded = useCallback(() => {
```
sdzh-atlassian, line 591: Random question - since postMessage is already a callback, does this need to be a callback too, or can this just be inlined?

sdzh-atlassian (approved):

src/analytics.ts
marcomura: actually I agree with you, and you just reminded me that I want to double check with Andrew if I can spell the word out 😄 I'll check with him before merging

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: Per my understanding, implementing a function with `useCallback` is always recommended when taking dependency from other callbacks or state.

I may be wrong tho

src/analytics.ts
sdzh-atlassian: I don't suggest we do that, rather we can have something like
```
rovodev mode:
    IDE
    special
```
or something ;)

Basically, even for our own purposes, abbreviations in something as long lived as event specs could be tricky to explain and remember down the line

sdzh-atlassian (approved):

src/analytics.ts
marcomura: done :)

sdzh-atlassian (approved):

PR #558, opened 2025-07-16
[AXON-529]: Polish tool calls and tool return styles

cabella-dot:
### What Is This Change?

Changed the tool call and tool return styles to be in line with Boysenberry designs. 
Added tool call/return functionality for bash and grep
Added icons based on tool type

Demo:
https://www.loom.com/share/fe790eec87d6435c94e2872837b73d6a

### How Has This Been Tested?

Added unit tests for ToolCallItem and ToolReturnItem

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/react/atlascode/rovo-dev/tools/ToolCallItem.test.tsx
```diff
@@ -0,0 +1,159 @@
+import { render } from '@testing-library/react';
```
marcomura, line 1: yooo love more react unit tests

marcomura (approved):
let's ship it!!

PR #568, opened 2025-07-18
[AXON-582]: Implement code plan button

cabella-dot:
### What Is This Change?

Added button which allows for user to execute the technical plan generated by the agent

Button persists only if there is an active code plan and does not appear if any other types of responses

Demo:
https://www.loom.com/share/8d9a6f0f8f424a019f34caa4ac67454b?sid=1750132a-6beb-4435-9526-10a7066a7722

### How Has This Been Tested?

Added simple UT + manual tests
Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/react/atlascode/rovo-dev/utils.tsx
+++ b/src/react/atlascode/rovo-dev/utils.tsx
@@ -234,3 +234,5 @@ export const toolNameIconMap: Record<string, string> = {
     delete_file: 'codicon codicon-trash',
     open_files: 'codicon codicon-go-to-file',
 };
+
+export const CODE_PLAN_EXECUTE_PROMPT = 'Execute the code plan that you have created';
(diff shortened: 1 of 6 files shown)


src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -234,3 +234,5 @@ export const toolNameIconMap: Record<string, string> = {
     delete_file: 'codicon codicon-trash',
     open_files: 'codicon codicon-go-to-file',
 };
+
+export const CODE_PLAN_EXECUTE_PROMPT = 'Execute the code plan that you have created';
```
marcomura, line 238: since this is used only in RovoDevView.tsx, you can declare it there without exporting it

marcomura (approved):

src/react/atlascode/rovo-dev/utils.tsx
cabella-dot: That's true! I can fix this in my next reorg pr

PR #575, opened 2025-07-20
[AXON-634][Rovo Dev] Center chat in wide window size

marcomura:
### What Is This Change?

As title states.

### How Has This Been Tested?

- [X] `manual tests`


sdzh-atlassian (approved):
Looks good, but also - do we not want this to be fullwidth?

marcomura:
> Looks good, but also - do we not want this to be fullwidth?

Discusses with Adam - no, there should be a max width, and center it if viewport is bigger

PR #576, opened 2025-07-20
[AXON-671][Rovo Dev] Fix the chat auto-scrolling that sometimes it doesn't scroll through the end

marcomura:
### What Is This Change?

As title states.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`


marcomura:
It was always "broken" because of the asynchronously nature of React.
Sometimes it was scrolling before the DOM for the last render was updated, so sometimes it wasn't scrolled thought the end.

PR #583, opened 2025-07-22
[AXON-676][Rovo Dev] Fixed tool return rendering and wrapping

marcomura:
### What Is This Change?

The issue is that the `tool-return` was trying to obtain the context from the current `pendingToolCall`, and when Rovo Dev runs multiple tools simultaneously it returns multiple `tool-call` before their `tool-return` are returned.

To solve this issue, `tool-return` messages are now returned with their corresponding `tool-call` message attached, so context can be easily retrieved.

This change also adjusts the `tool-return` styles to correctly wrap and not exceed the width of the chat container.
With this change, the horizontal scrollbar in chat is not appearing anymore.

| Before | After |
|---|---|
| image | image |

### How Has This Been Tested?

- [X] `npm run lint`
- [ ] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/react/atlascode/rovo-dev/utils.tsx
+++ b/src/react/atlascode/rovo-dev/utils.tsx
@@ -164,44 +164,29 @@ export function parseToolReturnMessage(rawMsg: ToolReturnGenericMessage): ToolRe
                     });
                 }
             }
-
-            if (resp.length === 0) {
-                // If no matches found, return the raw content
-                resp.push({
-                    content: msg.content,
-                });
-            }
             break;
 
         case 'bash':
             const args = msg.args && JSON.parse(msg.args);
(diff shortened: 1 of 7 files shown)


src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -164,44 +164,29 @@ export function parseToolReturnMessage(rawMsg: ToolReturnGenericMessage): ToolRe
                     });
                 }
             }
-
-            if (resp.length === 0) {
```
marcomura, line 168: I removed this but it's happening a lot - aka, lots of tool-return are not being rendered.

@cabella-dot , can you look into this? `AXON-677`

PR #584, opened 2025-07-22
[AXON-595]: Update chat messaging

cabella-dot:
Updated chat messaging to include collapsible 'thinking' blocks before summary.

Created new ChatHistory component to handle messaging functionality

Note: As there is no way to distinguish when a message stream is "complete" cannot collapse last thinking block until the next prompt.

Demo: https://www.loom.com/share/6f15fe95a3d24bdfa30b373f32dfcef9?sid=184fb5ab-4cce-426a-be80-f19b3e43a9a1
### How Has This Been Tested?

Added UTs and munual tests

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


marcomura:
> Note: As there is no way to distinguish when a message stream is "complete" cannot collapse last thinking block until the next prompt.

You can use the function `setWaitingForPrompt` in RovoDevView.tsx to determine that the stream has ended, as it triggers for both `CompleteMessage` and `ErrorMessage` events.

PR #599, opened 2025-07-24
AXON-604 - reset form fields after a successful form submission and issue creation

sky-ocean-spirits:
### What Is This Change?

**Issue**:
After a successful "Create Jira Issue" form submission and issue creation, the form fields were not being cleared
Reasons for that: the component's state was not being cleared. Besides that, some fields are uncontrolled. It should also be kept in mind that certain fields are pre-filled (e.g., `site`, `project`, `issue type`) and apparently should remain (WDYT?).

**Solution**:
Added logic to save initial values for pre-filled fields and update when the user changes them.
Reset all other fields after a successful form submission and issue creation.
Added a `key` prop to the `form` to handle uncontrolled fields properly, ensuring the form is re-rendered when necessary.

**Recordings**:
_Before:_
https://www.loom.com/share/a4a6ef0c4cb540ccb91514352bb1998d

_After:_
https://www.loom.com/share/71797bda700a4cb7be2afaf67047dde9
https://www.loom.com/share/c9cb5afabf2b48c7a9aecbf3f6176abd (creating a second issue after the fields are cleared)

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:


src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
```diff
@@ -81,6 +85,21 @@ export default class CreateIssuePage extends AbstractIssueEditorPage<Emit, Accep
                 case 'update': {
                     handled = true;
                     const issueData = e as CreateIssueData;
+
```
sspanchenko, line 89: Can we use `const { fieldValues } = issueData;` to avoid repeating `issueData.fieldValues` multiple times?

src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
```diff
@@ -81,6 +85,21 @@ export default class CreateIssuePage extends AbstractIssueEditorPage<Emit, Accep
...
+                    if (issueData.fieldValues) {
+                        if (Object.keys(this.initialFieldValues).length === 0) {
```
sspanchenko: What do you think about fixing nesting here?
```javascript
const isInitialEmpty = Object.keys(this.initialFieldValues).length === 0;
const hasFieldValues = Object.keys(fieldValues).length > 0;

if (isInitialEmpty && hasFieldValues) {
    this.initialFieldValues = { ...fieldValues };
} else if (!isInitialEmpty) {
    .....
}
```

src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
sky-ocean-spirits: Yep, I was thinking about that too 👍  
Updated

src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
sky-ocean-spirits: Not sure if it will fix the nesting significantly, but will likely be better for readability.
Updated

sspanchenko: @sky-ocean-spirits yes, it looks much better, thanks!

sspanchenko (approved):

PR #603, opened 2025-07-24
[AXON-522] Force correct initial state for the prompt context

sdzh-atlassian:
### What Is This Change?

Quick follow-up for the context PR - correctly set initial state of the context bar

When the user opens up the webview, we:
 * fire an event to the extension process
 * look at the open editors and whatnot
 * send an update back right away

Piggy-backing on the `Initialized` event turned out to be too slow 😱 

### How Has This Been Tested?

`m a n u a l l y`

### DIFF
--- a/src/react/atlascode/rovo-dev/rovoDevView.tsx
+++ b/src/react/atlascode/rovo-dev/rovoDevView.tsx
@@ -412,6 +412,13 @@ const RovoDevView: React.FC = () => {
         [sendButtonDisabled, currentState, isDeepPlanCreated, isDeepPlanToggled, postMessage, promptContextCollection],
     );
 
+    // On the first render, get the context update
+    React.useEffect(() => {
+        postMessage?.({
+            type: RovoDevViewResponseType.ForceUserFocusUpdate,
+        });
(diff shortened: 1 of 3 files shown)


src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -412,6 +412,13 @@ const RovoDevView: React.FC = () => {
         [sendButtonDisabled, currentState, isDeepPlanCreated, isDeepPlanToggled, postMessage, promptContextCollection],
     );
 
+    // On the first render, get the context update
+    React.useEffect(() => {
+        postMessage?.({
```
marcomura, line 417: can postMessage be undefined?

cabella-dot (approved):

src/react/atlascode/rovo-dev/rovoDevView.tsx
sdzh-atlassian: Looking at `useMessagingApi`, I honestly can't tell lol. Figured we should be safe but idk

PR #604, opened 2025-07-24
[AXON-522] Quickfix: don't use bad editor tabs as context

sdzh-atlassian:
### What Is This Change?

Some files you can open in vscode editor aren't actually files
With this, things like:
 * welcome page
 * extension webviews
 * literally nothing being open
- will no longer show up in the prompt context :D

### How Has This Been Tested?

`m a n u a l l y`

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -251,24 +251,37 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
     };
 
     private async forceUserFocusUpdate(editor: TextEditor | undefined = window.activeTextEditor, selection?: Range) {
+        if (!this._webView) {
+            return;
+        }
+
         selection = selection || (editor ? editor.selection : undefined);
 
-        const context = editor
-            ? {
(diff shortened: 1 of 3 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -251,24 +251,37 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
+                    invalid: true,
+                },
             });
+            return;
         }
+
+        const fileInfo = this.getOpenFileInfo(editor.document);
+
+        await this._webView.postMessage({
+            type: RovoDevProviderMessageType.UserFocusUpdated,
```
marcomura, line 275: why do we need to send a message with a fully populated object containing an `invalid` field?
can't we send the message with userFocus = undefined ?

sdzh-atlassian: Because it was the shortest way between "here" and "this working as intended"? 😉

In all seriousness, there's also `enabled` field in there which controls if the goes into context or not - which we want to preserve through the editor changes

Happy to redesign all that when we have a bit of a better grip on how the context will work long-term, what can be valid context, etc

PR #610, opened 2025-07-25
[AXON-721][Rovo Dev] Fire perf telemetry events

marcomura:
### What Is This Change?

This change implements the Rovo Dev performance telemetry events, measuring:
- time to respond with the first byte
- time to respond with the first message
- time to respond with a technical plan
- time to respond with the last message

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -173,32 +173,27 @@ export async function featureFlagClientInitializedEvent(
 
 // Perf events
 
+type RovoDevPerfEvents =
+    | 'rovodev.response.timeToFirstByte'
+    | 'rovodev.response.timeToFirstMessage'
+    | 'rovodev.response.timeToTechPlan'
+    | 'rovodev.response.timeToLastMessage';
+
 interface RovoDevCommonParams {
     sessionId: string;
     promptId: string;
 }
 
 export async function performanceEvent(
-    tag: 'rovodev.response.timeToFirstByte',
+    tag: RovoDevPerfEvents,
     measure: number,
     params: RovoDevCommonParams,
 ): Promise<TrackEvent>;
 export async function performanceEvent(
-    tag: 'rovodev.response.timeToFirstMessage',
+    tag: string,
     measure: number,
-    params: RovoDevCommonParams,
-): Promise<TrackEvent>;
-export async function performanceEvent(
-    tag: 'rovodev.response.timeToTechPlan',
-    measure: number,
-    params: RovoDevCommonParams,
-): Promise<TrackEvent>;
-export async function performanceEvent(
-    tag: 'rovodev.response.timeToLastMessage',
-    measure: number,
-    params: RovoDevCommonParams,
-): Promise<TrackEvent>;
-export async function performanceEvent(tag: string, measure: number, params: Record<string, any>): Promise<TrackEvent> {
+    params?: Record<string, any>,
+): Promise<TrackEvent> {
     return trackEvent('performanceEvent', 'atlascode', {
         attributes: { tag, measure, ...(params || {}) },
     });
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -171,6 +171,99 @@ export async function featureFlagClientInitializedEvent(
     });
 }
 
+// Perf events
+
+interface RovoDevCommonParams {
+    sessionId: string;
+    promptId: string;
+}
+
+export async function performanceEvent(
+    tag: 'rovodev.response.timeToFirstByte',
+    measure: number,
+    params: RovoDevCommonParams,
+): Promise<TrackEvent>;
+export async function performanceEvent(
+    tag: 'rovodev.response.timeToFirstMessage',
+    measure: number,
+    params: RovoDevCommonParams,
+): Promise<TrackEvent>;
+export async function performanceEvent(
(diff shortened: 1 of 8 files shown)


src/analytics.ts
```diff
@@ -171,6 +171,99 @@ export async function featureFlagClientInitializedEvent(
...
+    tag: 'rovodev.response.timeToTechPlan',
+    measure: number,
+    params: RovoDevCommonParams,
+): Promise<TrackEvent>;
+export async function performanceEvent(
+    tag: 'rovodev.response.timeToLastMessage',
+    measure: number,
+    params: RovoDevCommonParams,
+): Promise<TrackEvent>;
+export async function performanceEvent(tag: string, measure: number, params: Record<string, any>): Promise<TrackEvent> {
```
sdzh-atlassian: IIRC we don't even have them as an option in our really old analytics client, but these should arguably be [Operational](https://hello.atlassian.net/wiki/spaces/AAPT/pages/134329315/Operational+Events) as opposed to [Track](https://hello.atlassian.net/wiki/spaces/AAPT/pages/134329319/Track+Events) events per our guidelines

I guess let's keep them as is for now for time reasons though - just something to refactor down the line maybe

src/analytics.ts
```diff
@@ -171,6 +171,99 @@ export async function featureFlagClientInitializedEvent(
...
+    stepsCount: number,
+    filesCount: number,
+    questionsCount: number,
+) {
+    return trackEvent('rovoDevTechnicalPlanningShown', 'atlascode', {
+        attributes: { sessionId, stepsCount, filesCount, questionsCount },
+    });
+}
+
+export async function rovoDevFilesSummaryShownEvent(sessionId: string, filesCount: number) {
```
sdzh-atlassian, line 232: Similarly, these are either [UI](https://hello.atlassian.net/wiki/spaces/AAPT/pages/134329336/UI+Events) or [Screen](https://hello.atlassian.net/wiki/spaces/AAPT/pages/134329341/Screen+Events) events

src/analytics.ts
```diff
@@ -171,6 +171,99 @@ export async function featureFlagClientInitializedEvent(
...
 }
 
+// Perf events
+
+interface RovoDevCommonParams {
+    sessionId: string;
+    promptId: string;
+}
+
+export async function performanceEvent(
```
sdzh-atlassian, line 187: Sorry, I might be dumb here - but what is this for? Are we trying to enable completion options for literals? 

This might be an easier option:
```js
export async function performanceEvent(
  tag: 'a' | 'b' | 'c',
  ...
) {}
```

src/util/perf.ts
```diff
@@ -0,0 +1,30 @@
+const NS_PER_SEC = 1e9;
```
sdzh-atlassian, line 1: Can we please make this a class and put it somewhere close to where it's being used? It's a bit out of line with the rest of the code in immediate vicinity

sdzh-atlassian (dismissed):
We're pressed for time, so feel free to ship it if needed :)

src/util/perf.ts
marcomura: TypeScript doesn't have a native support for static classes, so implementing this as a module feels more natural.
I also may be biased because I have a strong functional background 😄

src/analytics.ts
marcomura: Yeah, this is supposed to be a Screen based on the atlassian guidelines

src/analytics.ts
marcomura: Don't have much choice here, do we? 😆

src/analytics.ts
marcomura: I'm keeping a track for now since today we are using Screen events to track entire "panels" rather than elements within a panel.
I understand this is not the right way tho.

Sooner or later I'd like to redo the entire telemetry for the product, and at some point we have to migrate from PE to Axon ownership, so we will revise everything.

src/util/perf.ts
sdzh-atlassian: I figured it would be something like this ;D

My 2 concerns here are:
 * This is kind of out of line with the rest of the code in the immediate surrounding files. It'd be ok if the rest of our code was like this, but it's a bit odd to have a pocket with different approach to module contents and exports
 * This is something in `utils`, but it's only being used in 1 place. IMO, there's an argument for keeping related code together in a big repo like this - so I'd drop this somewhere around `rovodev` maybe?

That said - ultimately this whole thing a big fat nitpick, so feel free to just ignore me and merge anyway ;)

src/analytics.ts
sdzh-atlassian: > Sooner or later I'd like to redo the entire telemetry for the product

Agreed 100%

sdzh-atlassian (approved):

src/util/perf.ts
marcomura: > This is something in utils, but it's only being used in 1 place. IMO, there's an argument for keeping related code together in a big repo like this - so I'd drop this somewhere around rovodev maybe?

It'll be used by Migel to fire performance telemetry events for Jira, too!

PR #615, opened 2025-07-25
[AXON-522] Add annotation instruction w/ keyboard shortcut

sdzh-atlassian:
### What Is This Change?

Add a little annotation showing the prompt and default keybinding to send something to rovodev chat
image

Notably, we can only use the default keybinding because it seems that there's no way to dynamically get it through the VSCode API :(

### How Has This Been Tested?

`m a n u a l l y`

### DIFF
--- a/package.json
+++ b/package.json
@@ -875,6 +875,12 @@
             "type": "object",
             "title": "Atlassian",
             "properties": {
+                "atlascode.rovodev.showKeybinding": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Show the Rovo Dev keybinding annotation in the editor when a selection is present.",
+                    "scope": "window"
+                },
                 "atlascode.rovodev.executablePath": {
                     "type": "string",
                     "description": "The absolute path to the Rovo Dev executable"
--- a/src/container.ts
+++ b/src/container.ts
@@ -31,6 +31,7 @@ import { Logger } from './logger';
 import OnboardingProvider from './onboarding/onboardingProvider';
 import { Pipeline } from './pipelines/model';
 import { RovoDevCodeActionProvider } from './rovo-dev/rovoDevCodeActionProvider';
+import { RovoDevDecorator } from './rovo-dev/rovoDevDecorator';
 import { RovoDevWebviewProvider } from './rovo-dev/rovoDevWebviewProvider';
 import { SiteManager } from './siteManager';
 import { AtlascodeUriHandler, ONBOARDING_URL, SETTINGS_URL } from './uriHandler';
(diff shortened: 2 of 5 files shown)


src/container.ts
```diff
@@ -411,8 +413,8 @@ export class Container {
         return this._onboardingProvider;
     }
 
-    private static _rovodevWebviewProvder: RovoDevWebviewProvider;
-    public static get rovodevWebviewProvder() {
-        return this._rovodevWebviewProvder;
+    private static _rovodevWebviewProvider: RovoDevWebviewProvider;
```
sdzh-atlassian, line 416: Found this typo

package.json
```diff
@@ -875,6 +875,12 @@
             "type": "object",
             "title": "Atlassian",
             "properties": {
+                "atlascode.rovodev.showKeybinding": {
+                    "type": "boolean",
+                    "default": true,
```
sdzh-atlassian, line 880: default to true for now, later we might want to disable it in the _normal_ builds ;)

src/rovo-dev/rovoDevDecorator.ts
```diff
@@ -0,0 +1,69 @@
...
+        window.onDidChangeTextEditorSelection(this.updateKeybindingDecoration, this, []);
+        window.onDidChangeTextEditorVisibleRanges(this.updateKeybindingDecoration, this, []);
+        workspace.onDidChangeConfiguration(this.onConfigChange, this, []);
+    }
+
+    private static buildKeybindingLabel(): string {
+        const isMac = process.platform === 'darwin';
+        const cmdKey = isMac ? '⌘' : 'Ctrl';
+        const altKey = isMac ? '⌥' : 'Alt';
+        return '\u2003'.repeat(10) + `${cmdKey} + ${altKey} + A: send to Rovo Dev`;
```
sdzh-atlassian, line 33: special character for spacing, couldn't find a better way

src/rovo-dev/rovoDevDecorator.ts
```diff
@@ -0,0 +1,69 @@
...
+export class RovoDevDecorator implements Disposable {
+    private enabled: boolean;
+    private keybindingDecoration;
+
+    constructor() {
+        this.enabled = this.getEnabledSetting();
+        this.keybindingDecoration = window.createTextEditorDecorationType({
+            after: {
+                contentText: RovoDevDecorator.buildKeybindingLabel(),
+                color: '#888888',
```
marcomura, line 19: I believe we need to use a vscode color variable here

marcomura (approved):

PR #624, opened 2025-07-26
[AXON-731][Rovo Dev] Tag Rovo Dev error telemetry events

marcomura:
### What Is This Change?

This change adds a tag on error telemetry events classifying where the error is coming from.
In this change, the only category added is "Rovo Dev", but in the future we can have more like Jira, Bitbucket, Notiications, etc.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,17 +1,21 @@
 import { ConfigurationChangeEvent, Event, ExtensionContext, OutputChannel, window } from 'vscode';
 import { EventEmitter } from 'vscode';
 
+import { ErrorProductArea } from './analyticsTypes';
 import { configuration, OutputLevel } from './config/configuration';
 import { extensionOutputChannelName } from './constants';
 import { Container } from './container';
 
-const ConsolePrefix = `[${extensionOutputChannelName}]`;
+function getConsolePrefix(productArea?: string) {
+    return productArea ? `[${extensionOutputChannelName} ${productArea}]` : `[${extensionOutputChannelName}]`;
+}
 
 export type ErrorEvent = {
     error: Error;
     errorMessage?: string;
     capturedBy?: string;
     params?: string[];
+    productArea?: ErrorProductArea;
 };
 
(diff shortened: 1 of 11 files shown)


src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -409,12 +411,12 @@ const RovoDevView: React.FC = () => {
                 case RovoDevProviderMessageType.CreatePRComplete:
                 case RovoDevProviderMessageType.GetCurrentBranchNameComplete:
                     break; // This is handled elsewhere
+
                 default:
+                    // this is never supposed to happen since there aren't other type of messages
```
sdzh-atlassian, line 416: FWIW the `@ts-expect-error` had an added benefit of failing your compilation if there's actually an event type not covered above explicitly, so there's an argument for keeping that. That said, happy to do whatever

marcomura: yeah I was just looking into that right now and you are right. Putting it back

src/logger.ts
```diff
@@ -102,33 +106,59 @@ export class Logger {
...
+    public static error(productArea: ErrorProductArea, ex: Error, errorMessage?: string, ...params: string[]): void;
+    public static error(ex: Error, errorMessage?: string, ...params: string[]): void;
+    public static error(...params: any[]): void {
         const callerName = retrieveCallerName();
-        this.Instance.errorInternal(ex, callerName, errorMessage, ...params);
+
+        // the following code is ugly, but it's the only way to handle a JS/TS method overload where a new parameter
+        // has been added at the beginning of the arg list.
+        // next improvement will be refactoring every Logger.error in the codebase
+        const productArea: ErrorProductArea = params[0] instanceof Error ? undefined : params.shift();
```
sdzh-atlassian, line 125: Yeah this is pretty cursed, we sort of really need to refactor this some time soon

sdzh-atlassian (approved):

PR #632, opened 2025-07-29
[AXON-530] Added general loading state

cabella-dot:
### What Is This Change?

Added general loading state for rovo dev chat when the response is generating, or executing plan.

Demo:
https://www.loom.com/share/24061e39023444ff9043b39bd520035a
### How Has This Been Tested?

manual testing

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
+++ b/src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
@@ -6,7 +6,7 @@ import { DefaultMessage } from '../utils';
 
 export const ChatMessageItem: React.FC<{
     msg: DefaultMessage;
-    index: number;
+    index?: number;
     icon?: React.ReactNode;
 }> = ({ msg, index, icon }) => {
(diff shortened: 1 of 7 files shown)


src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
```diff
@@ -6,7 +6,7 @@ import { DefaultMessage } from '../utils';
 
 export const ChatMessageItem: React.FC<{
     msg: DefaultMessage;
-    index: number;
+    index?: number;
```
marcomura, line 9: I believe we need to kill this key={index} as it's the source of several react warnings

marcomura (approved):
left a comment

src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
cabella-dot: I agree, let's clean this up

PR #633, opened 2025-07-29
[AXON-730] Fix state error when 'replay' and 'pending prompt'

cabella-dot:
### What Is This Change?

When replay would occur, it would not send that the response is generating to the frontend and it would send `CompleteMessage` event to the frontend which changes the state to `Waiting`. This messed up the streaming because we would not group messages if the state is still waiting for prompt. Now when replay is triggered or there is a pending prompt, the state is changes to `GeneratingResponse` so the stream can act correctly 

### How Has This Been Tested?

manual tests

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -61,7 +61,7 @@ export const baseConfigFor = (project: string, testExtension: string): Config =>
                 ? {
                       statements: 71,
                       branches: 64,
-                      functions: 64,
+                      functions: 63,
                       lines: 71,
                   }
                 : /* tsx */ {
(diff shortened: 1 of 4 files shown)


jest.config.ts
```diff
@@ -61,7 +61,7 @@ export const baseConfigFor = (project: string, testExtension: string): Config =>
                 ? {
                       statements: 71,
                       branches: 64,
-                      functions: 64,
+                      functions: 63,
```
marcomura: :(

marcomura (approved):
nice, thanks!!!

PR #639, opened 2025-07-29
Axon-732-new-session-proper-reset

bwieger-atlassian-com:
### What Is This Change?

- New sessions will reset the file changed section along with the file cache 
- Create PR button will also be reset 
- Required moving some of the state management from the view to the provider 

### How Has This Been Tested?
https://www.loom.com/share/97fe59daa72949ab90d7261d1619b6fd

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -34,6 +34,7 @@ import { Container } from '../../src/container';
 import { Logger } from '../../src/logger';
 import { rovodevInfo } from '../constants';
 import { RovoDevViewResponse, RovoDevViewResponseType } from '../react/atlascode/rovo-dev/rovoDevViewMessages';
+import { isCodeChangeTool, parseToolReturnMessage, ToolReturnParseResult } from '../react/atlascode/rovo-dev/utils';
 import { getHtmlForView } from '../webview/common/getHtmlForView';
 import { PerformanceLogger } from './performanceLogger';
 import { RovoDevResponse, RovoDevResponseParser } from './responseParser';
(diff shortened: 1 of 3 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -464,6 +466,23 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
                         questionsCount,
                     ).then((evt) => Container.analyticsClient.sendTrackEvent(evt));
                 }
+
+                // Update provider-managed state for code change tools
+                if (isCodeChangeTool(response.tool_name)) {
+                    const toolReturnMessage = {
+                        source: 'ToolReturn' as const,
```
bwieger-atlassian-com, line 473: This object creation needs to be done because the parseToolReturnMessage function expects a type that has an attribute of source

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -837,14 +859,15 @@ ${message}`;
         const promises = filePaths.map(async (filePath) => {
             const resolvedPath = this.makeRelativePathAbsolute(filePath);
             const cachedFilePath = await this.rovoDevApiClient!.getCacheFilePath(filePath);
-            await this.getPromise((callback) => fs.rm(resolvedPath, callback));
+            await this.getPromise((callback) => fs.rm(resolvedPath, { force: true }, callback));
```
bwieger-atlassian-com, line 862: There's a case when the file is deleted that the file will not exist in the system. Without the force, it throws an error.

marcomura:
Why not simply clearing the modified files in the view like this?
image

bwieger-atlassian-com:
> Why not simply clearing the modified files in the view like this? image

Because the file cache also needs to be updated. 

Otherwise, when you view changes made by the new session, it will show both the old and the new session's changes.

PR #661, opened 2025-07-31
Disable Rovo Dev process management in BBY

marcomura:
### What Is This Change?

The process management is used to manually execute the binary.
It is not meant to be executed in BBY, therefore this change is gating it behind the ROVODEV_BBY flag.

### How Has This Been Tested?

- [x] `npm run lint`
- [x] `npm run test`
- [x] `manual tests`

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -115,7 +115,7 @@ export async function activate(context: ExtensionContext) {
         }
     }
 
-    if (process.env.ROVODEV_BBY && !!process.env.ROVODEV_ENABLED) {
+    if (!!process.env.ROVODEV_BBY && !!process.env.ROVODEV_ENABLED) {
         commands.executeCommand('workbench.view.extension.atlascode-rovo-dev');
     }
 


src/extension.ts
```diff
@@ -107,23 +109,18 @@ export async function activate(context: ExtensionContext) {
...
 
-        if (process.env.ROVODEV_BBY) {
-            commands.executeCommand('workbench.view.extension.atlascode-rovo-dev');
+        if (!!process.env.ROVODEV_ENABLED) {
+            initializeRovoDevProcessManager(context);
         }
     }
 
-    const duration = process.hrtime(start);
+    if (process.env.ROVODEV_BBY && !!process.env.ROVODEV_ENABLED) {
```
sdzh-atlassian: This looks a bit sus to me - we're treating the env vars differently, is that intended?

marcomura: do you mean the lack of `!!` ?

src/extension.ts
```diff
@@ -25,12 +25,14 @@ import { registerResources } from './resources';
 import { deactivateRovoDevProcessManager, initializeRovoDevProcessManager } from './rovo-dev/rovoDevProcessManager';
 import { GitExtension } from './typings/git';
 import { Experiments, FeatureFlagClient, Features } from './util/featureFlags';
+import Performance from './util/perf';
 import { NotificationManagerImpl } from './views/notifications/notificationManager';
 
 const AnalyticDelay = 5000;
+const PerfStartMarker = 'extension.start';
```
sdzh-atlassian, line 32: certainly not for this PR, but one of these days we should standardize the capitalization on things like constants :D

Image

src/extension.ts
sdzh-atlassian: yep! Don't think an action is needed on this necessarily - the difference in how we treat these similar env vars just jumped at me

PR #685, opened 2025-08-05
AXON-537 added e2e test for pullrequest creation

mbohoslavskyi:
### What Is This Change?

Created e2e test for Pullrequest creation;
added fixture to delete workspace after test, so added repository won't be saved for the next test;
moved files with mocked data to separate folder;
mocked several endpoints;

### Demo
[video.webm](https://github.com/user-attachments/assets/9c67c981-0082-4664-843a-d058bdd85b11)

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


om-ukr (dismissed):

bhorai-atl (dismissed):

marcomura (dismissed):
nice!! thanks!

BHulovatyi (dismissed):

e2e/wiremock-mappings/bitbucket/pullrequest/get-pullrequest.json
```diff
@@ -0,0 +1,194 @@
+{
+    "priority": 2,
```
apeni: Just curious - why do we need priority here and in mock above, if these files mock different url ( with query param and without ). Tested it several times without priority field and it seems to work fine

e2e/wiremock-mappings/bitbucket/pullrequest/get-commits.json
```diff
@@ -0,0 +1,103 @@
+{
```
apeni: nit: Just checked that all mocks in this folder ( except `get-pullrequest` and `get-pullrequest--fields` ) are optional. We handle it in code and in case of getting errors from these requests - we use default data from pullrequest mock, so as we don't test the functionality related to these mocks ( comments, statuses, tasks, etc ) we can remove the mocks and use just necessary for test, it will reduce the amount of mock files.  
But it is just recommendation, up to you. Probably these requests can be usefull for future tests.

e2e/wiremock-mappings/bitbucket/pullrequest/get-pullrequest.json
mbohoslavskyi: Because otherwise inner job will be failed. Yes, test runs well without priorities because we don't check if there are no alerts on the page. We just check if PR page appears. But for correct work of the next tests these priorities should remain.
You can check mentioned alerts if you run docker serve (test:e2e:docker:serve) and follow the script.

e2e/wiremock-mappings/bitbucket/pullrequest/get-commits.json
mbohoslavskyi: I think these mocks will be useful for future tests and now PR view page has no errors as it should be.

bwieger-atlassian-com:
I looked through the video you posted. After the PR is created, it is not showing up in the side panel yet. This (PR showing in side panel) happens on my local with the current HEAD of main.  

Can we have the test check for this functionality?

PR #693, opened 2025-08-06
AXON-797: Added E2E test to authenticate with Jira DC

BHulovatyi:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


e2e/compose.yml
```diff
@@ -19,6 +19,22 @@ services:
         expose:
             - 443
 
+    wiremock-jira-dc:
```
apeni: We don't need a separate service for jira-dc. Currently we have 2 services: one for jira ( but without dc domain ) and one for bitbucket ( with cloud and dc domains ) to separate logic between the products. So, adding the dc domain to wiremock-mockedteams aliases should be enough

e2e/sslcerts/generate-certs.sh
```diff
@@ -85,5 +85,8 @@ generate_service_cert "wiremock-mockedteams" "mockedteams.atlassian.net"
 # # Generates a certificate for bitbucket.mockeddomain.com
 generate_service_cert "wiremock-bitbucket" "bitbucket.mockeddomain.com"
 
+# Generates a certificate for jira.mockeddomain.com
+generate_service_cert "wiremock-jira-dc" "jira.mockeddomain.com"
```
apeni: The same comment as in `compose.yml`, we can just add the dc domain to existing certificate for jira

e2e/wiremock-mappings/jira-dc/BTS-1/removeassignee.json
```diff
@@ -0,0 +1,23 @@
+{
```
apeni: This file is a duplicate for one that we have in mockedteams. Having one service for dc and cloud ( comment in `compose.yml`) should reduce the amount of duplicated mocks as all jira related mocks will be in one folder - `mockedteams`

e2e/wiremock-mappings/jira-dc/issueLinkType.json
```diff
@@ -0,0 +1,13 @@
+{
```
apeni, line 1: This is also a duplicate, but with different domain, so instead of duplication we can use wiremock admin api to update the body. The same for `myself.json`, `project.json` and `search-project.json`

e2e/wiremock-mappings/jira-dc/mypermissions.json
```diff
@@ -0,0 +1,21 @@
+{
```
apeni: Duplicate for one in wiremockteams. Can be removed as well

e2e/compose.yml
BHulovatyi: Done

e2e/wiremock-mappings/jira-dc/BTS-1/removeassignee.json
BHulovatyi: Make sense

bwieger-atlassian-com (approved):

PR #697, opened 2025-08-06
[WIP] AXON-762: Migrate frontend logic to new Start Work layout

sspanchenko:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


bwieger-atlassian-com:
Can we confirm that the analytics events that were fired on the old page are fired in the new page?

sspanchenko:
MUI v5 is used in this PR 😺

sspanchenko:
> Can we confirm that the analytics events that were fired on the old page are fired in the new page?
@bwieger-atlassian-com 

Looking at the implementation, both pages use the same `StartWorkWebviewController` which handles all analytics events. This controller fires these events:

1. Main work started event:
```typescript
fireIssueWorkStartedEvent(issue.siteDetails, pushBranchToRemote)
```

2. Common events through commonHandler:
- SendAnalytics
- OpenJiraIssue  
- ExternalLink
- and others

Since both pages use this controller through `useStartWorkController` hook, we can be confident that the same analytics events are fired with the same parameters.

The only difference is the view identifier (changed from `StartWorkPage` to `StartWorkPageV3`) to distinguish between versions in analytics data.

I've added console logs (locally) in the controller before events are fired and verified that correct data is being sent.

PR #701, opened 2025-08-07
Rovo Dev: Hide context on replay

bwieger-atlassian-com:
### What Is This Change?

- Bug: When replaying messages that had <context> on them, that would show in the UI. 
- Fix: when inspecting the user message, if the <context> tags are there, remove them and the content within them 

### How Has This Been Tested?
Manual. The regex was tested with the help of Copilot on several different cases. 

### Known issue:
- The pretty "file tokens" don't show up on the replay. 

Screenshot 2025-08-06 at 5 20 12 PM
Screenshot 2025-08-06 at 5 19 16 PM


cabella-dot (approved):
looks good!

marcomura (approved):

PR #708, opened 2025-08-08
Refactor-rovodev-processManager

bwieger-atlassian-com:
### What Is This Change?
- A refactor. No logic changes 
- Makes the rovoDevProcess manager a class. 

### How Has This Been Tested?
- Manually verified that changing workspaces and closing them out worked as expected

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,7 +26,7 @@ import {
     BB_PIPELINES_FILENAME,
 } from './pipelines/yaml/pipelinesYamlHelper';
 import { registerResources } from './resources';
-import { deactivateRovoDevProcessManager, initializeRovoDevProcessManager } from './rovo-dev/rovoDevProcessManager';
+import { RovoDevProcessManager } from './rovo-dev/rovoDevProcessManager';
 import { GitExtension } from './typings/git';
 import { Experiments, FeatureFlagClient, Features } from './util/featureFlags';
 import Performance from './util/perf';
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -222,7 +222,7 @@ async function sendAnalytics(version: string, globalState: Memento) {
 export function deactivate() {
     if (!process.env.ROVODEV_BBY) {
         if (!!process.env.ROVODEV_ENABLED) {
-            deactivateRovoDevProcessManager();
+            RovoDevProcessManager.getInstance().deactivateRovoDevProcessManager();
```
sdzh-atlassian, line 225: I think we might not be using the `Disposable` mechanism properly 🤔
Might not be for this PR, but there's an argument to make `RovoDevProcessManager` extend `Disposable`, and add it to context the way we add other things

src/extension.ts
```diff
@@ -26,7 +26,7 @@ import {
     BB_PIPELINES_FILENAME,
 } from './pipelines/yaml/pipelinesYamlHelper';
 import { registerResources } from './resources';
-import { deactivateRovoDevProcessManager, initializeRovoDevProcessManager } from './rovo-dev/rovoDevProcessManager';
+import { RovoDevProcessManager } from './rovo-dev/rovoDevProcessManager';
```
sdzh-atlassian, line 29: Another thing, might be out of scope - but this should probably live in `container.ts`, right? That's basically our "store everything" class :D

src/extension.ts
bwieger-atlassian-com: Agreed. In the next PR

src/extension.ts
bwieger-atlassian-com: Will do on next PR

PR #711, opened 2025-08-08
Revert "Refactor-rovodev-processManager #708"

marcomura:
This reverts #708 .

Unfortunately, I have to revert this since I'm doing a big change on the process manager to implement the auto-downloader of the Rovo Dev binaries, and the conflicts will be huge.

This refactoring will be restored afterwards.


sdzh-atlassian (approved):
I guess the relevant reviewer here is @bwieger-atlassian-com, but in case you need an approval quickly there we go :D

bwieger-atlassian-com (approved):

PR #715, opened 2025-08-09
Automatically enable Rovo Dev for internal users

marcomura:
### What Is This Change?

This change queries the endpoint http://github.internal.atlassian.com/ to detect if the user is an internal user.

After detection, this change automatically enables Rovo Dev for internal users.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -11,6 +11,7 @@ import { BitbucketCheckoutHelper } from './bitbucket/checkoutHelper';
 import { CheckoutHelper } from './bitbucket/interfaces';
 import { PullRequest, WorkspaceRepo } from './bitbucket/model';
 import { BitbucketCloudPullRequestLinkProvider } from './bitbucket/terminal-link/createPrLinkProvider';
+import { CommandContext, setCommandContext } from './commandContext';
 import { openPullRequest } from './commands/bitbucket/pullRequest';
 import { configuration, IConfig } from './config/configuration';
 import { Commands } from './constants';
@@ -78,6 +79,15 @@ export class Container {
     private static _assignedWorkItemsView: AssignedWorkItemsViewProvider;
 
     static async initialize(context: ExtensionContext, version: string) {
+        canFetchInternalUrl().then((success) => {
+            this._isInternalUser = success;
+            this._isRovoDevEnabled = success;
+
+            if (success) {
+                this.enableRovoDev(context);
+            }
+        });
+
         const analyticsEnv: string = this.isDebugging ? 'staging' : 'prod';
 
         this._analyticsClient = analyticsClient({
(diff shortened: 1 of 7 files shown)


src/container.ts
```diff
@@ -476,3 +500,17 @@ export class Container {
         return this._rovodevWebviewProvider;
     }
 }
+
+async function canFetchInternalUrl(): Promise<boolean> {
```
sdzh-atlassian, line 504: For the record, I still think this shouldn't be here - but happy to keep it around for now if needed

src/container.ts
```diff
@@ -336,6 +350,16 @@ export class Container {
         this._context.globalState.update(ConfigTargetKey, target);
     }
 
+    private static _isInternalUser: boolean;
+    public static get isInternalUser() {
+        return this._isInternalUser;
+    }
+
+    private static _isRovoDevEnabled: boolean;
```
sdzh-atlassian, line 358: should probably default to `false`, just in case?

src/container.ts
```diff
@@ -336,6 +350,16 @@ export class Container {
         this._context.globalState.update(ConfigTargetKey, target);
     }
 
+    private static _isInternalUser: boolean;
```
sdzh-atlassian: same here, let's be prudent and set it to `false` by default?

src/container.ts
bwieger-atlassian-com: What's the reason this is needed?

src/container.ts
```diff
@@ -78,6 +79,10 @@ export class Container {
     private static _assignedWorkItemsView: AssignedWorkItemsViewProvider;
 
     static async initialize(context: ExtensionContext, version: string) {
+        canFetchInternalUrl().then((success) => {
+            this._isInternalUser = success;
```
bwieger-atlassian-com: I don't think we use this. Am I missing something?

src/container.ts
marcomura: well, undefined is falsy :)

src/container.ts
marcomura: same as the other, undefined is falsy :)

src/container.ts
marcomura: I can move it to a separate PR - the idea is writing into telemetry if the user is an internal user, just for analysis

src/container.ts
marcomura: The idea is writing into telemetry if the user is an internal user, just for analysis. I haven't piped it into telemetry yet.
I can remove it and do it as a separate PR.

bwieger-atlassian-com: Cool, let's maybe position this as _isAtlassianNetwork

src/container.ts
bwieger-atlassian-com: Again _isAtlassianNetwork

src/container.ts
marcomura: I'm actually keeping three states `undefined`, `false`, and `true` intentionally to know if a telemetry event sent too early was sent before internal user detection

src/container.ts
sdzh-atlassian: Sure, but isn't this being `undefined` some kind of an invalid state, conceptually?

To me it's a bit of a red flag TBH, like something went wrong and we didn't account for it properly - but this is 100% something we can address down the line, so feel free to merge :D 

Not to mention, I just might be completely wrong about this lol

sdzh-atlassian (approved):
All good!

bwieger-atlassian-com (approved):

sdzh-atlassian (approved):

PR #736, opened 2025-08-14
[AXON-847] Rovo Dev: Improvements in tcp ports and process management

marcomura:
#### This PR changes the following things:

1. **TCP port**

Currently, the ports are assigned to workspaces using a globalState map, creating an association that was kept forever.
This has the main problem that ports were not checked for availability, and the process will fail if the port is already in use.

With this change, Rovo Dev will attempt to start the service on the first available port of the configured range.

2. **Process management**

Currently, there is a very rudimental logic to handle multiple folder in a workspace, spawning a Rovo Dev process for each of the folder.

With this change, we remove the support for multiple folders, and we spawn a single Rovo Dev instance for the first folder being opened by the workspace.

In the future we may want Rovo Dev to natively support multiple folders in a single instance.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/rovoDevProcessManager.ts
+++ b/src/rovo-dev/rovoDevProcessManager.ts
@@ -168,8 +168,8 @@ async function getOrAssignPortForWorkspace(): Promise<number> {
     const portEnd = rovodevInfo.portRange.end;
 
     for (let port = portStart; port <= portEnd; ++port) {
-        if (await isPortAvailable(8899)) {
-            return 8899;
+        if (await isPortAvailable(port)) {
+            return port;
         }
     }
 
--- a/src/rovo-dev/rovoDevProcessManager.ts
+++ b/src/rovo-dev/rovoDevProcessManager.ts
@@ -1,6 +1,7 @@
 import { ChildProcess, spawn } from 'child_process';
 import { access, constants } from 'fs';
 import fs from 'fs';
+import net from 'net';
 import path from 'path';
 import { downloadAndUnzip } from 'src/util/downloadFile';
 import { getFsPromise } from 'src/util/fsPromises';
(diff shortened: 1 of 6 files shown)


src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -44,253 +45,261 @@ function GetRovoDevURIs(context: ExtensionContext) {
...
-        rovoDevWebviewProvider.sendErrorToChat(
-            `Rovo Dev is not supported for the following platform/architecture: ${process.platform}/${process.arch}`,
-        );
-        return;
-    }
+/**
+ * Placeholder implementation for Rovo Dev CLI credential storage
+ */
+async function getCloudCredentials(): Promise<{ username: string; key: string; host: string } | undefined> {
+    const sites = Container.siteManager.getSitesAvailable(ProductJira);
```
bwieger-atlassian-com, line 75: You might want to look at getAllValidAuthInfo in the credentialManager class. 

It doesn't do quite what you want, but it's close. 

We could update the method to a new signature 


`getAllValidAuthInfo(product: Product, instanceType: InstanceType)`

And instanceType can be `cloud|server`

marcomura: Actually I haven't touched this method at all, it got flagged only because I moved it around during the refactoring 😄 

https://github.com/search?q=repo%3Aatlassian%2Fatlascode%20getCloudCredentials&type=code

src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -44,253 +45,261 @@ function GetRovoDevURIs(context: ExtensionContext) {
...
-let rovoDevWebviewProvider: RovoDevWebviewProvider;
+        server.once('listening', () => {
+            server.close(() => {
+                resolve(true);
+            });
+        });
 
-export function setRovoDevWebviewProvider(provider: any) {
-    rovoDevWebviewProvider = provider;
+        server.listen(port, '127.0.0.1');
```
bwieger-atlassian-com, line 67: Why do we listen here?

marcomura: Rovo Dev binds to 127.0.0.1

marcomura: ah I think I've misunderstood the question. This logic is to determine if the port is available or already in use.

bwieger-atlassian-com: Oh, I get it now. Thanks!

src/rovo-dev/rovoDevProcessManager.ts
bwieger-atlassian-com: Oh, then nvm

bwieger-atlassian-com (approved):

PR #738, opened 2025-08-14
[AXON-856] fix message breaking on error retry

cabella-dot:
### What Is This Change?

in `rovoDevWebviewProvider` added `_chatBusy` so we do not call the chat endpoint while there is a current message. This led to some weird functionality

in `rovoDevView` added fallback state changes for worst case handling

### How Has This Been Tested?

manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -93,6 +93,7 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
     private _disabled = false;
     private _initialized = false;
 
+    private _chatBusy = false;
     private _chatSessionId: string = '';
     private _currentPromptId: string = '';
     private _currentPrompt: RovoDevPrompt | undefined;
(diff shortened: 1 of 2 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -465,6 +466,7 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
     }
 
     private completeChatResponse(sourceApi: 'replay' | 'chat' | 'error') {
+        this._chatBusy = false;
```
bwieger-atlassian-com, line 469: I'm wondering how this looks when we think about the /v3 apis

bwieger-atlassian-com (approved):

src/rovo-dev/rovoDevWebviewProvider.ts
cabella-dot: yeah hopefully it will be able to detect if the chat is busy or not

PR #740, opened 2025-08-15
[AXON-849] Hook rovodev visibility to a feature flag

sdzh-atlassian:
### What Is This Change?

Hook rovodev visibility to a feature flag, using `tenantId` as the criterion

Disabling logic here is working but insufficient - will expand in a follow-up PR

### How Has This Been Tested?

Manually - see loom in #axon

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -180,7 +180,19 @@ export class Container {
                 analyticsClient: this._analyticsClient,
                 identifiers: {
                     analyticsAnonymousId: this.machineId,
+                    tenantId: this._siteManager.primarySite?.id,
                 },
+            });
+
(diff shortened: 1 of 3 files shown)


src/container.ts
```diff
@@ -226,6 +238,10 @@ export class Container {
         this.configureRovodevSettingsCommands(context);
     }
 
+    static disableRovoDev(context: ExtensionContext) {
```
sdzh-atlassian, line 241: This is not good enough long term, will improve the enable/disable logic in a follow-up

bwieger-atlassian-com (approved):

PR #741, opened 2025-08-15
[AXON-855] new session cancels current response then creates new session

cabella-dot:
### What Is This Change?

Before, when we trigger new session we hit the reset endpoint which does not cleanly cancel the response and does not create a proper new session

Now, we hit the cancel endpoint if there is a current response and hit new `/v2/sessions/create` endpoint to create new session

FYI: `/v2/sessions/create` does not work if there is a current response so we need to cancel first

### How Has This Been Tested?

manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -849,6 +860,7 @@ ${message}`;
         await this.sendPromptSentToView({ text: '', enable_deep_plan: false, context: undefined });
 
         await this.executeApiWithErrorHandling(async (client) => {
+            this._chatBusy = true;
```
marcomura, line 864: where is this reset to false?

cabella-dot: It happens in completeChatResponse: 
``` ts
private completeChatResponse(sourceApi: 'replay' | 'chat' | 'error') {
        this._chatBusy = false;
        const webview = this._webView!;
        return webview.postMessage({
            type: RovoDevProviderMessageType.CompleteMessage,
            isReplay: sourceApi === 'replay',
        });
    }
```

cabella-dot: and this is triggered at the end of `processChatMessage`
if there is a failure during the response, it will reset too

marcomura (dismissed):

PR #746, opened 2025-08-15
Add debug-mode only commands to speed up development

sdzh-atlassian:
### What Is This Change?

Debug-only commands to log in and out of sites. No effect on the properly distributed version of the extension

### How Has This Been Tested?

Manually - see loom in #axon

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/package.json
+++ b/package.json
@@ -456,6 +456,21 @@
                 "command": "atlascode.showOnboardingFlow",
                 "title": "Show Onboarding Flow",
                 "category": "Atlassian"
+            },
+            {
+                "command": "atlascode.debug.quickCommand",
+                "title": "[DEBUG] Atlascode: Quick command",
+                "enablement": "atlascode:debugMode"
+            },
+            {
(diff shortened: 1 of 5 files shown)


marcomura:
I see how these commands were (are?) useful to test the new logic for enabling Rovo Dev based on credentials, but after that PR is tested and merged, why do we still need them?

sdzh-atlassian:
> I see how these commands were (are?) useful to test the new logic for enabling Rovo Dev based on credentials, but after that PR is tested and merged, why do we still need them?

Basically, I want to never ever have to type `https://something-or-another.atlassian.net` and my API key into the login form :D 

Also, now that we have internal rollout capacity - I assume it would be fairly useful to swap between internal and external credentials to check behaviours in different circumstances

bwieger-atlassian-com (approved):

marcomura (approved):

PR #747, opened 2025-08-15
[AXON-849] Implement proper cleanup at credential change

sdzh-atlassian:
### What Is This Change?

Added missing bits for Rovodev cleanup that happens when the functionality gets toggled (e.g. on settings change or authentication).

As a general rule - we now store a `Disposable` container with all rovodev-related stuff, and `dispose`/re-create it when needed.

There are also some safeguards now around command declaration and implementation that check against `Container.isRovoDevEnabled`

The one thing notably missing here is the non-rovodev webview integrations - will catch them up in a follow-up PR

### How Has This Been Tested?

Manually, through exhaustive use of the new debug commands :)

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,15 +95,18 @@
         "commands": [
             {
                 "command": "atlascode.rovodev.askInteractive",
-                "title": "Rovo Dev: Quick Ask"
+                "title": "Rovo Dev: Quick Ask",
+                "enablement": "atlascode:rovoDevEnabled"
             },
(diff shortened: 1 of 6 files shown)


bwieger-atlassian-com (approved):

marcomura (approved):
🚀

PR #776, opened 2025-08-20
Rework feature flag client, implement retries

sdzh-atlassian:
### What Is This Change?

This is - hopefully - a fix for the intermittent networking errors in feature flag client that we've been observing today.
We now have a retry strategy for FF client initialization, with full reset of the instance from the FG library

When does it occur:
 * `FeatureFlagClient.initialize()`
 * `FeatureFlagClient.updateUser()`

What is not covered here:
 * Long-term, we might want a proper backoff strategy to re-initialize FF client asynchronously if it ever fails (e.g. if the user is off-network at init time)
 * Any kind of a broader rework of the FFClient - which is kind of badly needed IMO. This was as close to a drop-in replacement as I could get it

### How Has This Been Tested?

Manually - although the error being addressed here is delightfully elusive and inconsistent 😢

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


sdzh-atlassian:
Rebased to `main` just now after some `package.lock` conflicts

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -62,54 +69,93 @@ export abstract class FeatureFlagClient {
...
-            ...this.identifiers,
+        this.options.identifiers = {
+            ...this.options.identifiers,
             tenantId,
         };
 
         const clientOptions = await this.buildClientOptions();
 
         try {
-            await FeatureGates.updateUser(clientOptions, this.identifiers);
```
marcomura, line 109: what does it happen when updateUser fails? does the client get busted, or the client still works without any user update?

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -62,54 +69,93 @@ export abstract class FeatureFlagClient {
...
 
         Logger.debug(
             `FeatureGates: initializing, target: ${clientOptions.targetApp}, environment: ${clientOptions.environment}`,
         );
-
         try {
-            await FeatureGates.initialize(clientOptions, options.identifiers);
+            this.client = await this.initializeWithRetry(clientOptions, options.identifiers);
         } catch (err) {
             return Promise.reject(new FeatureFlagClientInitError(ClientInitializedErrorType.Failed, err));
```
marcomura, line 119: considering that `initializeWithRetry` is already throwing a FeatureFlagClientInitError, you can avoid the try/catch here

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -62,54 +69,93 @@ export abstract class FeatureFlagClient {
...
+            } catch (err) {
+                if (i < maxRetries - 1) {
+                    Logger.info(
+                        `FeatureFlagClient: Retrying reinitialization (${maxRetries - i - 1} retries left). Reason: ${err}`,
+                    );
+                }
+            }
+        }
+        return Promise.reject(
+            new FeatureFlagClientInitError(ClientInitializedErrorType.Failed, 'FeatureFlagClient: Max retries reached'),
```
marcomura, line 97: can you save the actual error and append it here?

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -62,54 +69,93 @@ export abstract class FeatureFlagClient {
...
+            return;
         } catch (e) {
             Logger.error(new Error(`FeatureFlagClient: Failed to update user: ${e}`));
         }
+
+        // Attempt to re-initialize with the new identifiers
+        try {
+            this.client = await this.initializeWithRetry(clientOptions, this.options.identifiers);
+        } catch (err) {
+            Logger.error(err, 'FeatureFlagClient: Failed to re-initialize');
```
marcomura, line 154: if we reach this catch, it means `this.client` was never updated and it's still referencing the old client where `.updateUser` failed. Is this correct?

src/util/featureFlags/featureFlagClient.ts
sdzh-atlassian: Generally, the client here fetches all the feature gate and experiment values on `initialize`/`updateUser`, and then doesn't make network calls at all after that IIRC

In the old implementation, the client keeps using previously saved flag values, and likely fails to update anything in the future since it becomes _busted_. E.g. if you logged out, you'd still see the stuff you're not supposed to see

In the new one, we try to reset on error. If the error was transient, this prevents the client from being _busted_ in the first place, and we get correct values :) If the error was actual (e.g. the client is offline) - we keep using whatever values were available before _updateUser_ was called

src/util/featureFlags/featureFlagClient.ts
sdzh-atlassian: yep! I haven't actually seen it get there during testing though, it always managed to recover in under 5 retries.

But there might be legitimate cases of this happening - like maybe if the user logs off while offline.

IMO there's some kind of a back-off strategy missing here, maybe we re-initialize the FF client every hour or something, or on a VSCode event - something to think about down the line

src/util/featureFlags/featureFlagClient.ts
sdzh-atlassian: IIRC tests got mad at me about this, I'll double-check

src/util/featureFlags/featureFlagClient.ts
marcomura: According to the other comment, if `updateUser` fails the client gets busted, so we shouldn't keep using the old client, no?

PR #778, opened 2025-08-20
[WIP] AXON-841: Upgrade basic interactive controls

sspanchenko:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


sspanchenko:
@marcomura @bhorai-atl 
We have a small issue with the new version of the lozenge. It looks like they changed the `default` color we use for the To Do status, and in the new version it now looks lighter and more transparent. 

Below I’ll provide an example for both dark and light themes, showing how it looked before and how it looks after the package upgrade.

Light theme (Before upgrade)
image

Light theme (After upgrage)
image

Dark theme (Before upgrage)
image

Dark theme (After upgrage)
image

I think the updated component looks particularly worse in the dark theme
What would you prefer we do about this?
I was thinking we could pick another color from the available ones, but to me none of them seem quite right for the `To Do` status. You can check them here: https://atlassian.design/components/lozenge/examples

Another option would be not to upgrade this package or we could upgrade not to the latest version (This has been broken since version 13)

image

Up to you

sspanchenko:
image

Lozenge `12.2.5` works fine

PR #793, opened 2025-08-21
Fallback FeatureFlag client for when updating the user fails

marcomura:
### What Is This Change?

This code introduces a "double feature flag client" used this way:
- a constant base client that references the user anonymously, without the tenant id
- a variable (possibly undefined) tenant client that references the user with the tenant id

When the user logs out / logs in, only the tenant client gets updated, or removed if the user doesn't have a tenant id correlation anymore.

When the tenant client is undefined, we fallback using the base client.
In case updating the tenant client fails (we attempt 5 times), we fallback using the base client.

In addition to the above changes, this change also refactor the FeatureFlagClient from a static class to a singleton class.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -182,26 +182,12 @@ export class Container {
         this._bitbucketHelper = new BitbucketCheckoutHelper(context.globalState);
         context.subscriptions.push(new HelpExplorer());
 
+        this._featureFlagClient = FeatureFlagClient.getInstance();
+
         try {
-            await FeatureFlagClient.initialize({
-                analyticsClient: this._analyticsClient,
-                identifiers: {
-                    analyticsAnonymousId: this.machineId,
-                    tenantId: this._siteManager.primarySite?.id,
-                },
+            await this._featureFlagClient.initialize({
+                analyticsAnonymousId: this.machineId,
             });
-
-            if (!process.env.ROVODEV_BBY) {
-                this._siteManager.onDidSitesAvailableChange(async () => {
-                    await FeatureFlagClient.updateUser({ tenantId: this._siteManager.primarySite?.id });
-
-                    if (FeatureFlagClient.checkGate(Features.RovoDevEnabled)) {
-                        this.enableRovoDev(context);
-                    } else {
-                        this.disableRovoDev(context);
-                    }
-                });
-            }
 
             Logger.debug(`FeatureFlagClient: Succesfully initialized the client.`);
             featureFlagClientInitializedEvent(true).then((e) => {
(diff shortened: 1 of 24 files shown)


src/util/featureFlags/featureFlagClient.ts
```diff
@@ -27,141 +26,155 @@ export enum PerimeterType {
...
-        return clientOptions as any;
+        this.featureGateOverrides = {} as FeatureGateValues;
+        this.experimentValueOverride = {} as ExperimentGateValues;
+        this.initializeOverrides();
     }
 
     /**
      * Workaround to account for a TLS-related ECONNRESET that sometimes occurs in undici
-     * when node attempts a naked `fetch`. The regular static client implementation of
+     * when node attempts a naked `fetch`. The regular client implementation of
```
sdzh-atlassian: Feels like some find-and-replace happened here, eh? :D

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -27,141 +26,155 @@ export enum PerimeterType {
...
+    private readonly featureGateOverrides: FeatureGateValues;
+    private readonly experimentValueOverride: ExperimentGateValues;
+    private readonly isExperimentationDisabled: boolean;
 
-    // TODO: rework this implementation to not be static
-    // - now that we're no longer using a static FG client
-    private static client?: FeatureGateClient = undefined;
+    private clientOptions: NewFetcherOptions = {} as any;
+    private identifiers: Identifiers = {};
+    private clientBasic?: FeatureGateClient;
```
sdzh-atlassian, line 47: NIT: we might want to document why the 2 clients exist, IMO it's not immediately clear from code

src/util/featureFlags/featureFlagClient.ts
```diff
@@ -265,7 +278,7 @@ export abstract class FeatureFlagClient {
         return gateValue;
     }
 
-    public static dispose() {
+    public dispose() {
         this.client?.shutdownStatsig();
```
sdzh-atlassian: doesn't really matter, but it might be prudent to nuke both fields directly here? If both clients exist, this will `shutdownStatsig` on only the tenant one, right?

src/container.ts
```diff
@@ -223,12 +226,22 @@ export class Container {
...
             this.enableRovoDev(context);
         }
     }
 
+    static async updateFeatureFlagTenantId(tenantId: string | undefined): Promise<boolean> {
+        try {
+            await this._featureFlagClient.updateUser({ tenantId });
+            return true;
+        } catch (err) {
+            Logger.error('RovoDev', err, "FeatureFlagClient: Failed to update user's tenantId");
```
sdzh-atlassian, line 240: Let's see how may of these we get in our error stats - I wonder if there's more legitimate cases of this happening that i couldn't think of yesterday

sdzh-atlassian (approved):
ship it! ⛵

src/util/featureFlags/featureFlagClient.ts
marcomura: this was a miss, thanks!

sdzh-atlassian (approved):
🚢 👍

PR #796, opened 2025-08-21
chore: ignore css chunk order warnings in devcompile

sdzh-atlassian:
### What Is This Change?

We've just caught a situation where `npm run compile` succeeds, but `npm run devcompile` fails on css chunk ordering warnings:
image

Arguably, this is _very_ undesirable - if the regular build succeeds, the `devcompile` should succeed as well.

To prevent this in the future, let's prevent css ordering warnings when webviews are being compiled

### How Has This Been Tested?

Did a before-and-after on a broken `main`

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/webpack.react.dev.js
+++ b/webpack.react.dev.js
@@ -21,6 +21,7 @@ module.exports = {
         mui: resolveApp('./src/react/index.tsx'),
     },
     devtool: 'cheap-module-source-map',
+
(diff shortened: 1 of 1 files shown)


marcomura (approved):
lgtm

PR #816, opened 2025-08-26
AXON-937: Bump @atlaskit/select to 21.2.5

sspanchenko:
### What Is This Change?

`@atlaskit/select 18.10.6 -> 21.2.5`

image

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


webpack.react.prod.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
     plugins: [
         new MiniCssExtractPlugin({
             filename: '[name].css',
+            ignoreOrder: true,
```
sspanchenko, line 77: I had the same errors as described in this PR. 
https://github.com/atlassian/atlascode/pull/796

Strange, I thought its merge was supposed to fix this, but in my case it only helps if we add the fix to the `webpack.react.prod.js` file.

om-ukr (approved):

webpack.react.prod.js
om-ukr: I am surprised we didn't replicate this change at the same time we updated webpack.react.dev.js

PR #828, opened 2025-08-27
Polish release scripts

marcomura:
### What Is This Change?

- Fixed the syntax of `npm version` in the yaml files
- Fixed the version regex to not allow leading zeros
- Cleaned up unused params

### How Has This Been Tested?

- [X] `manual tests`


scripts/version/assert-nightly.sh
```diff
@@ -3,7 +3,7 @@
 # Make sure that the minor version is odd
 version=$1
 
-if [[ $version =~ ^[0-9]+\.[0-9]*[13579]\.[0-9]+.*$ ]]; then
+if [[ $version =~ ^[0-9]+\.([1-9][0-9]*)?[13579]\.[0-9]+.*$ ]]; then
```
sdzh-atlassian, line 6: no zero-prefixed versions for us anymore eh? :)

scripts/version/get-latest-nightly.sh
```diff
@@ -1,13 +1,10 @@
 #!/usr/bin/env bash
 
-n_release=${1:-1}
-
 # Get latest nightly tag from the repository
 # Pre-releases are defined by the version number having an ODD minor number
 # https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions
-git tag --list | \
+git tag --list "*-nightly" | \
```
sdzh-atlassian: I'd keep the check of odd minor here instead, technically `*-nightly` only denotes that this particular version comes from CI

PR #831, opened 2025-08-27
chore: make nightly version resolution more verbose

sdzh-atlassian:
### What Is This Change?

Tiny change to make the version resolution logic that happens on builds a bit more verbose, to debug the odd discrepancy between `release` and `release-nightly`

image

### How Has This Been Tested?

See picture above ;)

### DIFF
--- a/scripts/version/get-next-nightly.sh
+++ b/scripts/version/get-next-nightly.sh
@@ -5,17 +5,37 @@
 # https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions
 
 git fetch --tags -q
+
+echo "Evaluating next nightly version..." >&2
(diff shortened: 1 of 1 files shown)


sdzh-atlassian:
This is to debug CI, overriding approvals

scripts/version/get-next-nightly.sh
```diff
@@ -5,17 +5,37 @@
 # https://code.visualstudio.com/api/working-with-extensions/publishing-extension#prerelease-extensions
 
 git fetch --tags -q
+
+echo "Evaluating next nightly version..." >&2
```
marcomura, line 9: I don't think you can echo in this file since the output is assigned to a variable

PR #839, opened 2025-08-28
[AXON-972] chore: decouple auth notfication from TreeNode implementation

sdzh-atlassian:
### What Is This Change?

Tiny change to decouple our oddly hardcoded login TreeNode from notifications 🤔
It looks like all we've been using from there was resource uri, and we're mainly using it as a map key - meaning that it should be OK to just replace it with a unique enough value?

This PR replaces the URI reference with whatever the URI is actually being set to, in the TreeNode

### How Has This Been Tested?

It wasn't :(


src/views/notifications/authNotifier.ts
```diff
@@ -49,12 +48,17 @@ export class AuthNotifier extends Disposable implements NotificationNotifier {
     }
 
     private checkJiraAuth(): void {
-        this.checkAuth(ProductJira, 'jira.login', 'Log in to Jira to view & manage work items', loginToJiraMessageNode);
+        this.checkAuth(
+            ProductJira,
+            'jira.login',
+            'Log in to Jira to view & manage work items',
+            Uri.parse('Please login to Jira'),
```
marcomura, line 55: why parsing a string into a Uri? this will become `file:///Please%20login%20to%20Jira`

marcomura: should this be `undefined` instead?

sdzh-atlassian: That's literally the value from `loginToJiraMessageNode.resourceUri`. If I understand usage correctly, it doesn't really matter what the value is, as long as it's unique - might want @bwieger-atlassian-com to weigh on on that though

Basically, my logic here was let's make as little changes as humanly possible to not break things - hence this somewhat odd literal

bwieger-atlassian-com: This looks correct. Testing locally

bwieger-atlassian-com: LGTM for the notifications. 

But for whatever reason, I can't get the OAuth button to work?

bwieger-atlassian-com: Okay, my issue was some staleness on my machine

PR #840, opened 2025-08-28
[AXON-802, AXON-979] Rovo Dev gracefully stops and restart on auth change

marcomura:
### What Is This Change?

This change implements the logic to stop and restart Rovo Dev on authentication change.

Specifically:
- when the API token in use is getting removed, Rovo Dev restarts
- when there are no available API tokens for Rovo Dev, Rovo Dev prints an error

This change also address the following Prompt Box issues:
1. Prompt Box doesn't get disabled on process terminated or rovo dev chat disabled
2. Prompt Box doesn't allow typing when the response is streaming back

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -226,7 +226,7 @@ export class Container {
                     if (this._featureFlagClient.checkGate(Features.RovoDevEnabled)) {
                         this.enableRovoDev(context);
                     } else {
-                        this.disableRovoDev(context);
+                        this.disableRovoDev();
                     }
                 }
             });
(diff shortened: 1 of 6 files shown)


src/container.ts
```diff
@@ -226,7 +226,7 @@ export class Container {
                     if (this._featureFlagClient.checkGate(Features.RovoDevEnabled)) {
                         this.enableRovoDev(context);
```
sdzh-atlassian: Should we `await` here now that the method is async? Seems like that's what we're trying to do elsewhere in this PR

src/container.ts
```diff
@@ -226,7 +226,7 @@ export class Container {
...
                     } else {
-                        this.disableRovoDev(context);
+                        this.disableRovoDev();
```
sdzh-atlassian: Same here

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -253,14 +252,14 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
-                                type: RovoDevProviderMessageType.WorkspaceChanged,
+                                type: RovoDevProviderMessageType.ProviderReady,
                                 workspaceCount: workspace.workspaceFolders?.length || 0,
                             });
                         }
 
-                        const port = parseInt(process.env[rovodevInfo.envVars.port] || '0');
-                        if (port) {
-                            this.signalProcessStarted(port);
+                        const fixedPort = parseInt(process.env[rovodevInfo.envVars.port] || '0');
```
sdzh-atlassian, line 260: As a general note - and I'm pretty bad at this myself, so I'm directing this to myself as well - it would be pretty nice to try and keep purely cosmetic refactoring to its own commits/PRs. It's kind of tricky to see what changes matter in this PR

src/jira/fetchIssue.test.ts
```diff
@@ -260,49 +260,4 @@ describe('fetchIssue', () => {
             await expect(fetchEditIssueUI(mockMinimalIssue)).rejects.toThrow(errorMessage);
         });
     });
-
-    describe('parallel execution verification', () => {
```
marcomura, line 264: Removed this test suite as it's testing the time required for executing the tests, and it's not good practice as the time is highly variable across machines.

sdzh-atlassian (approved):

sdzh-atlassian:
Odd, somehow GH wouldn't let me approve and comment at the same time :o

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: I hear you, but sometimes it's just in front of your eyes while you are doing other changes... 😄

src/container.ts
marcomura: yep, thanks for catching it

marcomura: adding the `await` exposed a pretty bad bug in Jira's `refreshAll()`, which returned a dangling Promise!

PR #862, opened 2025-08-28
Track Internal atlassian users

bwieger-atlassian-com:
### What Is This Change?
- If a user is logged into a site as @atlassian.com" email address associated with @atlassian.com, then there should be a flag to distinguish this. 

### How Has This Been Tested?
Manual, with break points

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -964,9 +964,9 @@ async function checkUserDomain(): Promise<string> {
 async function appendUserInfo<T>(e: Object): Promise<T> {
     let newObj: Object;
     const aaid = Container.siteManager?.getFirstAAID();
-    const userDomain = await checkUserDomain();
 
     if (aaid) {
+        const userDomain = await checkUserDomain();
         newObj = {
             ...e,
             ...{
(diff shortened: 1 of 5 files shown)


src/analytics.ts
```diff
@@ -927,14 +934,64 @@ function event(action: string, actionSubject: string, attributes: any): any {
...
+        }
+    }
+
+    return 'not-atlassian';
+}
+
+async function anyUserOrAnonymousWithAtlassianMetric<T>(e: Object): Promise<T> {
     let newObj: Object;
     const aaid = Container.siteManager?.getFirstAAID();
+    const userDomain = await checkUserDomain();
```
marcomura: this should be done inside `if (aaid)`

marcomura (approved):
left a comment but lgtm

src/analytics.ts
```diff
@@ -927,14 +934,64 @@ function event(action: string, actionSubject: string, attributes: any): any {
...
+        } catch {
+            // Continue to next site if this one fails
+            continue;
+        }
+    }
+
+    return 'not-atlassian';
+}
+
+async function appendUserInfo<T>(e: Object): Promise<T> {
```
sdzh-atlassian, line 969: Thanks for renaming this! :)

marcomura (approved):
lgtm!

PR #867, opened 2025-08-29
Build action: Run only for Pull Requests to main, and skip draft PRs

marcomura:
### What Is This Change?

Instead of running a build for every `push`, this change limits the build workflow to be run only for:
- Pull Requests that want to merge to main
- Pull Requests that are ready for review (not drafts)

NOTE: I still don't know if 
- this change will prevent draft from manually triggering a build action
- this change will prevent PR for a non-main branch to merge

### How Has This Been Tested?

- It hasn't.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,9 +3,10 @@ permissions:
   contents: read
 
 on:
-  push:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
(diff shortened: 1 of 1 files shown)


sdzh-atlassian (approved):
Personally, I quite like the build-on-push way of doing things - but I guess this really makes sense now that our builds are this long and complex

Let's try this for a while and see how it works 🥳

PR #874, opened 2025-08-30
[NOISSUE] use input validation to test API token/server credentials

sdzh-atlassian:
### What Is This Change?

This is kind of a proposed enhancement? Feel free to merge, nuke or iterate :)

This PR introduces the 1st part of our credential-saving logic as validation on the password input - basically, when user hits `enter`, we try to ping `/rest/api/2/myself` with the assembled credentials first, and show an error message if it fails.

There are 2 alternatives to this:
 * Don't do anything - user gets error at the end of the flow and has to restart from scratch
 * Fail during a special loading state like what we have for OAuth. Honestly, I kinda like that one better, but maybe we can iterate?

Works for API token:
image

Works for server auth - although we might need to do the special ssl/path settings before password for this to work in all cases:
image

### How Has This Been Tested?

Tried this against API token and server auth with instenv - in case of the server auth, the correct input takes a bit of time and correctly shows the loading state

Basic checks:

- [x] `npm run lint`
- [x ] `npm run test`

### DIFF
--- a/src/onboarding/quickFlow/baseUI.ts
+++ b/src/onboarding/quickFlow/baseUI.ts
@@ -25,6 +25,7 @@ export type ExtraOptions = {
     value?: string;
     prompt?: string;
     debounceValidationMs?: number;
+    preAcceptValidationFunc?: (value: string) => Promise<string | undefined>;
 };
 
 export class BaseUI {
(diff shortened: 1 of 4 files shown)


src/onboarding/quickFlow/baseUI.ts
```diff
@@ -64,24 +65,33 @@ export class BaseUI {
...
-                            input.validationMessage = result === null ? undefined : result;
-                        }
-                    });
-                } else {
-                    input.onDidChangeValue(async (value) => {
-                        const errorMessage = await props.validateInput?.(value);
-                        input.validationMessage = errorMessage || undefined;
-                    });
-                }
+            if (props.debounceValidationMs) {
```
sdzh-atlassian, line 68: Just removing the extra conditional here since it prevented `validationMessage` reset

PR #875, opened 2025-08-30
[NOISSUE] chore: server authentication - make password/PAT last step

sdzh-atlassian:
### What Is This Change?

> Note: this change takes care of edge cases in #874 

To make validation work in all cases, we might want to consider always having the PAT/password input at the end.
This change:
* Reorders server authentication process to ask for context path/SSL stuff first
* Does a minor cosmetic change of putting the default option for contextPath (`no`) on top

### How Has This Been Tested?

Manually - clicked through the process of server authentication

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


marcomura (approved):
lgtm

PR #876, opened 2025-08-30
[AXON-998] Disable OAuth auth option in remote environments

marcomura:
### What Is This Change?

OAuth doesn't work well in remote environments because the browser can't connect to localhost to provide the auth token.
This option should not be available when the environment is remote.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`


src/onboarding/quickFlow/authentication/authFlowUI.ts
```diff
@@ -91,32 +91,35 @@ export class AuthFlowUI {
...
-                },
-            ],
-            {
-                placeHolder: 'How would you like to authenticate?',
-                value: state.authenticationType || '',
-            },
-        );
+        const choices: QuickPickItem[] = [];
+
+        if (Container.isRemote) {
```
bwieger-atlassian-com: IsNotRemote?

marcomura: oops, I've inverted it for testing and forgot to revert it - thanks!

bwieger-atlassian-com (approved):

PR #879, opened 2025-08-31
[AXON-1002] Fix the inability to logout from individual sites

marcomura:
### What Is This Change?

The existing code doesn't allow to logout from individual sites because it's deleting the credentials of the site the user wants to logout from, causing all sites sharing the same credentials to be logged out as well.

With this change, we simply don't delete the credentials until they are not shared with any other sites.

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`


bwieger-atlassian-com:
Very hesitant to merge at this time 


> 879 I don't want to think about right now. I'm worried this might have some unknown side effect and it's not solving a customer-reported problem

PR #884, opened 2025-09-01
Axon-995: add the ability to update the status from the child issues table tree

ostoliar-atlassian:
### What Is This Change?

Add the ability to update the 'Status' column from the child issues table tree.
Add tooltip for the priority icon.

### Demo:
https://www.loom.com/share/110c96e980c6456aa3b7d75d2f7c3874

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/webviews/jiraIssueWebview.ts
```diff
@@ -362,6 +459,89 @@ export class JiraIssueWebview
...
+                    handled = true;
+                    const transitionMsg = msg as TransitionChildIssueAction;
+                    const issueKey = transitionMsg.issueKey;
+                    const statusName = transitionMsg.statusName;
+                    try {
+                        const client = await Container.clientManager.jiraClient(this._issue.siteDetails);
+
+                        const childIssue = await fetchMinimalIssue(issueKey, this._issue.siteDetails);
+
+                        const targetTransition = childIssue.transitions?.find((transition: any) => {
```
sspanchenko: Would it be better to extract this into a separate method—e.g., `const targetTransition = this.findMatchingTransition(childIssue.transitions, statusName)`? The current switch-case function is already excessively large

src/webviews/jiraIssueWebview.ts
```diff
@@ -121,6 +124,98 @@ export class JiraIssueWebview
...
+                                    transitions: enhanced.transitions,
+                                    status: enhanced.status,
+                                    assignee: enhanced.assignee,
+                                    priority: enhanced.priority,
+                                };
+                            }
+                        }
+                        if (issuelink.outwardIssue?.key) {
+                            const enhanced = enhancedIssuesMap.get(issuelink.outwardIssue.key);
+                            if (enhanced) {
```
sspanchenko: I noticed this `if` (or a very similar one) three times here — this should definitely be moved into a helper method

src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
```diff
@@ -350,6 +350,38 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
         );
     };
 
+    handleChildIssueUpdate = async (issueKey: string, fieldKey: string, newValue: any) => {
```
sspanchenko, line 353: This method has logic duplication - both if-else branches do essentially the same thing: call `postMessageWithEventPromise` with identical parameters 'editIssueDone', ConnectionTimeout, and nonce. Only the payload content differs.
The code could be simplified by extracting the common logic and creating the payload based on the field type, then calling `postMessageWithEventPromise` once. This would eliminate duplication and make the code more readable.

jimmyek85 (approved):

src/webviews/jiraIssueWebview.ts
ostoliar-atlassian: Thanks.
Done.

src/webviews/jiraIssueWebview.ts
ostoliar-atlassian: Thanks.
Done.

src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
ostoliar-atlassian: Thanks.
Done.

sspanchenko (approved):

PR #892, opened 2025-09-03
[AXON-1029] Rovo Dev should be disabled if the whole Jira feature is disabled

marcomura:
### What Is This Change?

Rovo Dev uses Jira sites' authentication token for its enablement and token usage.
If Jira is fully disabled, its sites shouldn't be used, therefore Rovo Dev should be disabled.

This PR addresses that.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -227,21 +227,24 @@ export class Container {
         }
 
         if (!process.env.ROVODEV_BBY) {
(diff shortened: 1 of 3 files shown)


cabella-dot (approved):
nice!

PR #943, opened 2025-09-10
AXON-904: Add a Rovo Dev Call to action in the Start work flow

ostoliar-atlassian:
### What Is This Change?
Add a Rovo Dev Call to action in the Start work flow
Requirements:
1. On the new Start work page, have a checkbox for “Start work with Rovo Dev”.
2. The checkbox should be default off for first time users.
3. If the user turns it on, the value should be sticky. Eg: when they return to start work again, the checkbox should be the same value they used last time. 
4. Checkbox should be located near the “final button” to start work so that the user can easily see it. 

### Before after screenshots:
axon-904-before
axon-904-after
axon-904-afterV3

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


bwieger-atlassian-com:
So far, looks great. Looking forward to having this on the new Start work screen as well!

src/react/atlascode/startwork/StartWorkPage.tsx
```diff
@@ -757,6 +787,21 @@ const StartWorkPage: React.FunctionComponent = () => {
                                             </Collapse>
                                         </Grid>
                                         <Grid item hidden={submitState === 'submit-success'}>
+                                            <FormControlLabel
```
bwieger-atlassian-com: I think we need this element to be guarded by a check if Rovo Dev is enabled. 

See how the "Start with Rovo Dev!" Button works on the view jira page.

ostoliar-atlassian: Thanks.
Done.

PR #944, opened 2025-09-10
[AXON-1072] Refactoring of Rovo Dev view currentState, fixed disappearing prompt after download

marcomura:
### What Is This Change?

1. This change merges the following view states into a single state, making the states transition little more predictable and reliable: `currentState`, `currentSubState`, `initState`, `downloadProgress`

2. It also fixes the disappearing prompt after the download has finished.

3. It also changes the "open workspace" message into a clearer view:

| Before | After |
|-----|-----|
| image | image |

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     ],
     "main": "./build/extension/extension",
     "rovoDev": {
-        "version": "0.11.16"
+        "version": "0.11.18"
     },
     "scripts": {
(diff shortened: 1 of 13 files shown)


cabella-dot (approved):
lgtm

PR #962, opened 2025-09-12
[Rovo Dev] Clean up workspaceCount view state

marcomura:
### What Is This Change?

Holding the `workspaceCount` is not required anymore since this state is now handled by `currentState`.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -62,7 +62,17 @@ jobs:
           docker tag ghcr.io/atlassian/atlascode-e2e:latest atlascode-e2e
 
       - name: Run E2E tests
-        run: npm run test:e2e:docker
+        run: |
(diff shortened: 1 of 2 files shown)


src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -58,6 +62,20 @@ export const CloseIconDeepPlan: React.FC<{}> = () => {
     );
 };
 
+function mapRovoDevDisabledReasonToSubState(reason: RovoDevDisabledReason): DisabledState['subState'] {
```
bwieger-atlassian-com, line 65: Need to double check the ordering here: 

Does Auth always trump the folder state? Auth > Folder

PR #968, opened 2025-09-13
AXON-1110: Simplify the auth display in the settings page

bwieger-atlassian-com:
### What Is This Change?
**User problem from interview:** 
Given I am first time user, when I authenticate I am presented a list of sites, I am confused because they don't look like authentication sessions and there are some many of them and they are visually overstimulating. 

**Solution**: 
- Dedup the OAuth sites by username. 
- Remove the images 
- Add the username to the row so it feels like an authentication session 

Out of scope for this PR: 
- API tokens for cloud should overwrite OAuth tokens. 
- API token should be scoped and work with Jira, Bitbucket and Rovo Dev 
- All auth flows should be updated to use just the scoped API token. Oauth will remain backwards compatible, but hidden.

| before | after |
|--------|-----|
| image |  image |

### How Has This Been Tested?
Manual


src/react/atlascode/config/auth/siteDeduplication.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (seenOAuthUsers.has(username)) {
+            return false; // Skip duplicate
+        }
+
+        seenOAuthUsers.add(username);
+
+        // Update the site name for OAuth sites
+        const productName = site.site.product.name;
+        if (productName === 'Jira') {
+            site.site.name = '';
```
sdzh-atlassian, line 36: Feels like this doesn't quite match the comment above. Is this intended to be `Jira Cloud` (and `Bitbucket Cloud` couple lines below)?

sdzh-atlassian:
Sorry, disabled the auto-merge just in case

PR #969, opened 2025-09-15
FLOW-486 Add aiResult analytic after viewing file for 5s

matt-lassian:
### What Is This Change?

When Boysenberry mode is enabled, track when users dwell on AI Generated code, and send an `aiResult viewed` analytic event if the user dwells for 5+seconds.

Loom demo: https://www.loom.com/share/026d89ce9d5449a49105817f7d48b82d

### How Has This Been Tested?

- Unit tests
- Tested locally - see loom demo

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 
N/A - only applicable for Boysenberry mode

Recommendations:
N/A - not user-facing change

### DIFF
--- a/src/rovo-dev/rovoDevDwellTracker.ts
+++ b/src/rovo-dev/rovoDevDwellTracker.ts
@@ -27,13 +27,6 @@ export class RovoDevDwellTracker implements Disposable {
 
         // Listen for editor changes and visible range changes (scrolling)
         this.disposables.push(window.onDidChangeActiveTextEditor((e) => this.onEditorFocusChanged(e)));
-        this.disposables.push(
-            window.onDidChangeTextEditorVisibleRanges((e) => {
-                if (e.textEditor === this.activeEditor) {
-                    this.startDwellTimer();
-                }
-            }),
-        );
 
         this.startDwellTimer();
     }
--- a/src/rovo-dev/rovoDevDwellTracker.ts
+++ b/src/rovo-dev/rovoDevDwellTracker.ts
@@ -1,5 +1,5 @@
 import { basename } from 'path';
-import { Disposable, TextEditor, window } from 'vscode';
+import { Disposable, window } from 'vscode';
 
 import { RovoDevApiClient } from './rovoDevApiClient';
 import { RovoDevTelemetryProvider } from './rovoDevTelemetryProvider';
@@ -13,8 +13,6 @@ import { RovoDevTelemetryProvider } from './rovoDevTelemetryProvider';
  */
 export class RovoDevDwellTracker implements Disposable {
     private disposables: Disposable[] = [];
-
-    private activeEditor: TextEditor | undefined;
     private dwellTimer: NodeJS.Timeout | undefined;
 
     constructor(
@@ -23,16 +21,13 @@ export class RovoDevDwellTracker implements Disposable {
         private readonly rovodevApiClient: RovoDevApiClient | undefined,
         private readonly dwellMs: number = 5000,
     ) {
-        this.activeEditor = window.activeTextEditor;
-
-        // Listen for editor changes and visible range changes (scrolling)
-        this.disposables.push(window.onDidChangeActiveTextEditor((e) => this.onEditorFocusChanged(e)));
+        // Listen for editor changes
+        this.disposables.push(window.onDidChangeActiveTextEditor(() => this.onEditorFocusChanged()));
 
         this.startDwellTimer();
     }
 
-    private onEditorFocusChanged(editor: TextEditor | undefined) {
-        this.activeEditor = editor;
+    private onEditorFocusChanged() {
         this.startDwellTimer();
     }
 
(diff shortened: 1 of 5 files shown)


Blastoplex:
👍 lgtm

sdzh-atlassian (dismissed):
Some advanced analytics, nice :)
I have no context on this, but LGTM

src/rovo-dev/rovoDevDwellTracker.ts
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }, this.dwellMs);
+    }
+
+    private getFileNamefromPath(path: string): string | undefined {
+        if (path === undefined) {
+            return undefined;
+        }
+        // Support windows and unix paths
+        const parts = path.split('\\')?.pop()?.split('/');
```
marcomura: you can use node's `path` for this:
https://www.w3schools.com/nodejs/shownodejs_cmd.asp?filename=demo_met_path_basename

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -941,6 +946,13 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
         await this._chatProvider.setReady(this._rovoDevApiClient);
 
         if (this.isBoysenberry) {
+            // Initialize global dwell tracker now that API client exists
+            this._dwellTracker?.dispose();
+            this._dwellTracker = new RovoDevDwellTracker(
+                this._telemetryProvider,
+                () => this._chatProvider.currentPromptId,
+                () => this._rovoDevApiClient,
```
marcomura: why not passing the client directly since the dwellTracker gets killed/recreated when the client do the same?

src/rovo-dev/rovoDevDwellTracker.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (!doc || doc.isUntitled || (doc.uri.scheme !== 'file' && doc.uri.scheme !== 'vscode-userdata')) {
+            return;
+        }
+
+        const editorAtStart = this.activeEditor;
+        const uriAtStart = doc.uri.toString();
+
+        this.dwellTimer = setTimeout(async () => {
+            // Ensure we are still on the same editor and document
+            const current = window.activeTextEditor;
```
marcomura, line 59: in which scenario we may be in a different editor?
shouldn't the editor be refreshed at the beginning of startDwellTimer instead of taking it from the cache set in the constructor?

src/rovo-dev/rovoDevDwellTracker.ts
matt-lassian: I tried using `path.basename` but the unit tests really didn't like it for some reason. Actually I did end up changing the tests a bunch after I hit that dead end so let me try using it again

src/rovo-dev/rovoDevWebviewProvider.ts
matt-lassian: Fair enough, no particular reason. I'll update it

src/rovo-dev/rovoDevDwellTracker.ts
matt-lassian: If the editor focus does change, the `activeEditor` is updated inside `onEditorFocusChanged`.

I suppose the edge case that is being guarded against here is if you start a dwellTimer inside of Editor_A, then in the middle of the timer you open a new window and change focus to this new Editor_B.

`this.activeEditor` would be updated from `onEditorFocusChanged`, but `editorAtStart` is still cached from the start of the timer in Editor_A and once the timer finishes it would be able to check that the editor has changed and cancel itself.

src/rovo-dev/rovoDevDwellTracker.ts
```diff
@@ -0,0 +1,108 @@
...
+        private readonly getCurrentPromptId: () => string,
+        private readonly rovodevApiClient: RovoDevApiClient | undefined,
+        private readonly dwellMs: number = 5000,
+    ) {
+        this.activeEditor = window.activeTextEditor;
+
+        // Listen for editor changes and visible range changes (scrolling)
+        this.disposables.push(window.onDidChangeActiveTextEditor((e) => this.onEditorFocusChanged(e)));
+        this.disposables.push(
+            window.onDidChangeTextEditorVisibleRanges((e) => {
```
marcomura: Sorry I haven't noticed it before.
This event fires when user scroll through the file and it's resetting the timer for every scroll.

Don't we want to consider dwelling even if the user is scroll through the changes?

matt-lassian: Good point, I wonder if it should be smarter and actually check if the visible ranges include AI changes. I'm leaning towards just removing the `onDidChangeTextEditorVisibleRanges` listener completely and just rely on the file being open

marcomura (approved):
looks good, thank you for this change and for addressing all the comments!

PR #971, opened 2025-09-15
AXON-1088: [Rovo Dev] Add buildPrompt to utils

sspanchenko:
### What Is This Change?

Previously, we had two functions that did the same thing. I moved this function to utilities so that it could be used in both places + and now we receive page data in the same format and we display the message as text rather than code.

**Before**: 
image

**After:**
image

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/util/rovoDevPrompt.ts
+++ b/src/util/rovoDevPrompt.ts
@@ -2,6 +2,10 @@
  * Utility functions for building Rovo Dev prompts from Jira issues
  */
 
+/**
+ * Removes Jira-specific markup from text to make it readable for Rovo Dev prompts.
+ * Converts: +bold text+ → bold text, [~accountid:123:user] → @user, !image.png! → [image attachment]
+ */
 const cleanJiraMarkup = (text: string): string => {
     return text
         .replace(/\+([^+]+)\+/g, '$1') // Remove bold markup


src/util/rovoDevPrompt.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Utility functions for building Rovo Dev prompts from Jira issues
+ */
+
+/**
+ * Removes Jira-specific markup from text to make it readable for Rovo Dev prompts.
+ * Converts: +bold text+ → bold text, [~accountid:123:user] → @user, !image.png! → [image attachment]
+ */
+const cleanJiraMarkup = (text: string): string => {
```
sspanchenko, line 9: When we fetch issue data from Jira API, we receive the raw storage format, not the processed UI version:
What users see in Jira UI: `@John Doe` 
What API returns: `[~accountid:557058:john.doe]` (internal storage format)
The Jira frontend processes this markup client-side to show readable names, but the API gives us the raw data.

So I added a function to handle this. I'd be happy to hear if you have any better ideas on how we should process incoming data from Jira for this Rovo message.

bwieger-atlassian-com: @sspanchenko Can you show what the output would look like if you did none of this clean up? 

I'm concerned that the clean up might reduce the context passed to the AI agent and therefore make the AI agent perform worse.

sspanchenko: Hey @bwieger-atlassian-com 

**With cleanup**:
Screenshot 2025-09-17 at 10 35 07

**Without cleanup**:
Screenshot 2025-09-17 at 10 51 29




Please let me know if this option is suitable for us or if we need to change it in some way

Thanks.

bwieger-atlassian-com: @sspanchenko Let's simplify this. 

https://www.loom.com/share/4ce0e4f21c404c799d35908c6137d303?sid=faf5c5a8-b14c-4846-8ec7-1950778ed557

sspanchenko: We discussed this during the daily meeting - we will merge it as it is. 
The next ticket I will work on is AXON-918.
The ticket includes requirements from loom-demo above.

PR #979, opened 2025-09-16
AXON-1138: Fix Create Issue form stability and input rendering

cabella-dot:
This PR addresses AXON-1138: Interface disappears when selecting issue type in 'Create Jira Issue' dropdown.\n\nKey changes:\n- Merge incoming fieldValues on 'update' to preserve form state (prevents transient blank states)\n- Harden validateString to guard !value and non-string inputs from DC project switches\n- Ensure text inputs/textareas/inline editors render primitives only (string/number/boolean), otherwise '' to avoid [object Object]\n- Initialize/reset EditRenderedTextArea state to avoid undefined leaking into UI\n\nManual verification steps:\n1. Switch project and issue types on both Cloud and DC; form fields should not disappear and should retain values unless changed\n2. For DC sites, switching projects no longer crashes on validateString and text fields no longer show [object Object]\n\nJira: https://softwareteams.atlassian.net/browse/AXON-1138

### DIFF
--- a/src/webviews/components/fieldValidators.test.ts
+++ b/src/webviews/components/fieldValidators.test.ts
@@ -105,6 +105,15 @@ describe('fieldValidators', () => {
             expect(validateString('test', state)).toBeUndefined();
             expect(validateString('', state)).toBe('EMPTY');
         });
+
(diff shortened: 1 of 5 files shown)


marcomura (approved):
lgtm

PR #986, opened 2025-09-17
[AXON-747] Dialog for consenting the use of third party MCP servers

marcomura:
### What Is This Change?

This change implements the dialog for requesting the consent for using configured third-party MCP servers.
The user can `allow` or `deny` individual servers, or `allow` all servers at once.

image

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/commands/jira/startWorkWithRovoDev.ts
+++ b/src/commands/jira/startWorkWithRovoDev.ts
@@ -1,8 +1,8 @@
 import { MinimalIssue, MinimalORIssueLink } from '@atlassianlabs/jira-pi-common-models';
-import TurndownService from 'turndown';
 
 import { DetailedSiteInfo } from '../../atlclients/authInfo';
 import { Container } from '../../container';
+import { buildRovoDevPrompt } from '../../util/rovoDevPrompt';
 
 export async function startWorkWithRovoDev(issue: MinimalORIssueLink<DetailedSiteInfo>) {
     const fullIssue = issue as MinimalIssue<DetailedSiteInfo>;
@@ -11,37 +11,12 @@ export async function startWorkWithRovoDev(issue: MinimalORIssueLink<DetailedSit
         throw new Error(`Jira issue not found`);
     }
 
-    const buildPrompt = (summary: string, description: string): string => {
-        return (
-            `
-            Let's work on this issue:
-
-            Summary:
-            ${summary}
-            ` +
-            (description
-                ? `
-            Description:
-            ${description}
-            `
-                : '') +
-            `
-            Please provide a detailed plan to resolve this issue, including any necessary steps, code snippets, or references to documentation.
-            Make sure to consider the context of the issue and provide a comprehensive solution.
-            Feel free to ask for any additional information if needed.
-        `
-        );
-    };
+    const client = await Container.clientManager.jiraClient(fullIssue.siteDetails);
+    const fullIssueData = await client.getIssue(fullIssue.key, ['description'], '');
 
     const summary = fullIssue.summary || '';
+    const description = fullIssueData.fields.description || '';
 
-    let description = '';
-    if (fullIssue.descriptionHtml) {
-        const turnDownService = new TurndownService();
-        description = turnDownService.turndown(fullIssue.descriptionHtml);
-    }
-
-    const prompt = buildPrompt(summary, description);
-
+    const prompt = buildRovoDevPrompt(summary, description);
     Container.rovodevWebviewProvider.invokeRovoDevAskCommand(prompt);
 }
--- a/src/util/rovoDevPrompt.test.ts
+++ b/src/util/rovoDevPrompt.test.ts
@@ -1,0 +1,78 @@
+import { buildRovoDevPrompt } from './rovoDevPrompt';
+
+describe('rovoDevPrompt', () => {
+    describe('buildRovoDevPrompt', () => {
+        it('should create prompt with summary and description', () => {
+            const summary = 'Fix authentication bug';
+            const description = 'Users cannot login after update';
+
+            const result = buildRovoDevPrompt(summary, description);
+
+            expect(result).toContain("Let's work on this issue:");
+            expect(result).toContain('Summary:');
+            expect(result).toContain(summary);
+            expect(result).toContain('Description:');
+            expect(result).toContain(description);
+            expect(result).toContain('Please provide a detailed plan to resolve this issue');
+        });
+
+        it('should create prompt with summary only when description is empty', () => {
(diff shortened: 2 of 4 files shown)


src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
```diff
@@ -205,23 +209,26 @@ export const ChatStream: React.FC<ChatStreamProps> = ({
...
-                return currentState.isPromptPending;
-            default:
-                return true;
-        }
-    }, [currentState]);
+    const isChatHistoryDisabled =
+        (currentState.state === 'Initializing' && currentState.subState === 'MCPAcceptance') ||
+        currentState.state === 'Disabled';
+
+    const shouldShowToolCall =
```
sdzh-atlassian, line 214: Love to see one less thing memoized :)

src/react/atlascode/rovo-dev/rovoDevLanding.tsx
```diff
@@ -75,6 +77,44 @@ export const RovoDevLanding: React.FC<{
                     </button>
                 </div>
             )}
+
+            {currentState.state === 'Initializing' && currentState.subState === 'MCPAcceptance' && (
+                <div className="form-container" style={{ marginTop: '24px', textAlign: 'left', gap: '18px' }}>
```
sdzh-atlassian, line 82: Any reason not to make this a separate component, now or in the future?

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -696,6 +706,32 @@ const RovoDevView: React.FC = () => {
...
+                serverName,
+            });
+        },
+        [postMessage],
+    );
+
+    const onMcpDecline = useCallback(
+        (serverName?: string, all?: boolean) => {
+            postMessage({
+                type: RovoDevViewResponseType.McpAcceptance,
```
sdzh-atlassian: Nit: I honestly don't have a suggestion for this, but having an `acceptance` event also handle `deny` might probably get confusing down the line. Maybe `McpChoiceSubmit` or something?

src/rovo-dev/rovoDevApiClient.ts
```diff
@@ -33,12 +33,24 @@ export interface RovoDevChatRequest {
     enable_deep_plan?: boolean;
 }
 
-export interface RovoDevHealthcheckResponse {
-    status: 'healthy' | 'unhealthy' | 'entitlement check failed' | 'pending user review';
+export interface RovoDevHealthcheckResponse_Base {
```
sdzh-atlassian: Are we trying to enforce type validation where `status: 'pending user review'` would correspond to a  defined `mcpServers`? I get the benefit but IMO this is a bit cursed :D Can we at least keep to a consistent naming convention without an underscore?

Edit: nvm I guess this is a thing we do elsewhere in the codebase, too

src/rovo-dev/rovoDevTypes.ts
```diff
@@ -46,27 +46,35 @@ export interface TechnicalPlan {
 
 // ---- Rovo Dev Chat States ----
 
-export interface BasicState {
-    state: 'WaitingForPrompt' | 'GeneratingResponse' | 'CancellingResponse' | 'ExecutingPlan' | 'ProcessTerminated';
+export interface InitializingState_Base {
```
sdzh-atlassian: Looks like we're doing `_Base` here as well. Sorry if I missed when this became a thing - if this is actually consistent in the codebase, I have no objection

src/rovo-dev/rovoDevApiClient.ts
marcomura: yeah, type discrimination based on the status type. It's one of the main points of using typescript 😄

src/react/atlascode/rovo-dev/rovoDevLanding.tsx
marcomura: imho too little to be a separate component, but I don't have strong opinions

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: Renamed to `McpConsentChoiceSubmit`

src/rovo-dev/rovoDevApiClient.ts
marcomura: Renamed to `Abstract<name>`

src/rovo-dev/rovoDevTypes.ts
marcomura: Renamed to `Abstract<name>`

src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
```diff
@@ -39,6 +39,8 @@ interface ChatStreamProps {
     sendFeedback: (feedbackType: FeedbackType, feedack: string, canContact: boolean, lastTenMessages: boolean) => void;
     onLoginClick: () => void;
     onOpenFolder: () => void;
+    onMcpAccept: (serverName?: string, all?: boolean) => void;
```
bwieger-atlassian-com: We can simplify the onMcpAccept signature

if the serverName is not supplied, then the assumption is "All" -> the second param here can be dropped

src/react/atlascode/rovo-dev/rovoDevView.tsx
sdzh-atlassian: ty ty!

sdzh-atlassian (approved):

src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
marcomura: I thought about that, but this could lead to a situation where an external (CLI) bug preventing the serverName to be defined would cause the acceptance of all servers, despite the user's choice to be on a specific one.

To ensure that all servers are accepted only when the user clicks on the right button, we should keep the two actions explicitly distinct.

bwieger-atlassian-com:
@marcomura This is progress, but looks like a fatal bug in the flow:  

https://www.loom.com/share/3c8d9036352a46f894a0b74a3faa8760?sid=9a3e55a0-725c-40ca-9b87-21fc9296cca3

src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
marcomura: Simplified in a different and better way 😄

PR #998, opened 2025-09-19
[Rovo Dev] Fixed the healtcheck polling, and more informative API call failures

marcomura:
### What Is This Change?

Fixed a bug in the healthcheck poll that was preventing the newly changed status to be detected.
Initialization time reduced thanks to this fix.

Also, added a try/catch in the inner fetch function of the api client to inform what's the failure reason when Node's fetch failed to send a request.

### How Has This Been Tested?

- [x] `npm run lint`
- [x] `npm run test`
- [x] `manual tests`

### DIFF
--- a/src/util/waitFor.ts
+++ b/src/util/waitFor.ts
@@ -1,0 +1,81 @@
+import { setTimeout } from 'timers/promises';
+
+type ExecuteCheckReturnType<T> = { checkPassed: true; result: T } | { checkPassed: false };
+
+const executeCheck = async <T>(
+    check: () => Promise<T> | T,
+    condition: (value: T | undefined) => Promise<boolean> | boolean,
+): Promise<ExecuteCheckReturnType<T>> => {
+    try {
+        const result = await check();
+        const passed = await condition(result);
+        return passed ? { checkPassed: true, result } : { checkPassed: false };
+    } catch {
+        return { checkPassed: false };
+    }
+};
+
+/**
+ * Polls the provided `check` function every `interval` milliseconds until the callback `condition` returns true, or the timeout time `timeout` occurrs.
+ * @param condition The condition to wait for
+ * @param check The check to perform at the given interval time.
+ * @param timeout The timeout, in milliseconds, after which we give up.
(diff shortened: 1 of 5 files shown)


src/util/waitFor.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Polls the provided `check` function every `interval` milliseconds until the callback `condition` returns true, or the timeout time `timeout` occurrs.
+ * @param condition The condition to wait for
+ * @param check The check to perform at the given interval time.
+ * @param timeout The timeout, in milliseconds, after which we give up.
+ * @param interval The interval in milliseconds.
+ * @param abortIf An optional callback, executed between intervals, that causes this wait to abort prematurely if its condition verifies.
+ * @returns The last value returned by `check` when the `condition` satisfies.
+ * @throws If `abortIf` returns true, or the `timeout` occurs.
+ */
+export async function waitFor<T>(
```
sdzh-atlassian: Love this, but IMO with this number of parameters, and _especially_ with some of them being optional (like timeout/abortIf) - I'd seriously consider using an [options object pattern](https://medium.com/@bchadwickfrance/the-options-object-pattern-with-typescript-a14f33306ec8)

```ts
export async function waitFor<T>({
    condition,
    check,
    timeout = 5000,
    interval = 200,
    abortIf
} : {
    condition: (value: T | undefined) => Promise<boolean> | boolean,
    check: () => Promise<T> | T,
    timeout?: number,
    interval?: number,
    abortIf?: () => boolean,
}) {}
```

(or some such)

Then usage becomes a lot more readable IMO, and we can skip optionals without putting them at the end
```ts
const initialized = await safeWaitFor({
    condition: (value) => !!value,
    check: () => !!this._webView,
});
```

This is going into super nitpick territory, but there might be a way to use more self-documenting parameter names here, too - `condition` and `check` are a bit generic and might not be clear in 6 months w/o reading the docs (but thanks for writing the docs though! :D)

sdzh-atlassian (approved):
Awesome, thanks!

PR #1000, opened 2025-09-19
[Rovo Dev][Debuggability] Don't shutdown Rovo Dev when unhealthy so the terminal can stay alive

marcomura:
### What Is This Change?

When Rovo Dev can't initialize (unhealthy status), today we shut it down to dispose resources.
Unfortunately, when debugging, this is inconvenient because we can't access the terminal anymore to inspect its logs.

This change prevents the shutdown to happen when Rovo Dev fails to initialize.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/rovoDevProcessManager.ts
+++ b/src/rovo-dev/rovoDevProcessManager.ts
@@ -11,6 +11,7 @@ import { Disposable, ExtensionContext, Terminal, Uri, window, workspace } from '
 import { isBasicAuthInfo, ProductJira } from '../atlclients/authInfo';
 import { rovodevInfo } from '../constants';
 import { Container } from '../container';
+import { RovoDevApiClient } from './rovoDevApiClient';
 import { RovoDevWebviewProvider } from './rovoDevWebviewProvider';
 
 export const MIN_SUPPORTED_ROVODEV_VERSION = packageJson.rovoDev.version;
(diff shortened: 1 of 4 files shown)


src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -420,11 +431,25 @@ class RovoDevTerminalInstance extends RovoDevInstance {
...
+            this.finalizeStop();
+        }
+    }
+
+    // don't rely on the RovoDevWebviewProvider for shutting Rovo Dev down, as it may
+    // already have set itself as Terminated and lost the reference to the API client
+    private async shutdownRovoDev() {
+        if (this.httpPort) {
+            try {
+                await new RovoDevApiClient('localhost', this.httpPort).shutdown();
```
sdzh-atlassian, line 446: I get why we're doing it, but this feels a bit sus. Don't have a suggestion for how to improve it though 🤔 Did we get issues around process shutdown before?

sdzh-atlassian (approved):
Since this is primarily for debug anyway - ship it! :)

src/rovo-dev/rovoDevProcessManager.ts
marcomura: No, this is the first time. It's needed because not shutting down the process in webviewProvider, it gets out of sync with the state of the process.

Not ideal, but it still makes sense that the process manager is able to kill the process without relying on the webview provider.
Unfortunately the Terminal APIs don't offer anything to kill the process, so this is the only way.

PR #1003, opened 2025-09-20
ESLint: Disallow the use of React dangerouslySetInnerHTML

marcomura:
### What Is This Change?

Added the following rule: https://eslint-react.xyz/docs/rules/dom-no-dangerously-set-innerhtml
to disallow the usage of React's dangerouslySetInnerHTML, since it's a security risk if not properly guarded.

Current usage has been allow-listed.

### How Has This Been Tested?

- [X] `npm run lint`

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -11,7 +11,7 @@ const compat = new FlatCompat({
 });
 
 export default tseslint.config(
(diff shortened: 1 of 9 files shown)


sdzh-atlassian (approved):
There's something a bit funny about us adding a guardrail lint rule, only to permanently allow the thing it's guarding against everywhere :)

I get the logic though, let's ship it

PR #1010, opened 2025-09-22
Fix Logger.error signature and some incorrect Error data (alt version)

marcomura:
### What Is This Change?

The `Logger.error` signature was temporarily hacked to allow being overloaded with and without the first argument of type ProductArea.

This is leading to some incorrect error data when the error (when passed as first param) is of type `string` instead of `Error`, which ideally it shouldn't happen.

This refactoring makes sure the only correct signature is being called.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -51,7 +51,7 @@ export class Logger {
     }
 
     // constructor is private to ensure only a single instance is created
-    private constructor() {}
+    protected constructor() {}
 
     public static get Instance(): Logger {
         return this._instance || (this._instance = new this());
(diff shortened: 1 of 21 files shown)


src/rovo-dev/rovoDevFeedbackManager.ts
```diff
@@ -1,7 +1,7 @@
 import { truncate } from 'lodash';
 import { Container } from 'src/container';
 import { getAxiosInstance } from 'src/jira/jira-client/providers';
-import { Logger } from 'src/logger';
+import { RovoDevLogger } from 'src/logger';
```
sdzh-atlassian, line 4: I'd actually even keep the change to the import statement alone here
```ts
import { RovoDevLogger as Logger } from 'src/logger'; // import the specific logger as the generic logger

// usage - no change
Logger.error(error, 'something');
```

But either way works IMO

src/rovo-dev/rovoDevPullRequestHandler.ts
```diff
@@ -1,5 +1,5 @@
 import { exec } from 'child_process';
-import { Logger } from 'src/logger';
+import { Logger, RovoDevLogger } from 'src/logger';
```
sdzh-atlassian: Do we need the legacy logger here for anything?

src/logger.ts
```diff
@@ -196,3 +190,19 @@ export class Logger {
         return `[${time}:${('00' + now.getUTCMilliseconds()).slice(-3)}]`;
     }
 }
+
+export class RovoDevLogger extends Logger {
+    public static override error(ex: Error, errorMessage?: string, ...params: string[]): void {
+        // `retrieveCallerName` must be called from the VERY FIRST FUNCTION that the called invoked from Logger.
+        // If not, the function will return the name of a method inside Logger.
+        const callerName = retrieveCallerName();
+        this.errorInternal('RovoDev', ex, callerName, errorMessage, ...params);
```
sdzh-atlassian, line 199: Are these overrides actually needed here? I'd go with
```ts
class Logger {
    public static error(ex: Error, errorMessage?: string, ...params: string[]): void {
        // do stuff

        // not sure if we still need errorInternal with this approach, using it for reference
        this.errorInternal(this.productArea, /* other parameters */); // in a static method, `this` will resolve to the correct class
    }

    static productArea?: ErrorProductArea = undefined;
}

class RovoDevLogger {
    // literally only change the productArea
    static productArea: ErrorProductArea = 'RovoDev';
}
```

But this is basically a giant nitpick, feel free to disregard :D

marcomura: unfortunately yes, because today we are still mixing static usage with instance usage

src/rovo-dev/rovoDevFeedbackManager.ts
marcomura: I'd prefer to have it clear which Logger is being used.
In some places, more than a Logger can be used, like in `Container.ts`

src/rovo-dev/rovoDevPullRequestHandler.ts
marcomura: Can simplify

sdzh-atlassian (approved):

PR #1022, opened 2025-09-24
FLOW-626 Fix image error loop if fallback fails

matt-lassian:
### What Is This Change?

The `attachImageErrorHandler` attempts to load a fallback image `images/no-image.svg` if an image fails to load. However if `images/no-image.svg` also fails to load then it gets into a loop where 10s-100s of errors start flooding the console.

Screenshot 2025-09-24 at 9 37 39 PM

These errors were observed in Boysenberry embedded in an iframe. Attempting to load `no-image.svg` returns a `401 Unauthorized`.
I can try and fix the root cause for why the image fails in a different PR, but for now this should at least stop it from getting into a loop.

### How Has This Been Tested?

Tested locally.

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/react/imageErrorHandler.tsx
+++ b/src/react/imageErrorHandler.tsx
@@ -3,6 +3,12 @@ export const attachImageErrorHandler = () => {
         'error',
         (ee: ErrorEvent) => {
             const targetEL = ee.target as HTMLElement;
+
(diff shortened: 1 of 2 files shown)


sdzh-atlassian (approved):
Nice, ty!

PR #1027, opened 2025-09-24
AXON-1197: user should be given a first time user experience

ostoliar-atlassian:
### Ticket:
https://softwareteams.atlassian.net/browse/AXON-1197

### What Is This Change?
When a user re-installs / upgrade the extension, it should acts as if it is a first time user if they haven't been active in the past 90 days
When the extension is installed / upgraded / re-installed, User should be given a “First time user” experience when 

They haven’t been active for 90 days (see pmfStats.touchActivity) 

AND They are not logged in to Jira 

Note: this behavior should only be trigger at most once per week. (This will require a new tracker to be set up) 

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ### [Report an Issue](https://github.com/atlassian/atlascode/issues)
 
+## What's new in 3.8.19
+
+### Bug Fixes
+
+- Fix error while connecting several jira sites with an API token
+
 ## What's new in 3.8.17
 
 ### Features
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,8 +8,9 @@
 # AXON TEAM: @bwieger-atlassian-com @sdzh-atlassian @cabella-dot @marcomura
 # DEV INNER LOOP: @akalia-atlassian @pastelsky
 # SOFTSERVE TEAM: @bhorai-atl
+# BBY FOLKS: @Blastoplex @mattcolman
 
-*       @bwieger-atlassian-com @sdzh-atlassian @cabella-dot @marcomura @akalia-atlassian @pastelsky @bhorai-atl @om-ukr
+*       @bwieger-atlassian-com @sdzh-atlassian @cabella-dot @marcomura @akalia-atlassian @pastelsky @bhorai-atl @om-ukr @mattcolman @Blastoplex
 
 # Order is important; the last matching pattern takes the most
 # precedence. When someone opens a pull request that only
--- a/package-lock.json
+++ b/package-lock.json
@@ -14024,6 +14024,40 @@
                 "node": ">=14.17.0"
             }
         },
+        "node_modules/@emnapi/core": {
+            "version": "1.5.0",
+            "resolved": "https://packages.atlassian.com/api/npm/npm-remote/@emnapi/core/-/core-1.5.0.tgz",
+            "integrity": "sha512-sbP8GzB1WDzacS8fgNPpHlp6C9VZe+SJP3F90W9rLemaQj2PzIuTEl1qDOYQf58YIpyjViI24y9aPWCjEzY2cg==",
+            "dev": true,
+            "license": "MIT",
+            "optional": true,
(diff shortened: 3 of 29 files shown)


src/extension.ts
```diff
@@ -88,7 +88,13 @@ export async function activate(context: ExtensionContext) {
         if (previousVersion === undefined) {
             commands.executeCommand(Commands.ShowOnboardingFlow);
         } else {
-            showWelcomePage(atlascodeVersion, previousVersion);
+            // Check if user should get first-time experience due to long inactivity
+            const shouldShowFirstTimeExperience = await checkForFirstTimeExperienceOnReinstall();
+            if (shouldShowFirstTimeExperience) {
+                commands.executeCommand(Commands.ShowOnboardingFlow);
```
bwieger-atlassian-com: We can probably condense this if () else ( if() else()) to just be one if else,

src/extension.ts
```diff
@@ -88,7 +88,13 @@ export async function activate(context: ExtensionContext) {
         if (previousVersion === undefined) {
             commands.executeCommand(Commands.ShowOnboardingFlow);
         } else {
-            showWelcomePage(atlascodeVersion, previousVersion);
+            // Check if user should get first-time experience due to long inactivity
+            const shouldShowFirstTimeExperience = await checkForFirstTimeExperienceOnReinstall();
+            if (shouldShowFirstTimeExperience) {
```
bwieger-atlassian-com: This logic looks correct!

src/extension.ts
```diff
@@ -204,6 +210,30 @@ async function sendAnalytics(version: string, globalState: Memento) {
     });
 }
 
+// Check if user should get first-time experience on reinstall/upgrade
+async function checkForFirstTimeExperienceOnReinstall(): Promise<boolean> {
```
bwieger-atlassian-com, line 217: This all looks good. 

I'm wondering if there's a great way to detect if the user has manually click the "install" / "upgrade" button.

ostoliar-atlassian: @bwieger-atlassian-com 
Regarding this goal - “As a user who manually upgrades/installs the extension, I should get a first time experience.” AND “As a user who has VS Code automatically upgrading the extension, nothing should happen” 

**VS Code can not distinguish Manual vs Automatic updates.** VS Code Extension API does not provide any mechanism to detect whether an extension was installed/updated manually by the user or automatically by VS Code. VS Code automatically updates extensions in the background without any extension-side notification about the trigger source. Both manual and automatic updates result in the same activation lifecycle. 

**I've implemented a heuristic-based solution that attempts to distinguish between manual and automatic extension activations using timing and behavioral patterns.** 
Heuristic Logic:
The system considers an activation "likely manual" when ALL of these conditions are met:
_Working Hours:_ Activation happens between 9 AM - 6 PM (when users are typically active)
_Reasonable Interval:_ At least 1 hour since last activation (prevents rapid-fire automatic updates)
_Recent User Activity:_ User has used the extension within the last 7 days
_First Install:_ If no previous activation time exists, it's assumed manual

**Important Limitations:**
_Not 100% Accurate:_ This is heuristic-based and can have false positives/negatives
_User Behavior Dependent:_ Assumes users work standard hours and use VS Code regularly
_Time Zone Agnostic:_ Uses local time, which may not work for all users
_No Direct API:_ VS Code doesn't provide actual manual vs automatic detection

Why This Approach:
Since direct detection is technically impossible in VS Code, this heuristic approach provides reasonable accuracy for most users. Is better than no distinction at all.

src/extension.ts
ostoliar-atlassian: Done

bwieger-atlassian-com:
@ostoliar-atlassian Super sorry. Because we can't guarantee if the user is manually updating vs auto-updating, I don't think this will go well. Closing this PR. 

I deeply appreciate you looking into this and coming up with some good heuristics. Well done!

PR #1030, opened 2025-09-24
Axon 1181 better auth nudge

sdzh-atlassian:
### What Is This Change?

Add API token authentication nudge to the create issue page, in line with how we do it in rovodev

|Before|After|
|------|------|
|image|image|

Notably, this _doesn't_ refresh the page with the new credentials, that will be a separate PR

### How Has This Been Tested?

Manually! Also, more changes to come, turns out the logic where we check for token availability is a bit janky

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


marcomura (approved):
nice!

PR #1031, opened 2025-09-24
[AXON-1811] chore: use direct call instead of command when possible

sdzh-atlassian:
### What Is This Change?

Small PR tangentially related to the #1030 - for authentication, use direct call when possible if triggering quick flow commands

Ran into a type safety issue which was a bit unpleasant to debug because commands don't do argument type validation

### How Has This Been Tested?

Manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/onboarding/quickFlow/index.ts
+++ b/src/onboarding/quickFlow/index.ts
@@ -3,25 +3,38 @@ import { Commands } from 'src/constants';
 import { commands, Disposable, window } from 'vscode';
 
 import { AuthFlow, AuthFlowData } from './authentication';
+import { QuickFlowStatus } from './types';
+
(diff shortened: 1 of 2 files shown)


src/onboarding/quickFlow/index.ts
```diff
@@ -3,25 +3,38 @@ import { Commands } from 'src/constants';
 import { commands, Disposable, window } from 'vscode';
 
 import { AuthFlow, AuthFlowData } from './authentication';
+import { QuickFlowStatus } from './types';
+
+type QuickAuthOrigin = 'settings' | 'nudge' | 'command';
+
+export async function runQuickAuth({
```
sdzh-atlassian, line 10: Moved this out of the command definition below

marcomura (approved):

PR #1032, opened 2025-09-24
[AXON-1215] Migrate to the Rovo Dev v3 APIs, still in YOLO mode

marcomura:
### What Is This Change?

Consume the /v3 APIs instead of the /v2, which offer the ability to pause the streaming before every tool call to request user confirmation before executing them.

With this change, we automatically allow every tool call, which keeps the logic equivalent to the YOLO mode.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/responseParser.ts
+++ b/src/rovo-dev/responseParser.ts
@@ -89,7 +89,7 @@ interface RovoDevWarningChunk {
     data: RovoDevWarningResponseRaw;
 }
 
-// ??
+// https://bitbucket.org/atlassian/acra-python/src/9ce5910e61d00e91f70c7978e067bde2690a1c97/packages/cli-rovodev/docs/serve/streaming-events.md?at=RDA-307-emit-warning-events-related-to-rate-limits-and-other-api-request-problems#:~:text=Server%20Error%20Warnings
 interface RovoDevClearResponseRaw {
     message: string;
 }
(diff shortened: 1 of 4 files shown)


src/rovo-dev/responseParser.ts
```diff
@@ -109,6 +109,21 @@ interface RovoDevPruneChunk {
     data: RovoDevPruneResponseRaw;
 }
 
+// ??
```
sdzh-atlassian: Was something supposed to be here? :)

marcomura: fixed :)

src/rovo-dev/rovoDevChatProvider.ts
```diff
@@ -358,8 +358,17 @@ export class RovoDevChatProvider {
                     },
                 });
 
+            case 'on_call_tools_start':
+                // simulate YOLO mode
```
sdzh-atlassian, line 362: Are we basically silently approving tool calls here?

marcomura: correct

sdzh-atlassian (approved):
LGTM, let's ship it and fix stuff down the line if needed

sdzh-atlassian (approved):

PR #1041, opened 2025-09-26
AXON-1181 use any available key

sdzh-atlassian:
### What Is This Change?

This change rewrites how we resolve API token availability for the purpose of issue suggestions

|  | Before | After |
|--|------|-------|
| When deciding if suggestions are available | Check for API token credential on specific site | Check for an API token applicable to the selected site (via matching email) |
| When making the API call | Use first available cloud site with an API token | Use the exact site the user has selected in the `Create Issue` page |

Also featuring: fix to a very annoying bug where the generated values would reset when switching sites

### How Has This Been Tested?

Manually :)
https://www.loom.com/share/3302b30e620947eeb9d3d7a1d58254b8

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/atlclients/issueBuilder.ts
+++ b/src/atlclients/issueBuilder.ts
@@ -21,49 +21,41 @@ type SiteId = string;
 
 const ASSIST_API = '/gateway/api/assist/api/ai/v2/ai-feature/jira/issue/source-type/code/suggestions';
 
-export const isSiteCloudWithApiKey = async (site?: DetailedSiteInfo | SiteId): Promise<boolean> => {
+export const findApiTokenForSite = async (site?: DetailedSiteInfo | SiteId): Promise<BasicAuthInfo | undefined> => {
     const siteToCheck = typeof site === 'string' ? Container.siteManager.getSiteForId(ProductJira, site) : site;
 
-    if (!siteToCheck || !siteToCheck.host) {
-        return false;
+    if (!siteToCheck || !siteToCheck.host.endsWith('.atlassian.net')) {
+        return undefined;
     }
 
-    if (!siteToCheck.host.endsWith('.atlassian.net')) {
-        return false;
-    }
+    const sites = Container.siteManager.getSitesAvailable(ProductJira);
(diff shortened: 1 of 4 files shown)


src/atlclients/issueBuilder.ts
```diff
@@ -21,49 +21,41 @@ type SiteId = string;
...
-        )
-    ).filter((x) => x !== undefined) as DetailedSiteInfo[];
-
-    // Any site is fine, just need an API key
-    return sites.length > 0 ? sites[0] : undefined;
-};
-
-export const fetchIssueSuggestions = async (prompt: string, context?: string): Promise<SuggestedIssuesResponse> => {
+export const fetchIssueSuggestions = async (
+    site: DetailedSiteInfo,
```
sdzh-atlassian, line 51: Accepts `site` and uses it explicitly now

src/webviews/createIssueWebview.ts
```diff
@@ -389,6 +389,12 @@ export class CreateIssueWebview
             return;
         }
 
+        this._partialIssue = {
```
sdzh-atlassian, line 392: Without this, the generated fields get nuked on switching sites

marcomura (approved):
lgtm!

src/atlclients/issueBuilder.ts
```diff
@@ -21,49 +21,41 @@ type SiteId = string;
 
 const ASSIST_API = '/gateway/api/assist/api/ai/v2/ai-feature/jira/issue/source-type/code/suggestions';
 
-export const isSiteCloudWithApiKey = async (site?: DetailedSiteInfo | SiteId): Promise<boolean> => {
+export const findApiTokenForSite = async (site?: DetailedSiteInfo | SiteId): Promise<BasicAuthInfo | undefined> => {
```
bwieger-atlassian-com, line 24: can we push this into the credentialManager?

bwieger-atlassian-com (approved):

src/atlclients/issueBuilder.ts
sdzh-atlassian: Sure thing, will do in a follow-up

PR #1063, opened 2025-09-30
AXON-1257: 'Start with Rovo Dev' should not be available when Rovo Dev is disabled

ostoliar-atlassian:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -4,7 +4,7 @@ import * as vscode from 'vscode';
 
 import { featureFlagClientInitializedEvent } from './analytics';
 import { AnalyticsClient, analyticsClient } from './analytics-node-client/src/client.min.js';
-import { Product, ProductJira } from './atlclients/authInfo';
+import { isBasicAuthInfo, Product, ProductJira } from './atlclients/authInfo';
 import { CredentialManager } from './atlclients/authStore';
 import { ClientManager } from './atlclients/clientManager';
 import { LoginManager } from './atlclients/loginManager';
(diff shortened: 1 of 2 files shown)


src/container.ts
```diff
@@ -483,6 +483,23 @@ export class Container {
         return this._isRovoDevEnabled;
     }
 
+    public static async hasValidRovoDevCredentials(): Promise<boolean> {
```
marcomura: there are many conditions that can cause Rovo Dev to be disabled. I think a better way to be sure if hooking up on RovoDevWebViewProvider's ProcessState, and check that it's `Started`

ostoliar-atlassian: @marcomura 
Thanks.
Done.

src/container.ts
```diff
@@ -481,6 +481,10 @@ export class Container {
         return this._isRovoDevEnabled;
     }
 
+    public static isRovoDevReady(): boolean {
+        return this._isRovoDevEnabled && this._rovodevWebviewProvider && !this._rovodevWebviewProvider.isDisabled;
```
marcomura, line 485: I think you meant `this._rovodevWebviewProvider.isReady` instead of `!this._rovodevWebviewProvider.isDisabled` 😆

ostoliar-atlassian: @marcomura 
isReady method doesn't exist.

The available methods are:
- isVisible() - checks if webview is visible
- isStarted() - checks if process state is Started
- isDisabled() - checks if process state is Disabled

The current implementation checks:
- this._isRovoDevEnabled - Rovo Dev is enabled in the extension
- this._rovodevWebviewProvider - Provider exists
- !this._rovodevWebviewProvider.isDisabled - Rovo Dev is not explicitly disabled

We want to show the checkbox when:
- Rovo Dev is enabled
- Provider exists
- Rovo Dev is not disabled (could be NotStarted, Starting, or Started)

**Probably you are thinking of isStarted() instead of isReady()**, but that would be too restrictive because:
- isStarted() only returns true when _processState === RovoDevProcessState.Started
- This would hide the checkbox during initialization (NotStarted, Starting states)
- Users wouldn't see the option until Rovo Dev is fully initialized

marcomura: In this case, I think you should exclude both `Disabled` and `Terminated`.

I'll polish the change myself and merge.
Thanks

PR #1064, opened 2025-09-30
AXON-1273: Running YOLO should be sticky on a per repo basis

ostoliar-atlassian:
### Ticket:
https://softwareteams.atlassian.net/browse/AXON-1273

### What Is This Change?
Running YOLO should be sticky on a per repo basis.
Different repositories can have different YOLO mode settings. The implementation ensures that YOLO mode is now sticky on a per-repository basis. Users can enable/disable YOLO mode in different repositories independently, and the setting will be remembered the next time they use Rovo Dev in that repository.

### Demo:
https://www.loom.com/share/00886c87404540c49bd94f53b7f3fc8c

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -102,6 +102,26 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
+
+    private getYoloModeStorageKey(): string {
+        const workspaceRoot = this.getWorkspaceRoot();
+        return workspaceRoot ? `yoloMode_${workspaceRoot}` : 'yoloMode_global';
+    }
+
+    private async loadYoloModeFromStorage(): Promise<boolean> {
+        const key = this.getYoloModeStorageKey();
+        const stored = this._context.workspaceState.get<boolean>(key);
+        return stored ?? this.isBoysenberry;
```
marcomura: this logic should be bypassed altogether when in boysenberry

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -102,6 +102,26 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
+    }
+
+    private getYoloModeStorageKey(): string {
+        const workspaceRoot = this.getWorkspaceRoot();
+        return workspaceRoot ? `yoloMode_${workspaceRoot}` : 'yoloMode_global';
+    }
+
+    private async loadYoloModeFromStorage(): Promise<boolean> {
+        const key = this.getYoloModeStorageKey();
+        const stored = this._context.workspaceState.get<boolean>(key);
```
marcomura, line 117: should we add a configuration item to store which repos are in YOLO mode?
@sdzh-atlassian

sdzh-atlassian: Hmm, good question! I think Adam might have an opinion on this :) For my 5 cents - IMO keeping it simple is probably better?

Like, a sticky setting is a pretty familiar thing, but if I'm a user and it starts toggling on and off by itself, e.g. when I change workspaces - I might think that something is broken, even if there are actually reasonable rules under the hood

I'm honestly happy with either option though

sdzh-atlassian: Let's maybe have it as a global thing for now, and ask design for long-term advice?
That is, of course, if I'm reading the proposed change right :)

ostoliar-atlassian: Thanks @sdzh-atlassian 
Now using global storage.

src/rovo-dev/rovoDevWebviewProvider.ts
ostoliar-atlassian: @marcomura 
Thanks done.

marcomura (approved):

PR #1069, opened 2025-09-30
Bug fix: Move logic to @atlassian.com in the other place too for Rovo Dev (no longer using Rovo Dev FF)

bwieger-atlassian-com:
### What Is This Change?
Rachit was experiencing an issue in cursor where the rovo dev ff was resolving to false. I believe this will fix his issue. 

### How Has This Been Tested?
Manually

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -258,13 +258,7 @@ export class Container {
 
         this._onboardingProvider = new OnboardingProvider();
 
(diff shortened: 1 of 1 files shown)


sdzh-atlassian (approved):
LGTM

marcomura (approved):

PR #1072, opened 2025-09-30
[AXON-1276] instrument issue suggestions

sdzh-atlassian:
### What Is This Change?

Analytics events for issue suggestions. We'll now be able to see:
 * When the user is nudged to authenticate
 * When the user successfully calls suggestion API
 * When the user calls suggestion API but something goes wrong
 * When the user toggles the suggestion checkbox

What's not here - advanced stuff like if the suggestions were edited before the issue being submitted (there are more pressing matters right now than navigating the UI implementation 🤦)

### How Has This Been Tested?

TBD

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`
- [ ] Events actually created, FeelsBadMan


src/vscAnalyticsApi.ts
```diff
@@ -351,4 +356,24 @@ export class VSCAnalyticsApi implements AnalyticsApi {
             this._analyticsClient.sendTrackEvent(e);
         });
     }
+
+    async fireIssueSuggestionGeneratedEvent() {
+        const event = await issueSuggestionGeneratedEvent();
```
sdzh-atlassian, line 361: changed up the chaining pattern we have above, since the functions are async. Didn't oneline for easier debugging

src/vscAnalyticsApi.ts
```diff
@@ -351,4 +356,24 @@ export class VSCAnalyticsApi implements AnalyticsApi {
...
+        this._analyticsClient.sendTrackEvent(event);
+    }
+
+    async fireIssueSuggestionFailedEvent({ error }: { error: string }) {
```
marcomura, line 365: we should define a guideline for args, because defining them like this for a single argument is more verbose than useful

sdzh-atlassian: ah, I've had some more stuff in there but nuked it, my bad :/ I'll keep it for now if that's OK

sdzh-atlassian:
Created the events and tested that stuff is being sent locally. Sadly, I'm 0% more confident around stuff actually showing up in BI as expected lol

PR #1077, opened 2025-10-01
[AXON-1277] Fix tool permission dialog not disappearing, and update Rovo Dev tool-call interface

marcomura:
### What Is This Change?

The new tool call interface includes the `mcp_server` field when the tool is invoked from an MCP server.

Rovo Dev also introduced a new tool call named `mcp_invoke_tool` which represent the invocation of a tool from an MCP server:
image

This change also fixes the issue:
[AXON-1277] Update Rovo Dev tool-call interface, and small refactoring

This change also refactors the parser interfaces into their own file, and creates the `RovoDevToolName` type enumerating all possible tool call names.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

[AXON-1277]: https://softwareteams.atlassian.net/browse/AXON-1277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/react/atlascode/rovo-dev/rovoDevView.tsx
+++ b/src/react/atlascode/rovo-dev/rovoDevView.tsx
@@ -7,12 +7,12 @@ import { highlightElement } from '@speed-highlight/core';
 import { detectLanguage } from '@speed-highlight/core/detect';
 import { useCallback, useState } from 'react';
 import * as React from 'react';
+import { RovoDevResponse } from 'src/rovo-dev/responseParserInterfaces';
 import { ToolPermissionChoice } from 'src/rovo-dev/rovoDevApiClientInterfaces';
 import { RovoDevContextItem, State } from 'src/rovo-dev/rovoDevTypes';
 import { v4 } from 'uuid';
 
 import { DetailedSiteInfo } from '../../../atlclients/authInfo';
-import { RovoDevResponse } from '../../../rovo-dev/responseParser';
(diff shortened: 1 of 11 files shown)


src/react/atlascode/rovo-dev/tools/ToolCallItem.tsx
```diff
@@ -29,12 +30,12 @@ export const ToolCallItem: React.FC<{
     );
 };
 
-export function parseToolCallMessage(msgToolName: string): string {
+export function parseToolCallMessage(msgToolName: RovoDevToolName): string {
```
sdzh-atlassian, line 33: Nit: not for this PR, but this should really be a something like this:

```
const friendlyName = friendlyNames[msgToolName] ?? msgToolName
```

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -231,6 +231,7 @@ const RovoDevView: React.FC = () => {
                     ) {
                         finalizeResponse();
                     }
+                    setModalDialogs([]);
```
sdzh-atlassian, line 234: Is this an unrelated bugfix?

sdzh-atlassian (approved):
ship it

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: This is the actual fix for the AXON ticket lol

PR #1080, opened 2025-10-01
Simplify Rovo Dev messaging to webview

marcomura:
### What Is This Change?

Merged the communication for several type of messages into one, to prepare a future change where messages can be sent in bulk (== replay optimization). 

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/react/atlascode/rovo-dev/rovoDevView.tsx
+++ b/src/react/atlascode/rovo-dev/rovoDevView.tsx
@@ -7,7 +7,6 @@ import { highlightElement } from '@speed-highlight/core';
 import { detectLanguage } from '@speed-highlight/core/detect';
 import { useCallback, useState } from 'react';
 import * as React from 'react';
-import { RovoDevResponse } from 'src/rovo-dev/responseParserInterfaces';
 import { ToolPermissionChoice } from 'src/rovo-dev/rovoDevApiClientInterfaces';
 import { RovoDevContextItem, State } from 'src/rovo-dev/rovoDevTypes';
 import { v4 } from 'uuid';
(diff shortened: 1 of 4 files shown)


src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -96,12 +95,9 @@ const RovoDevView: React.FC = () => {
...
 
-    const removeModifiedFileToolReturns = useCallback(
-        (files: ToolReturnParseResult[]) => {
-            setTotalModifiedFiles((prev) => prev.filter((x) => !files.includes(x)));
-        },
-        [setTotalModifiedFiles],
-    );
+    const removeModifiedFileToolReturns = useCallback((files: ToolReturnParseResult[]) => {
+        setTotalModifiedFiles((prev) => prev.filter((x) => !files.includes(x)));
+    }, []);
```
marcomura, line 100: React guarantees that state setters have stable references, so it's impossible for the set status function to change, and thus no need to list it as a dependency.

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -209,17 +204,32 @@ const RovoDevView: React.FC = () => {
...
+                            tool_name: object.tool_name,
+                            content: object.content || '',
+                            parsedContent: object.parsedContent,
+                            tool_call_id: object.tool_call_id, // Optional ID for tracking
+                            args: object.toolCallMessage.args,
+                        };
+
+                        setPendingToolCallMessage(DEFAULT_LOADING_MESSAGE); // Clear pending tool call
+                        handleAppendResponse(returnMessage);
                     }
```
sdzh-atlassian, line 233: NIT: would we benefit from a final `else` here? Even if it's just a console log about the unknown event kind?

marcomura: there should be one, but actually we are not processing the 'retry-prompt' message and I need to ask Christian if it's a bug or intentional

PR #1081, opened 2025-10-01
[AXON-830] Add feedback collection from issue suggestions

sdzh-atlassian:
### What Is This Change?

* Implement a somewhat generic JSD sender for feeature feedback that works with a new queue
image

* Add feedback form to the create issue page, in line with Slack design
image

Coming in a follow-up:
 * some rewords
 * hiding the thumbs buttons after feedback is sent
 * using an actual queue instead of a test one (only requires some ID chages)

### How Has This Been Tested?

See images above :)

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/feedback/JSDSubmitter.ts
+++ b/src/feedback/JSDSubmitter.ts
@@ -1,0 +1,110 @@
+import truncate from 'lodash.truncate';
+import { version } from 'vscode';
+
+import { ProductBitbucket, ProductJira } from '../atlclients/authInfo';
+import { Container } from '../container';
+import { getAgent, getAxiosInstance } from '../jira/jira-client/providers';
+
+type FeatureFeedbackData = {
+    description: string;
+    feature: string;
+    email: string;
+};
+
+type FeedbackPayload = {
+    summary?: string;
+    context?: string;
+};
+
+class JSDSubmitter<FeedbackData extends FeatureFeedbackData, FullPayload = FeedbackData & FeedbackPayload> {
+    private readonly widgetId: string;
+    private readonly requestTypeId: number;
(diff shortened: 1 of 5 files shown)


src/feedback/JSDSubmitter.ts
```diff
@@ -0,0 +1,110 @@
...
+    feature: string;
+    email: string;
+};
+
+type FeedbackPayload = {
+    summary?: string;
+    context?: string;
+};
+
+class JSDSubmitter<FeedbackData extends FeatureFeedbackData, FullPayload = FeedbackData & FeedbackPayload> {
```
sdzh-atlassian, line 19: I tried making this guy a bit more generic - the 2 other implementations we have for this have everything hardcoded. This isn't perfect, but hopefully an improvement? idk

src/feedback/JSDSubmitter.ts
```diff
@@ -0,0 +1,110 @@
...
+    private readonly widgetId: string;
+    private readonly requestTypeId: number;
+    private readonly fieldMapper: Record<keyof FullPayload, string>;
+    private readonly valueMapper?: Partial<Record<keyof FullPayload, (value: FullPayload[keyof FullPayload]) => any>>;
+
+    constructor({
```
marcomura, line 25:
```
constructor(
    private readonly widgetId: string,
    private readonly requestTypeId: number,
    private readonly fieldMapper: Record<keyof FullPayload, string>,
    private readonly valueMapper?: Partial<Record<keyof FullPayload, (value: FullPayload[keyof FullPayload]) => any>>,
) { }
```

marcomura (approved):
lgtm

PR #1093, opened 2025-10-02
[AXON-1301] retain API token credentials through OAuth

sdzh-atlassian:
### What Is This Change?

Up until now, we've been overwriting API tokens when logging in with OAuth. Our new way of doing things suggests that API token credential is actually preferred, so this change changes the logic so that the API token is retained, instead

Changes:
 * Exclude sites which are already using an API token from OAuth response
 * Fix the logic where we merge all cloud sites after OAuth login to exclude sites with API tokens
 * Add analytics

### How Has This Been Tested?

Manually - please see #axon for the demos

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -570,6 +570,10 @@ export async function apiTokenNudgeClickedEvent(source: string): Promise<TrackEv
     return trackEvent('clicked', 'apiTokenNudge', { attributes: { source } });
 }
 
+export async function apiTokenRetainedEvent() {
+    return trackEvent('retained', 'apiToken');
+}
+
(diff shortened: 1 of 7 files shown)


src/analytics.ts
```diff
@@ -570,6 +570,10 @@ export async function apiTokenNudgeClickedEvent(source: string): Promise<TrackEv
     return trackEvent('clicked', 'apiTokenNudge', { attributes: { source } });
 }
 
+export async function apiTokenRetainedEvent() {
+    return trackEvent('retained', 'apiToken');
```
sdzh-atlassian, line 574: Still need to create this event if we want it

marcomura (approved):
I trust you have tested it thoroughly 😄

PR #1108, opened 2025-10-04
[AXON-1322] Fixed process state transitions, chat history polluted after Disabled state, and bug during initialization

marcomura:
### What Is This Change?

Fixed:
- AXON-1322: Logging out from API token when Rovo Dev is in disabled states doesn't trigger the Login page
- Chat history was not cleaned up when Rovo Dev transitioned to disabled state
- Incorrect condition during initialization caused Rovo Dev to initialize with some delay

Also:
- Some clean ups
- Renamed the ugly `Service at at "${baseApiUrl}" wasn't ready within ${timeout} ms` error message to a simpler `Rovo Dev service is unreachable.`

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -66,6 +66,16 @@ enum RovoDevProcessState {
     Disabled,
 }
 
+// this map sets the priority level of Disabled states
+// higher priority can override lower priority
+const RovoDevDisabledPriority: Record<RovoDevDisabledReason | 'none', number> = {
+    none: 0,
+    Other: 1,
+    EntitlementCheckFailed: 2,
+    NeedAuth: 3,
+    NoWorkspaceOpen: 4,
+};
+
 export class RovoDevWebviewProvider extends Disposable implements WebviewViewProvider {
     private readonly viewType = 'atlascodeRovoDev';
     private readonly isBoysenberry = process.env.ROVODEV_BBY === 'true';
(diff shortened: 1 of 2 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -1279,11 +1293,21 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
     // Disabled and Terminated states are almost identical, except that
     // with Terminated you can restart Rovo Dev with the [+] button,
     // and with Disabled you can't.
-    private setRovoDevTerminated(processState: RovoDevProcessState.Disabled | RovoDevProcessState.Terminated) {
-        this._processState = processState;
+    private setRovoDevTerminated(processState: RovoDevProcessState.Terminated): Promise<void>;
```
sdzh-atlassian, line 1296: At this point there should really just be several methods for this IMO

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -1292,4 +1316,11 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
 
         return this.refreshDebugPanel();
     }
+
+    private setProcessState(processState: Exclude<RovoDevProcessState, RovoDevProcessState.Disabled>): void;
```
sdzh-atlassian, line 1320: Same here. Like, is this really better than something like:
```
setProcessState(value: RovoDevProcessState) { /*...*/ } // with Exclude if we want to
setProcessStateDisabled(reason: RovoDevDisabledReason) { /*...*/ }
```
?

Obviously NIT, etc etc

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: Probably this would benefit from two different methods, since the possible states are only 2

marcomura: Except for the first `if`, the body is the same. If I separate the methods I would have to create an additional implementation to share the common code (if I don't duplicate).
I think overloading is better.

src/rovo-dev/rovoDevWebviewProvider.ts
marcomura: Probably personal preference, but I think overloading is better

PR #1109, opened 2025-10-05
[AXON-1010] Batch Rovo Dev replay messages and refactoring of webview message interfaces

marcomura:
### What Is This Change?

With this change we are disabling the streaming of the `replay` API's response, and sending batches of messages to the webview for a faster rendering.

This change also refactors the interfaces used to represent Rovo Dev responses in the webview.
Specifically, it reuses the same parsed interfaces for `text`, `tool-call`, and `tool-return` messages.

The benefits can be seen in `utils.tsx` where lots of re-definition of interfaces are now not necessary anymore.
Also, this bring stronger typing when consuming the data.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`


src/react/atlascode/rovo-dev/common/common.tsx
```diff
@@ -78,21 +76,21 @@ export const renderChatHistory = (
...
-                source: 'RovoDev',
-            };
-            return (
-                <ChatMessageItem
-                    msg={retryMsg}
-                    icon={<StatusErrorIcon color="var(--ds-icon-danger)" label="error-icon" spacing="none" />}
-                    openFile={openFile}
-                />
-            );
+        // case 'RovoDevRetry':
```
marcomura: this branch seems unused

sdzh-atlassian: Probably best to either delete, or leave a comment about this in the code itself IMO

marcomura: agreed - removed

src/rovo-dev/rovoDevChatProvider.ts
```diff
@@ -339,49 +411,54 @@ export class RovoDevChatProvider {
...
             case 'warning':
-                return webview.postMessage({
+                await webview.postMessage({
                     type: RovoDevProviderMessageType.ShowDialog,
                     message: {
                         type: 'warning',
                         text: response.message,
                         title: response.title,
-                        source: 'RovoDevDialog',
+                        event_kind: '_RovoDevDialog',
```
sdzh-atlassian, line 434: Looks like we're introducing a rule for naming these which isn't super obvious right away. Are we using the leading underscore here to indicate a generated event?

IMO it could be good to document this logic somewhere - unless it's already done and I missed it :D

src/rovo-dev/rovoDevChatProvider.ts
```diff
@@ -402,8 +480,8 @@ export class RovoDevChatProvider {
                         return webview.postMessage({
                             type: RovoDevProviderMessageType.ShowDialog,
                             message: {
+                                event_kind: '_RovoDevDialog',
```
sdzh-atlassian, line 483: Not for this PR, but we should probably streamline the naming convention some time in the future (e.g. account for pyhton-y API responses)

src/rovo-dev/rovoDevWebviewProviderMessages.ts
```diff
@@ -51,19 +46,19 @@ export type RovoDevDisabledReason = DisabledState['subState'];
 
 export type RovoDevEntitlementCheckFailedDetail = EntitlementCheckRovoDevHealthcheckResponse['detail'];
 
-export type RovoDevResponseMessageType =
-    | RovoDevTextResponse
-    | RovoDevToolCallResponse
-    | RovoDevToolReturnResponse
-    | RovoDevRetryPromptResponse;
+export type RovoDevResponseMessageType = RovoDevTextResponse | RovoDevToolCallResponse | RovoDevToolReturnResponse;
+//| RovoDevRetryPromptResponse;
```
sdzh-atlassian, line 50: Same here, let's maybe kill this completely or document it?

src/rovo-dev/rovoDevChatProvider.ts
marcomura: yeah, event_kind without underscore = coming from Rovo Dev, with underscore = generated by us.

You are right, it should be documented and it's not.
Will do in another PR :) 
Thanks!

src/rovo-dev/rovoDevWebviewProviderMessages.ts
marcomura: damn it - I'll clean it up in a separate PR

PR #1113, opened 2025-10-06
[AXON-1328] Re-introduce rovodev feature flag check

sdzh-atlassian:
### What Is This Change?

Re-introduce rovodev feature flag check to work as an alternative condition to the internal user check.
Effectively returns what was removed here:
https://github.com/atlassian/atlascode/pull/1069

### How Has This Been Tested?

TBD

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -290,9 +290,11 @@ export class Container {
     }
 
     private static async refreshRovoDev(context: ExtensionContext) {
-        const isJiraEnabledAndAtlassianUser = this.config.jira.enabled && (await this.isAtlassianUser(ProductJira));
+        const isJiraAndRovodevEnabled =
+            this.config.jira.enabled &&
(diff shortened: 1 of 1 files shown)


src/container.ts
```diff
@@ -290,9 +290,11 @@ export class Container {
     }
 
     private static async refreshRovoDev(context: ExtensionContext) {
-        const isJiraEnabledAndAtlassianUser = this.config.jira.enabled && (await this.isAtlassianUser(ProductJira));
+        const isJiraAndRovodevEnabled =
```
bwieger-atlassian-com, line 293: This var can simply be 

```
this.config.jira.enabled
```

sdzh-atlassian: Introduced the alternative solution in https://github.com/atlassian/atlascode/pull/1115

PR #1114, opened 2025-10-06
[AXON-1324] Update Rovo Dev, adjust tool permission flow, and enablement strings

marcomura:
### What Is This Change?

Upgraded Rovo Dev to 0.12.1.

This new version includes breaking changes in:

- The tool permissions, adding
  - A field to indicate if permissions are required
  - A map to indicate which tools requires permission, which ones are auto-allowed, and which ones are auto-denied
  - A new arg to specify the permissions in settings should be respected
  - NOTE: now the message `on_call_tools_start` is always coming for every `tool-call`

- Entitlement strings
  - Moved the strings inside the data's payload

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     ],
     "main": "./build/extension/extension",
     "rovoDev": {
-        "version": "0.12.0"
+        "version": "0.12.1"
     },
     "scripts": {
         "vscode:uninstall": "node ./build/extension/uninstall.js",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     ],
     "main": "./build/extension/extension",
     "rovoDev": {
-        "version": "0.12.0"
+        "version": "0.12.1"
```
sdzh-atlassian, line 48: One would hope that breaking changes wouldn't happen in this kind of a version bump, but I guess not this time, FeelsBadMan :D

sdzh-atlassian, line 48: As a general note - before merging, please consider if this is something we need _today_ or if this can wait

PR #1115, opened 2025-10-06
[AXON-1328] remove rovodev ff check

sdzh-atlassian:
### What Is This Change?

An alternative to https://github.com/atlassian/atlascode/pull/1113 where we remove the check instead

> Note: Pending confirmation with the team

### How Has This Been Tested?

TBD 

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,11 +1,12 @@
 {
     "name": "atlascode",
-    "displayName": "Atlassian: Jira & Bitbucket",
-    "description": "Bringing the power of Jira and Bitbucket to VS Code - With Atlassian for VS Code you can create and view issues, start work on issues, create pull requests, do code reviews, start builds, get build statuses and more!",
+    "displayName": "Atlassian: Jira, Rovo Dev, Bitbucket",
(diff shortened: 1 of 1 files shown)


bwieger-atlassian-com (approved):

src/container.ts
```diff
@@ -290,9 +290,8 @@ export class Container {
     }
 
     private static async refreshRovoDev(context: ExtensionContext) {
-        const isJiraEnabledAndAtlassianUser = this.config.jira.enabled && (await this.isAtlassianUser(ProductJira));
         const isBoysenberryMode = !!process.env.ROVODEV_BBY;
-        const shouldEnableRovoDev = isJiraEnabledAndAtlassianUser || isBoysenberryMode;
+        const shouldEnableRovoDev = this.config.jira.enabled || isBoysenberryMode;
```
marcomura, line 294: do we have an event listener for this, that disables Rovo Dev when Jira gets disabled?

PR #1132, opened 2025-10-07
[AXON-1265] Add 'Allow all' and 'Enable YOLO mode' for mutiple tool requests

cabella-dot:
### What Is This Change?

Add new Dropdown Button component which takes a main action and a set of dropdown list items.

Used new component to add an Allow all/enable yolo mode button which appears in chat when there are mutiple tool permission requests 

DEMO: https://www.loom.com/share/cd78e8664adc4b0db3cd7591b6e86f44

### How Has This Been Tested?

Manually

Added UT for new component

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/react/atlascode/rovo-dev/rovoDevView.tsx
+++ b/src/react/atlascode/rovo-dev/rovoDevView.tsx
@@ -7,6 +7,7 @@ import { highlightElement } from '@speed-highlight/core';
 import { detectLanguage } from '@speed-highlight/core/detect';
 import { useCallback, useState } from 'react';
 import * as React from 'react';
+import { RovoDevToolReturnResponse } from 'src/rovo-dev/responseParserInterfaces';
 import { RovoDevContextItem, State, ToolPermissionDialogChoice } from 'src/rovo-dev/rovoDevTypes';
 import { v4 } from 'uuid';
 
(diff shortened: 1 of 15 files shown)


src/react/atlascode/rovo-dev/rovoDevViewMessages.tsx
marcomura, line 41: I think you can define another type here for `ToolPermissionChoice | 'allowAll'` since it's repeated 3 or 4 times already

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -667,19 +667,26 @@ const RovoDevView: React.FC = () => {
     }, []);
 
     const onToolPermissionChoice = useCallback(
-        (toolCallId: string, choice: ToolPermissionChoice) => {
+        (toolCallId: string, choice: ToolPermissionChoice | 'allowAll' | 'enableYolo') => {
             // remove the dialog after the choice is submitted
-            setModalDialogs((prev) =>
-                prev.filter((x) => x.type !== 'toolPermissionRequest' || x.toolCallId !== toolCallId),
-            );
+            if (choice === 'enableYolo' || choice === 'allowAll') {
```
marcomura: if the option is 'enableYolo', you can send a `YoloModeToggled` message instead

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -667,19 +667,26 @@ const RovoDevView: React.FC = () => {
...
+                if (choice === 'enableYolo' && !isYoloModeToggled) {
```
marcomura: no need to check for `isYoloModeToggled`, and you can remove it as dependency

marcomura (approved):
lgtm!

PR #1148, opened 2025-10-09
AXON-1196: optimize chatstream logic

cabella-dot:
### What Is This Change?

This change optimizes the chatstream logic by:

1. Seperating complicated logic into seperate components
2. Memoizing these complicated components to avoid unnecessary rerenders
3. Memoizing complicated values to avoid unnecessary computations

Note: This ticket was to see if we can improve the performance of Rovo dev chat through React optimization. Although this does remove unnecessary rerenders, this does not answer the question of why it runs slow in certain BBY containers. I believe it is a necessary improvement for the chat nonetheless.

### How Has This Been Tested?

Manually

Added UTs for new components

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
+++ b/src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
@@ -219,21 +219,15 @@ export const ChatStream: React.FC<ChatStreamProps> = ({
         [setFeedbackVisible],
     );
 
-    const isChatHistoryDisabled = React.useMemo(
-        () =>
-            (currentState.state === 'Initializing' && currentState.subState === 'MCPAcceptance') ||
-            (currentState.state === 'Disabled' && currentState.subState !== 'Other'),
-        [currentState],
-    );
+    const isChatHistoryDisabled =
+        (currentState.state === 'Initializing' && currentState.subState === 'MCPAcceptance') ||
+        (currentState.state === 'Disabled' && currentState.subState !== 'Other');
 
-    const shouldShowToolCall = React.useMemo(
-        () =>
-            currentState.state !== 'Disabled' &&
-            currentState.state !== 'ProcessTerminated' &&
-            currentState.state !== 'WaitingForPrompt' &&
-            (currentState.state !== 'Initializing' || currentState.isPromptPending),
-        [currentState],
-    );
+    const shouldShowToolCall =
+        currentState.state !== 'Disabled' &&
+        currentState.state !== 'ProcessTerminated' &&
+        currentState.state !== 'WaitingForPrompt' &&
+        (currentState.state !== 'Initializing' || currentState.isPromptPending);
 
     return (
         <div ref={chatEndRef} className="chat-message-container">


src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -291,8 +291,10 @@ export const appendResponse = (
     if (!Array.isArray(latest)) {
         // Streaming text response, append to current message
         if (latest?.event_kind === 'text' && response?.event_kind === 'text') {
-            latest.content += response.content;
-            return [...prev, latest];
+            const appendedMessage = { ...latest };
```
marcomura, line 294: this will create a new object instead of reusing the existing one, so it goes against optimizations. why was it changed?

cabella-dot: Oh I must have missed this. This was intended to force the chatHistory to be changed so the components with this dependency render with the new content. I will check if it works with the old impl

src/react/atlascode/rovo-dev/messaging/ChatStreamMessageRenderer.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    onCollapsiblePanelExpanded={onCollapsiblePanelExpanded}
+                    renderProps={renderProps}
+                    currentState={currentState}
+                    drawerOpen={idx === openDrawerIdx}
+                />
+            );
+        });
+    },
+    (prevProps, nextProps) =>
+        prevProps.chatHistory === nextProps.chatHistory &&
```
marcomura, line 67: isn't this the same as doing `React.useMemo(<>, [currentState, handleCopyResponse, handleFeedbackTrigger, onToolPermissionChoiceonCollapsiblePanelExpanded, renderProps])` ?

src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
```diff
@@ -222,15 +219,21 @@ export const ChatStream: React.FC<ChatStreamProps> = ({
...
+            (currentState.state === 'Disabled' && currentState.subState !== 'Other'),
+        [currentState],
+    );
 
-    const shouldShowToolCall =
-        currentState.state !== 'Disabled' &&
-        currentState.state !== 'ProcessTerminated' &&
-        currentState.state !== 'WaitingForPrompt' &&
-        (currentState.state !== 'Initializing' || currentState.isPromptPending);
+    const shouldShowToolCall = React.useMemo(
```
marcomura: I previously written this without memo because being a simple evaluation, and of a primitive type, it doesn't actually bring any advantage to have it memoized.

marcomura: (same for isChatHistoryDisabled)

src/react/atlascode/rovo-dev/messaging/ChatStreamMessageRenderer.tsx
cabella-dot: For a memoized react component, it needs a propAreEqual function to determine if it needs to rerender

src/react/atlascode/rovo-dev/utils.tsx
cabella-dot: Because the modification of `chatHistory` is needed to trigger a rerender, only modifying a part of the object will not enable the streaming effect. Although this is creating a new object, it forces `chatHistory` and its dependencies to recognize it has been changed

PR #1166, opened 2025-10-14
Fix error-state messages in Boysenberry

marcomura:
### What Is This Change?

Error state messages displayed when the Rovo Dev healtcheck is not healthy were hidden by the `signalRovoDevDisabled` function.
Fixed that.

Also, re-introduced the `.focus` command to auto-focus Rovo Dev in Boysenberry.

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
+++ b/src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
@@ -18,6 +18,8 @@ import { DialogMessage, PullRequestMessage, Response, scrollToEnd } from '../uti
 import { ChatStreamMessageRenderer } from './ChatStreamMessageRenderer';
 import { DropdownButton } from './dropdown-button/DropdownButton';
 
+const IsBoysenberry = process.env.ROVODEV_BBY === 'true';
+
 interface ChatStreamProps {
     chatHistory: Response[];
     modalDialogs: DialogMessage[];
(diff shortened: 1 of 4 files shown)


src/react/atlascode/rovo-dev/messaging/ChatStream.tsx
```diff
@@ -192,16 +194,18 @@ export const ChatStream: React.FC<ChatStreamProps> = ({
...
 
     // Other state management effect
-    React.useEffect(() => {
-        if (process.env.ROVODEV_BBY && currentState.state === 'WaitingForPrompt') {
-            const canCreatePR = chatHistory.length > 0;
-            setCanCreatePR(canCreatePR);
-            if (canCreatePR) {
-                // Only check git changes if there's something in the chat
-                checkGitChanges();
+    if (IsBoysenberry) {
```
cabella-dot, line 197: How does this work? So this will mount the `useEffect` on initial render and then will continue to listen to the deps?

marcomura: this works because `IsBoysenberry` is a static immutable value, and its condition never changes during the lifetime of the webview.

When it's false, the useEffect is never mounted, like the entire block is not even there.

When it's true, the useEffect is mounted and refreshed at every render, like it's not under an if.

cabella-dot: Thanks for the explanation!

PR #1175, opened 2025-10-15
Fix 'Could not create workers' bug for monaco editor

cabella-dot:
### What Is This Change?

This disables Web Workers for the monaco editor

This gets rid of the error relating to monaco web workers

### How Has This Been Tested?

Manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


marcomura (approved):
lgtm!

PR #1196, opened 2025-10-17
AXON-1391: add setting to toggle thinking block rendering

cabella-dot:
### What Is This Change?

You can now toggle thinking block rendering in extension settings

`atlascode.rovodev.thinkingBlockEnabled`

Demo:
https://www.loom.com/share/311c5f160871460ca0d0d4aebcbbfb5c

### How Has This Been Tested?

Manually

Updated UT

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,11 +6,10 @@
 # @global-owner1 and @global-owner2 will be requested for
 # review when someone opens a pull request.
 # AXON TEAM: @bwieger-atlassian-com @sdzh-atlassian @cabella-dot @marcomura
-# DEV INNER LOOP: @akalia-atlassian @pastelsky
-# SOFTSERVE TEAM: @bhorai-atl
-# BBY FOLKS: @Blastoplex @mattcolman
+# SOFTSERVE TEAM: @bhorai-atl @om-ukr
+# BBY FOLKS: @Blastoplex @matt-lassian
 
-*       @bwieger-atlassian-com @sdzh-atlassian @cabella-dot @marcomura @akalia-atlassian @pastelsky @bhorai-atl @om-ukr @mattcolman @Blastoplex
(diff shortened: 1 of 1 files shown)


src/config/model.ts
```diff
@@ -219,6 +220,7 @@ const emptyJiraConfig: JiraConfig = {
 const emptyRovoDevConfig: RovoDevConfig = {
     enabled: false,
     debugPanelEnabled: false,
+    thinkingBlockEnabled: true,
```
marcomura: NIT: these empty thing are usually filled with falsy values even if the default is true

cabella-dot: Will switch

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -224,6 +229,15 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
         }
     }
 
+    private async refreshThinkingBlock(force?: boolean) {
+        if (this._thinkingBlockEnabled || force) {
```
marcomura: this `force` logic is necessary for the debug panel to avoid refreshing it with new data (not only when it switches) when it's disabled.

for config-only, you can always send it

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -91,6 +91,7 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
     private _debugPanelEnabled = false;
     private _debugPanelContext: Record<string, string> = {};
     private _debugPanelMcpContext: Record<string, string> = {};
+    private _thinkingBlockEnabled = true;
```
marcomura: I think you don't need to store its value

PR #1213, opened 2025-10-21
AXON-1379: Fixed blinking in  summary field

BHulovatyi:
### What Is This Change?

**Fixed blinking in the summary field**

Before/After:
https://www.loom.com/share/f7e907cc02434fdf971e4c7c367b69f7?sid=931ef8c1-02a4-4258-8865-dc25ddb18a2b

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
+++ b/src/webviews/components/issue/create-issue-screen/CreateIssuePage.tsx
@@ -181,23 +181,13 @@ export default class CreateIssuePage extends AbstractIssueEditorPage<Emit, Accep
                         fieldValues: mergedFieldValues,
                     };
 
-                    // Only reset summaryKey if project or issuetype changed (which affect summary validation)
-                    const projectChanged =
-                        fieldValues?.project && this.state.fieldValues.project?.key !== fieldValues.project.key;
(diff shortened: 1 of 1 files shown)


src/webviews/components/issue/AbstractIssueEditorPage.tsx
```diff
@@ -592,9 +592,11 @@ export abstract class AbstractIssueEditorPage<
                                 let markup = (
                                     <Textfield
                                         {...fieldArgs.fieldProps}
-                                        value={defaultVal}
                                         className="ac-inputField"
-                                        isDisabled={this.state.isSomethingLoading || this.state.isGeneratingSuggestions}
+                                        isDisabled={
```
sspanchenko, line 596: WDYT, should we add a method for this and use it in both places?

BHulovatyi: When we fully switch to the Atlaskit editor, the second part will not be needed, so I think there is no need to move it to a separate method. It will only be used once.
Now we use FF this.state.isAtlaskitEditorEnabled and show
` <JiraIssueTextAreaEditor` only when our FF disabled

PR #1214, opened 2025-10-21
[AXON-1024, AXON-1405] Support drag and drop of context (files and Jiras) into Rovo Dev chat

marcomura:
### What Is This Change?

This change introduces the ability to drag and drop resources into the Rovo Dev chat webview, so they can be attached a context to the next prompt for the agent.

This change enables dragging and dropping `files` and `Jira work items`.

As part of this change, the following changes are also applied:
- When clicking on a Jira work item in Rovo Dev suggestions, now the Jira work item is attached instead of its URL being pasted into the prompt itself
- When clicking on 'Start work with Rovo Dev` in the start work page, now the Jira work item is attached instead of its URL being pasted into the prompt itself

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

Recommendations:

### DIFF
--- a/src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
+++ b/src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
@@ -6,7 +6,7 @@ import Tooltip from '@atlaskit/tooltip';
 import React, { useCallback, useState } from 'react';
 import { RovoDevTextResponse } from 'src/rovo-dev/responseParserInterfaces';
 
-import { MarkedDown, OpenFileFunc } from '../common/common';
+import { MarkedDown, OpenFileFunc, OpenJiraFunc } from '../common/common';
 import { PromptContextCollection } from '../prompt-box/promptContext/promptContextCollection';
 import { UserPromptMessage } from '../utils';
 
@@ -17,7 +17,7 @@ export const ChatMessageItem: React.FC<{
     onCopy?: (text: string) => void;
     onFeedback?: (isPositive: boolean) => void;
     openFile: OpenFileFunc;
-    openJira: (url: string) => void;
+    openJira: OpenJiraFunc;
     onLinkClick?: (href: string) => void;
 }> = ({ msg, icon, enableActions, onCopy, onFeedback, openFile, openJira, onLinkClick }) => {
     const [isCopied, setIsCopied] = useState(false);
--- a/src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextCollection.tsx
+++ b/src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextCollection.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { RovoDevContextItem, RovoDevFileContext, RovoDevJiraContext } from 'src/rovo-dev/rovoDevTypes';
 
-import { OpenFileFunc } from '../../common/common';
+import { OpenFileFunc, OpenJiraFunc } from '../../common/common';
 import { PromptContextFileItem, PromptContextJiraItem } from './promptContextItem';
 
 // PromptContextCollection: displays a row or column of PromptContextItem
@@ -14,7 +14,7 @@ export const PromptContextCollection: React.FC<{
     onRemoveContext?: (item: RovoDevContextItem) => void;
     inChat?: boolean;
     openFile: OpenFileFunc;
-    openJira: (url: string) => void;
+    openJira: OpenJiraFunc;
 }> = ({
     content,
     direction = 'row',
--- a/src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextItem.tsx
+++ b/src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextItem.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { RovoDevFileContext } from 'src/rovo-dev/rovoDevTypes';
 
-import { OpenFileFunc } from '../../common/common';
+import { OpenFileFunc, OpenJiraFunc } from '../../common/common';
 
 const isHighContrastTheme = (): boolean => {
     return (
@@ -186,7 +186,7 @@ export const PromptContextFileItem: React.FC<
 };
 
 export const PromptContextJiraItem: React.FC<
-    { name: string; url: string; openJira: (url: string) => void } & PromptContextItemRemoveProps
+    { name: string; url: string; openJira: OpenJiraFunc } & PromptContextItemRemoveProps
 > = ({ name, url, onRemove, openJira }) => {
     return (
         <span style={chipStyle}>


src/react/atlascode/rovo-dev/messaging/ChatMessageItem.tsx
```diff
@@ -16,9 +16,10 @@ export const ChatMessageItem: React.FC<{
     enableActions?: boolean;
     onCopy?: (text: string) => void;
     onFeedback?: (isPositive: boolean) => void;
-    openFile?: OpenFileFunc;
+    openFile: OpenFileFunc;
+    openJira: (url: string) => void;
```
sdzh-atlassian: `OpenJiraFunc`?

src/react/atlascode/rovo-dev/messaging/ChatStreamMessageRenderer.tsx
```diff
@@ -13,6 +13,7 @@ interface ChatStreamMessageRendererProps {
     handleFeedbackTrigger: (isPositive: boolean) => void;
     renderProps: {
         openFile: OpenFileFunc;
+        openJira: OpenJiraFunc;
```
sdzh-atlassian, line 16: Not for the scope of this PR - I wonder if there's a refactoring here that would let us _not_ pass things like this through a bunch of components. A context, maybe one for `postMessage` down the line? idk 🤔

src/react/atlascode/rovo-dev/prompt-box/prompt-input/PromptInput.tsx
```diff
@@ -22,6 +22,8 @@ import {
     SLASH_COMMANDS,
 } from './utils';
 
+const IsBoysenberry = process.env.ROVODEV_BBY === 'true';
```
sdzh-atlassian, line 25: Why is this change here?

marcomura: unrelated to this change actually, I just wanted to bring consistency how we check for Boysenberry

src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextCollection.tsx
```diff
@@ -71,6 +74,15 @@ export const PromptContextCollection: React.FC<{
                     openFile={openFile}
                 />
             ))}
+            {jiraItems.map((item, index) => (
```
sdzh-atlassian, line 77: QQ:
 * Can there actually be more than one of these?
 * Do we intend to render them _after_ files in all cases?

marcomura: 1) yes, it's possible to add more than one jira issue
2) I've chosen so, but I'm open to different perspectives

src/react/atlascode/rovo-dev/prompt-box/promptContext/promptContextItem.tsx
```diff
@@ -10,6 +10,36 @@ const isHighContrastTheme = (): boolean => {
     );
 };
 
+const JiraLogo: React.FC<{
```
sdzh-atlassian, line 13: NIT: might be best to put into its own file?

marcomura: actually I'm hoping I can fetch it from an svg resource in the future 😄

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -718,10 +745,14 @@ const RovoDevView: React.FC = () => {
         [postMessage],
     );
 
-    const onJiraItemClick = useCallback((issue: MinimalIssue<DetailedSiteInfo>) => {
-        const template = createRovoDevTemplate(issue.key, issue.siteDetails);
-        setPromptText(template);
-    }, []);
+    const onJiraItemClick = useCallback(
```
sdzh-atlassian, line 748: How is this intended to work in the new logic - wouldn't we feed the jira items into the context, instead? Not sure I quite follow where the `setPromptText` comes in

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -509,7 +487,7 @@ const RovoDevView: React.FC = () => {
                 type: RovoDevViewResponseType.Prompt,
                 text,
                 enable_deep_plan: isDeepPlanToggled,
-                context: promptContextCollection.filter((x) => x.enabled),
+                context: promptContextCollection,
```
sdzh-atlassian, line 490: Quick sanity check - might this cause us to always include the currently viewed file into context? Not sure if that's the intent of the change

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: Note, this is for the Jira suggestions that we show when the chat is empty.

The Context is now managed by the `rovoDevChatContextProvider`, so we have to communicate back that a new item needs to be added.
The provider will send a new message to the view to update the context collection.

The prompt is changed directly by the view.

src/react/atlascode/rovo-dev/rovoDevView.tsx
```diff
@@ -774,103 +805,123 @@ const RovoDevView: React.FC = () => {
...
-                    <UpdatedFilesComponent
-                        modifiedFiles={totalModifiedFiles}
-                        onUndo={undoFiles}
-                        onKeep={keepFiles}
-                        openDiff={openFile}
-                        actionsEnabled={currentState.state === 'WaitingForPrompt'}
-                        workspacePath={workspacePath}
-                        homeDir={homeDir}
+            >
+                <div id="rovoDevDragDropOverlayMessage">
```
sdzh-atlassian, line 825: NIT: should probably be a component, with `id` as a constant?

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: This is intended. Since now the context is managed by `rovoDevChatContextProvider`, I've moved all logic there.

Technically speaking the view doesn't even need to echo back the context since the provider already has it, but I left this refactoring for another PR.
This PR is already too big 😄

src/react/atlascode/rovo-dev/messaging/ChatStreamMessageRenderer.tsx
marcomura: Yeah, I was pretty annoyed too that I had to change so many files just to pass the new `openJira` func around.
Definitely a refactoring is needed here.

src/react/atlascode/rovo-dev/utils.tsx
```diff
@@ -349,3 +349,22 @@ export const appendResponse = (
...
+
+export async function processDropDataTransferItems(
+    items: DataTransferItemList,
+    callback: (value: string[]) => void,
+): Promise<void> {
+    if (!items || items.length === 0) {
+        return;
+    }
+
+    const promises: Promise<string>[] = [];
```
sdzh-atlassian, line 361: NIT: do we have a for loop here because the array is a special type?
If so, this should work:

```ts
const values = Promise.all(
    Array.from(items).map(x => new Promise((resolve) => x.getAsString(resolve))
);
```

(disclaimer - only tried it in a [playground](https://developer.mozilla.org/en-US/play?uuid=fc8dc8107875894113d99bc1a677c8636a78b248&state=tVXtbts2FH2VC67IZMCTtv10IwNdE2ABMmxI0j8FMZQRryUuFEmQ13aCwECeJg%2FWJylISo6zpInbYfpjizr38Jz7Qd6yjnrNZuxQqtWcG4BDB0rWnAW79A1yBtKLthWXGmvOyC%2BRs4QDOEeNDQF1KgBq7NHQNKFBEZAF6hCOvHXw0RoEYWRcMeBRowiYPodEoazJjGSht6v8aWAsk6bKzbk5rLLGKDVrJOFbJM7m221GDJuyJgQ2YxF7Gzl64VtlZvAz9m%2FjuxNSKtPO4Ne0sOGGmx%2By6RzQWG39DC71ElPApfUS%2FQx%2BcdcQrFYSLrVorraxWUyOfRHLpuyfKK2xJhAMW9YgbbNMjluk42z%2Bt5sTWTyUYvKWmxw07PVS0JibGMRNpiiFlMcrNHSqAqFBX3AWCxZIeOJsCgWuJlDPxwSYYDWW2rYD7jzjMiVAVcE7KR81wI8BlByLn3rBiRtthYQwrlkHnTBSo4e10nogujJ2DetONd1IFVmETGSKApBHHFURSEEimoAacFXGtwsvTFigLxVhH1LFRlC0XeCqzCkplZwCZ4TXVDktlHliKNgewVKHHhLZc36e8HN26OZ%2FCS9aL1z3%2Be4%2BNS3bbhXnLO%2F0JLAjcrOqWq%2FXJV6L3mksrW93Ypde%2FaRVGFK%2F2aekaOQeBT2OqKxp77wOiXqvUXgQ5gY89kIZZdqcoBj0yGQTkcWO8qEOzyq37hXZRwmSNeOqdB4jxxEuxFJTsVPJU2sdUOftsu22redQDjWN55G2bfyi%2FFb0wnoocioiDOzi%2BTxMsjQAtYAirpRXykio6xo4C%2BSVaTmDg4PEUtKNw7IX1HRF9XesKOe59arJlihpvoijlDZWIWkeBt1YiSMsMbZI78J52qcowkOu8vPQ68I5NPJ9p7QsvnZWhEnOZnw2498NoA74%2FfZiu%2F%2Fb3dHYHluLv1%2F8cfoNvnY74dPnu%2Ft0vcwSC9Tw5jZsPv0PVsbpe93Oh7OT%2F%2B7mw9nJS2biz2avURKtXaHf4xT4M8H2GKlzpO0kHS8W6fbPV%2FYQ%2BmhQdlAx1wnGxlOATRl12GO8ooW%2FYpsv&srcPrefix=%2Fen-US%2Fdocs%2FWeb%2FAPI%2FDataTransferItemList%2F))

src/react/atlascode/rovo-dev/rovoDevView.tsx
sdzh-atlassian: NVM found the answer to this below, we now filter this elsewhere

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
```diff
@@ -196,4 +209,11 @@ export class AssignedWorkItemsViewProvider extends Disposable implements TreeDat
             ? issues.map((issue) => new JiraIssueNode(JiraIssueNode.NodeType.JiraAssignedIssuesNode, issue))
             : [];
     }
+
+    public async handleDrag(source: readonly TreeItem[], treeDataTransfer: DataTransfer, _token: CancellationToken) {
+        treeDataTransfer.set(
+            'application/vnd.code.tree.jiraAssignedWorkItem',
```
sdzh-atlassian, line 215: QQ - where is this value coming from?

src/react/atlascode/rovo-dev/prompt-box/prompt-input/PromptInput.tsx
sdzh-atlassian: > unrelated to this change actually

IMO finding an unrelated and undocumented change when going through a big PR is a bit of an issue :)

I get the point of the change itself and it's fine for this PR, but please consider not bundling things in the future

src/react/atlascode/rovo-dev/rovoDevView.tsx
sdzh-atlassian: Ah, got it, ty!

src/react/atlascode/rovo-dev/utils.tsx
marcomura: yeah, `DataTransferItemList` doesn't expose `map` and it can't even be iterated with a `for of`

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
marcomura: doc suggests to add a `application/vnd.code.tree.customid` mime value here 😄

src/react/atlascode/rovo-dev/rovoDevView.tsx
marcomura: I don't have a strong opinion here, but since it's so small I would keep it there

src/react/atlascode/rovo-dev/utils.tsx
marcomura: The `Array.from(items)` does work, I just tried it. But I'm not sure if it's a better approach since it's allocating one extra array unnecessarily...

sdzh-atlassian: I guess it's just a matter of preference :) Happy to keep this as-is

src/react/atlascode/rovo-dev/rovoDevView.tsx
sdzh-atlassian: Sure, let's keep it then :)

src/views/jira/treeViews/jiraAssignedWorkItemsViewProvider.ts
sdzh-atlassian: Oh, I see, it's in the docs for [handleDrag](https://github.com/microsoft/vscode/blob/dc2f5d8dd1790ac4fc6054e11b44e36884caa4be/src/vscode-dts/vscode.d.ts#L9843-L9858) itself

src/react/atlascode/rovo-dev/prompt-box/prompt-input/PromptInput.tsx
marcomura: fair, I even forgot it was in the same branch when I opened this PR

Issue #1161, opened 2025-10-13
Does the public build collect or send telemetry/usage analytics?


robertlanier:
Hi team — could you please provide an extension-specific statement about telemetry/analytics for Atlassian: Jira, Rovo Dev, Bitbucket (atlassian.atlascode)?

- Does the public build collect or send telemetry/usage analytics?
  - If yes, please list the event types/fields, destinations (domains/services), and retention.
  - Is any personal data (user ID/email/account/site URL) included in events?

- Is experimentation/feature-flagging (FX3) active in public builds?
  - If active, what data is sent for assignment/evaluation?
  - If inactive without an internal key, please confirm no telemetry is emitted from FX3 in public builds.

- Opt-out / controls:
  - Is there an extension setting or environment variable to disable telemetry/experiments?
  - Does the extension honor VS Code’s telemetry.telemetryLevel=off?

- Documentation:
  - Could you add a short “Privacy/Telemetry” section to the README, Marketplace page, and/or under the Security section of this Github repo that summarizes the above and links to relevant Atlassian privacy terms?

This clarification helps security/compliance reviews (e.g., for regulated environments). Thanks!

cabella-dot:
Hi @robertlanier thank you for raising this! We took your suggestion and created this wiki: https://github.com/atlassian/atlascode/wiki/Privacy-&-Telemetry with telemetry/experimentation data collection information. Please let us know if this will suffice for you and if you need any further information

robertlanier:
Thank you!! That is exactly what I was looking for. I appreciate the transparency!!

cabella-dot:
No worries! Again, thank you for raising this.

PR #1223, opened 2025-10-23
AXON-1402: add config for telemetry and experimentation

cabella-dot:
### What Is This Change?

This addresses concerns raised in https://github.com/atlassian/atlascode/issues/1161

This PR adds impl so extension abides by VS Code's `telemetry.telemetryLevel` so if a user has their `telemetryLevel` as `off` we do not send analytics or include them in experiments.

Demo:

https://www.loom.com/share/23c0d23bf54f4db2a5333c42340998d0?sid=9a049d0c-c80c-466e-ae26-2e64e09c8846

### How Has This Been Tested?

manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/util/featureFlags/featureFlagClient.test.ts
+++ b/src/util/featureFlags/featureFlagClient.test.ts
@@ -34,6 +34,15 @@ jest.mock('./utils', () => {
     };
 });
 
+jest.mock('vscode', () => {
+    return {
+        ...jest.requireActual('jest-mock-vscode').createVSCodeMock(jest),
+        env: {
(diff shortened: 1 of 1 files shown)


marcomura (changes requested):
Thanks for doing this change, @cabella-dot . We definitely need to respect the built-in telemetry setting.

Let's discuss this change tomorrow, but I believe respecting the `telemetry.telemetryLevel` native settings should suffice as implementing extra settings would be redundant.

marcomura (dismissed):

PR #1227, opened 2025-10-24
chore: revert to using npm version directly

sdzh-atlassian:
### What Is This Change?

Revert the use of `yarn version` in release scripts

This is what we're seeing in a `release` job today:

image

We can repro this locally by doing a shallow checkout, like the one that `release` job does by default:
```sh
git clone --depth 1 -b v4.0.5 git@github.com:atlassian/atlascode.git
```

`yarn version` fails in an environment like that, `npm version` works as expected

IMO we should just revert to using `npm` for versioning.  Based on the [docs](https://classic.yarnpkg.com/lang/en/docs/cli/version/), `yarn version` is pretty opinionated, and intended for more complex usage than what we want here.

Notably, `yarn version` works in nightly builds. I didn't spend the time to properly investigate this - but it's probably one of the 2 differences at play:
 * In nightly, we do `git fetch --tags --force -q` as a first action, via `get-next-nightly.sh`
 * The pipeline is triggered differently, meaning there could be a different setup for `git clone` in the checkout action

In a quick local test, I couldn't get `yarn version` to work even after `git fetch --tags --force -q` though - so I'm not proposing that as a solution here 😛 

### How Has This Been Tested?

Tried `npm version` in a shallow copy, it worked as expected. Also, this is literally what we had before

Basic checks:

- [x] `yarn lint`
- [x] `yarn test`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -55,7 +55,7 @@ jobs:
           restore-keys: yarn-cache-
 
       - name: Set version in package.json
-        run: yarn version $PACKAGE_VERSION --immediate
+        run: npm version $PACKAGE_VERSION --no-git-tag-version --allow-same-version -f
 
       - name: Install dependencies
         run: yarn install --immutable
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -48,7 +48,7 @@ jobs:
           restore-keys: yarn-cache-
 
       - name: Set version in package.json
-        run: yarn version $PACKAGE_VERSION --immediate
+        run: npm version $PACKAGE_VERSION --no-git-tag-version --allow-same-version -f
 
       - name: Install dependencies
         run: yarn install --immutable


.github/workflows/release.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
           restore-keys: yarn-cache-
 
       - name: Set version in package.json
-        run: yarn version $PACKAGE_VERSION --immediate
+        run: npm version $PACKAGE_VERSION --no-git-tag-version --allow-same-version -f
```
bhorai-atl, line 58: we have actions/checkout@v4 (line 32), it performs a shallow clone (fetch 1 commit)
so we do not have history here (btw, nightly build works, because it checks out from main branch, when release checks out from tag)

so, we can try to add (release.yaml, line 32)

```
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
```

this should help

marcomura: thanks @bhorai-atl for taking a look as well.
I'm merging this now to unblock the release, but we can iterate on it

sdzh-atlassian: Thanks @bhorai-atl! I've considered that option as well, but IMO it's probably better to just use `npm version` for what we're doing here. Also, I may or may not have been too lazy to check this solution in the actual pipelines :D

PR #1268, opened 2025-11-06
[AXON-1477, AXON-1478] Fix some errors while enabling Rovo Dev

marcomura:
### What Is This Change?

This change fixes the following errors:
`Error: Rovo Dev instance is already running.`
`Error: Multiple initialization of Rovo Dev attempted`

These were happening because of a race condition between `enableRovoDev()` and `disableRovoDev()` when they were called one after the other.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/rovoDevProcessManager.ts
+++ b/src/rovo-dev/rovoDevProcessManager.ts
@@ -7,6 +7,7 @@ import path from 'path';
 import { RovoDevLogger } from 'src/logger';
 import { downloadAndUnzip } from 'src/util/downloadFile';
 import { getFsPromise } from 'src/util/fsPromises';
+import { waitFor } from 'src/util/waitFor';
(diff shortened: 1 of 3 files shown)


src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -432,8 +433,20 @@ export abstract class RovoDevProcessManager {
         }
     }
 
-    public static deactivateRovoDevProcessManager() {
+    public static async deactivateRovoDevProcessManager() {
+        // wait for the lock to be released before stopping the instance
+        await waitFor({
```
sdzh-atlassian, line 438: IMO this is super readable, love how `waitFor` looks like now

sdzh-atlassian (approved):
Race conditions are tricky, thanks for looking into it!

PR #1270, opened 2025-11-06
chore: extract rovodev client, update references

sdzh-atlassian:
### What Is This Change?

All this does is move some files around, putting the rovodev serve API client implementation into its own folder to reduce coupling. The only added file is `index.ts` that re-imports things

| Before | After |
|----|----|
| image | image|

The changes here are in imports only

### How Has This Been Tested?

`npm run extension:install` -> poked around to make sure rovodev isn't broken :D

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/rovo-dev/rovoDevWebviewProvider.test.ts
```diff
@@ -108,7 +108,7 @@ jest.mock('../../src/rovo-dev/rovoDevDwellTracker', () => ({
     })),
 }));
 
-jest.mock('../../src/rovo-dev/rovoDevApiClient', () => ({
+jest.mock('../../src/rovo-dev/client/rovoDevApiClient', () => ({
```
sdzh-atlassian, line 111: An interesting way to construct a module path - probably a product of our old config not picking up `src/*` stuff 🤔

cabella-dot (dismissed):
LGTM!

src/rovo-dev/rovoDevWebviewProviderMessages.ts
```diff
@@ -5,12 +5,12 @@ import { DetailedSiteInfo } from '../atlclients/authInfo';
...
 import { DialogMessage } from '../react/atlascode/rovo-dev/utils';
 import {
+    EntitlementCheckRovoDevHealthcheckResponse,
     RovoDevRetryPromptResponse,
     RovoDevTextResponse,
     RovoDevToolCallResponse,
     RovoDevToolReturnResponse,
-} from './responseParserInterfaces';
-import { EntitlementCheckRovoDevHealthcheckResponse } from './rovoDevApiClientInterfaces';
+} from './client';
```
sdzh-atlassian, line 13: Needed to rebase because of a conflict here

PR #1277, opened 2025-11-07
AXON-1485: Setup module aliases

BHulovatyi:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -25,6 +25,20 @@
         "strictPropertyInitialization": true,
         "typeRoots": ["node_modules/@types", "src/typings/"],
         "paths": {
+            "@atlclients/*": ["src/atlclients/*"],
+            "@jira/*": ["src/jira/*"],
+            "@bitbucket/*": ["src/bitbucket/*"],
+            "@config/*": ["src/config/*"],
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -25,6 +25,20 @@
         "strictPropertyInitialization": true,
         "typeRoots": ["node_modules/@types", "src/typings/"],
         "paths": {
+            "@atlclients/*": ["src/atlclients/*"],
```
marcomura, line 28: this will become confusing as we'll don't know if an import comes from a third party package (npm package) or from local code.

also, I think it's not needed as @sdzh-atlassian did a change to allow 'src' to be similarly referenced for a semi-absolute path, like in here:
[
import { IssueSuggestionSettings } from 'src/config/model';
import { QuickFlowAnalyticsEvent } from 'src/onboarding/quickFlow/types';](https://github.com/atlassian/atlascode/blob/537acdd6dea0fe2e6c6fb9677447c34516535425/src/lib/analyticsApi.ts#L1-L2)

PR #1284, opened 2025-11-10
[AXON-1487] extract RovoDevLogger and use it consistently

sdzh-atlassian:
### What Is This Change?

I've noticed that we have a `RovoDevLogger` defined already, but we weren't using it consistently throughout the rovodev code

This PR addresses that:
 * Move `RovoDevLogger` into a separate file to live in `src/rovo-dev` 
 * Use it across rovodev instead of the generic logger
 * Change the generic logger a little bit to avoid a circular dependency with `container`

One example of the dependency:
 ```
(src/logger.ts) <- (src/rovo-dev/util/rovoDevLogger.ts) <- (src/rovo-dev/rovoDevProcessManager.ts) <-(src/commands.ts) <- (src/container.ts) <- (src/logger.ts) <- (src/bitbucket/bitbucket-cloud/pullRequests.ts) <- (src/atlclients/clientManager.test.ts)
 ```

Basically, the gist of it was, we import `container` from `logger`, and `container` imports a WHOLE LOT of other things - including `logger` itself. So, when rovodev logger is in its own file, this is what happens in basically every test:

```
  ● Test suite failed to run

    TypeError: Class extends value undefined is not a constructor or null

      1 | import { Logger, retrieveCallerName } from '../../logger';
      2 |
    > 3 | export class RovoDevLogger extends Logger {
```

### How Has This Been Tested?

Same as the rest of these changes - small sanity check against a properly built extension

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/rovo-dev/util/rovoDevLogger.ts
```diff
@@ -0,0 +1,17 @@
+import { Logger, retrieveCallerName } from '../../logger';
```
sdzh-atlassian, line 1: Moved from `src/logger.ts`, see that file below

src/logger.ts
```diff
@@ -4,7 +4,16 @@ import { EventEmitter } from 'vscode';
 import { ErrorProductArea } from './analyticsTypes';
 import { configuration, OutputLevel } from './config/configuration';
 import { extensionOutputChannelName } from './constants';
-import { Container } from './container';
+
+// Let's prevent a circular dependency with Container and the other modules that it imports
+function isDebugging(): boolean {
```
sdzh-atlassian, line 9: There's probably a better rework to be done for this - arguably `logger` shouldn't be depending on `container` in the first place, especially for something this small - since it's inherrently circular. However, that felt out of scope for this work :)

Happy to move this thing out of `container.ts` in another PR, if that makes more sense:
```
    private static _isDebugging: boolean | undefined;
    public static get isDebugging() {
        if (this._isDebugging === undefined) {
            try {
                const args = process.execArgv;

                this._isDebugging = args ? args.some((arg) => isDebuggingRegex.test(arg)) : false;
            } catch {}
        }

        return !!this._isDebugging;
    }
```

PR #1068, opened 2025-09-30
[AXON-1268, AXON-1267] Update logic for rendering Jira suggestions in Rovo Dev

marcomura:
### What Is This Change?

- Simplified the logic, and cleaned up some message types and states
- Jira items are now fetched only from the site associated with Rovo Dev
- Only "To Do" items are fetched
- Items are correctly cached in the view for future reusing, until the items are updated
- Items gets updated every 60 seconds to ensure freshness

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

PR #1286, opened 2025-11-12
fix: don't call API if issues are found in cache

sdzh-atlassian:
### What Is This Change?

Noticed a small broken conditional, restored what was probably the intended logic here - go to API if cache doesn't return anything. We probably missed it on introduction [here](https://github.com/atlassian/atlascode/pull/1068/files#diff-6dd5455adc5e4421d8e6ef6722c03b39255364f53133fe9c640ff2b39aaa5f39R63-R66)

### How Has This Been Tested?

`npm run dev` - poked around with breakpoints, with and without cache. Looks the same - but skipping the call makes the loading a bit faster

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/rovo-dev/rovoDevJiraItemsProvider.ts
+++ b/src/rovo-dev/rovoDevJiraItemsProvider.ts
@@ -77,7 +77,7 @@ export class RovoDevJiraItemsProvider extends Disposable {
 
         // Filter to only include MinimalIssue items (not IssueLinkIssue)
         let assignedIssuesForSite = this.fetchJiraIssuesFromCache(this.jiraSiteHostname);
-        if (!assignedIssuesForSite || assignedIssuesForSite.length === 0) {
+        if (assignedIssuesForSite.length === 0) {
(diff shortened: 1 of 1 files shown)


src/rovo-dev/rovoDevJiraItemsProvider.ts
```diff
@@ -77,7 +77,7 @@ export class RovoDevJiraItemsProvider extends Disposable {
 
         // Filter to only include MinimalIssue items (not IssueLinkIssue)
         let assignedIssuesForSite = this.fetchJiraIssuesFromCache(this.jiraSiteHostname);
-        if (assignedIssuesForSite) {
+        if (!assignedIssuesForSite || assignedIssuesForSite.length === 0) {
```
marcomura: The original logic is intentional to distinguish between "not found in cache" and "found in cache but it's empty".
In the second case, we shouldn't refetch them since we already know there are none.

sdzh-atlassian: Had a separate chat - dropped the null check but kept the empty array, looks like this was unintended

PR #1287, opened 2025-11-12
[AXON-1520]: Implement starting provider & webview for simplified create work item

cabella-dot:
### What Is This Change?

Demo:

https://www.loom.com/share/234c141b7d6042db8a7dae24b94f69ee

This PR creates creates the initial provider, messaging, and React logic for the sidebar create work item view.

Now:
* Create simple tickets with only summary field
* Navigate right to viewing issue after creation
* Navigate right to Start Work flow after creation 
* Navigate right to Rovo Dev with Work Item as context after creation

Next: 
* Polish UI/Copy 
* Add support to all required fields

*This is behind gate so will not be user facing*

### How Has This Been Tested?

Manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -65,10 +65,10 @@ export const baseConfigFor = (project: string, testExtension: string): Config =>
         global:
             testExtension === 'ts'
                 ? {
-                      statements: 66,
-                      branches: 56,
+                      statements: 65,
+                      branches: 55,
                       functions: 59,
-                      lines: 66,
+                      lines: 65,
                   }
                 : /* tsx */ {
                       statements: 14,


src/react/atlascode/create-work-item/createWorkItemWebview.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+            return undefined;
+        },
+        [state.selectedIssueTypeId, state.selectedProjectId, state.selectedSiteId, state.summary],
+    );
+
+    const onMessageHandler = React.useCallback(
+        (msg: CreateWorkItemWebviewProviderMessage) => {
+            switch (msg.type) {
+                case CreateWorkItemWebviewProviderMessageType.InitFields: {
```
sdzh-atlassian, line 92: Please take this with a grain of salt, but this looks a bit sus to me. We're introducing a reducer - isn't the main point of it to handle this kind of logic?

For instance, if we make `isLoading` part of the `createFormState`, and maybe do something about `handleSubmit` - shouldn't we get something pretty close to

```ts
const onMessageHandler = (msg) => { dispatch(msg); }
```

Basically, feels a bit weird to have this big `switch` statement in 2 places 🤔 But again, I might be totally wrong here, please consider with a grain of salt ;)

src/work-items/create-work-item/createWorkItemWebviewProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+            if (this._isWebviewReady) {
+                await this.initFields();
+            }
+        } catch (err) {
+            console.error('Error initializing Create Work Item webview fields:', err);
+            this._pendingState = false;
+        }
+    }
+
+    private async updateField(fieldType: 'site' | 'project' | 'issueType', id: string): Promise<void> {
```
sdzh-atlassian, line 246: Completely out of scope for this PR, but as a general note - IMO we're badly overusing switch statements across the code.

Please don't actually make this change here, but out of curiosity - is there any reason not to do something like this instead?

```
const fieldUpdaters = {
    site: this.updateSelectedSite,
    project: this.updateSelectedProject,
    issueType: this.updateSelectedIssueType,
}

const updateFunc = fieldUpdaters[field]?.(id)
```

src/react/atlascode/create-work-item/createWorkItemWebview.tsx
cabella-dot: Yeah that's true! I can get it so all message handling happens in the reducer

sdzh-atlassian (approved):
Ship it! (or feel free to address any of the NIT comments) :D

PR #1288, opened 2025-11-14
AXON-1516: Call API to check Rovo Dev Entitlement

cabella-dot:
### What Is This Change?

Creates the `RovoDevEntitlementChecker` which calls the endpoint to see if a user's site has correct entitlements to access Rovo Dev.

Did not change current Rovo Dev implementation

Checks entitlement on startup and emits [entitlement event](https://data-portal.internal.atlassian.com/analytics/registry/92438?tabIndex=Details)

### How Has This Been Tested?

UTs for new class

manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -171,7 +171,7 @@
                 "@types/vscode": ">=1.77.0 <=1.96.0",
                 "@types/websocket": "^1.0.0",
                 "@types/ws": "^8.18.1",
-                "@vscode/vsce": "^3.6.0",
+                "@vscode/vsce": "^3.7.0",
                 "ajv": "^8.17.1",
                 "ajv-formats": "^3.0.1",
                 "audit-ci": "^7.1.0",
@@ -1540,12 +1540,6 @@
                 "react": "^18.2.0"
             }
         },
-        "node_modules/@atlaskit/editor-common/node_modules/argparse": {
-            "version": "2.0.1",
-            "resolved": "https://packages.atlassian.com/api/npm/npm-remote/argparse/-/argparse-2.0.1.tgz",
-            "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
-            "license": "Python-2.0"
-        },
         "node_modules/@atlaskit/editor-common/node_modules/date-fns": {
             "version": "2.30.0",
             "resolved": "https://packages.atlassian.com/api/npm/npm-remote/date-fns/-/date-fns-2.30.0.tgz",
@@ -2130,12 +2124,6 @@
             "dependencies": {
                 "@babel/runtime": "^7.0.0"
             }
-        },
-        "node_modules/@atlaskit/editor-markdown-transformer/node_modules/argparse": {
-            "version": "2.0.1",
-            "resolved": "https://packages.atlassian.com/api/npm/npm-remote/argparse/-/argparse-2.0.1.tgz",
-            "integrity": "sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==",
-            "license": "Python-2.0"
         },
         "node_modules/@atlaskit/editor-markdown-transformer/node_modules/entities": {
             "version": "3.0.1",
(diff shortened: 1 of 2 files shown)


src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
```diff
@@ -0,0 +1,134 @@
+import { rovoDevEntitlementCheckEvent } from 'src/analytics';
```
sdzh-atlassian, line 1: Would you mind moving this under `src/rovo-dev/util` instead @cabella-dot? Since this is only relevant to rovodev, it's probably best to consolidate it with the rest of the code IMO

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
```diff
@@ -0,0 +1,134 @@
...
+    private _analyticsClient: AnalyticsClient;
+
+    constructor(analyticsClient: AnalyticsClient) {
+        super(() => {});
+        this._analyticsClient = analyticsClient;
+    }
+
+    public async checkEntitlement(): Promise<boolean>;
+    public async checkEntitlement(site: DetailedSiteInfo): Promise<boolean>;
+    public async checkEntitlement(site?: DetailedSiteInfo): Promise<boolean> {
```
sdzh-atlassian: Is there a point in having this overload? I might be missing something, but why not just do
```ts
public async checkEntitlement(site?: DetailedSiteInfo): Promise<boolean>
```

Edit: also, are we actually ever using the `site` version of this? If not, perhaps we could simplify?

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
```diff
@@ -0,0 +1,134 @@
...
+
+            // If checking specific site, see if valid
+            if (site) {
+                credentials = allValidCredentials.find((cred) => cred.host === site.host);
+            } else {
+                credentials = allValidCredentials.find((cred) => cred.isValid);
+            }
+
+            if (!credentials) {
+                throw new Error('No valid Rovo Dev credentials found for the specified site');
```
sdzh-atlassian: NIT: it's possible to get this error text without actually specifying the site, right?

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
```diff
@@ -0,0 +1,134 @@
...
+        } catch (err) {
+            Logger.error(err, 'Unable to check Rovo Dev entitlement');
+            rovoDevEntitlementCheckEvent(false, err.message).then((e) => {
+                this._analyticsClient.sendTrackEvent(e);
+            });
+            return false;
+        }
+    }
+
+    private async getValidCredentials(): Promise<RovoDevCredential[] | undefined> {
```
sdzh-atlassian: Probably a SUPER NIT but this feels a bit sus:

 * Assuming this is generic "give me all valid credentials of a certain type" logic, 
     - Do we actually want something like this for the check?
     - Is this in line with how we get credentials for rovodev later?
     - Maybe it would best fit in `siteManager` or `credentialManager`?
 * The error handling here mentions rovodev entitlement, but it's now what the block actually does. Why not just throw something else - we would catch it in `checkEntitlement` anyway

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                const authInfo = await Container.credentialManager.getApiTokenIfExists(site);
+
+                if (!authInfo) {
+                    return;
+                }
+
+                // verify the credentials work
+                let isValid: boolean;
```
sdzh-atlassian: MEGA NIT, but I've seen us use this exact pattern in a bunch of places, and it feels a bit off to me with how we have to use a `let`. Should we maybe do something along the lines of

```ts
// here
if (!Container.clientManager.hasValidCredentials(site)) {
    return;
}

// ClientManager.ts
public hasValidCredentials(site: DetailedSiteInfo): Promise<boolean> {
    try {
        await this.jiraClient(site);
        return true;
    } catch (/* whatever */) {
        // log or suppress idk :D
        return false;
    }
}
```

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
cabella-dot: Sounds good! Sorry I was under the impression all the code in that directory was being removed

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
cabella-dot: I added that optional param if we end up needing to check per site going forward, but if not I can remove!

src/util/rovo-dev-entitlement/rovoDevEntitlementChecker.ts
cabella-dot: I mirrored the logic of how we get valid credentials in `rovoDevProccessManager` so the site we validate is the one we actually use for the rovo dev process. I agree, we can put this in credential manager to share logic

sdzh-atlassian (dismissed):

sdzh-atlassian (approved):

PR #1300, opened 2025-11-18
[AXON-1488] chore: export types in rovodev via extension API

sdzh-atlassian:
### What Is This Change?

This PR identifies the types we use in rovodev related to authentication (`AuthInfo`, `DetailedSiteInfo`, `ProductJira`), and consolidates the imports - so that we could easily copy the underlying data structures later

No changes in behaviour or presentation, refactoring only

### How Has This Been Tested?

`npm run extension:install` -> small sanity checks to make sure things still run as expected

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/rovo-dev/api/extensionApiTypes.ts
+++ b/src/rovo-dev/api/extensionApiTypes.ts
@@ -1,0 +1,3 @@
+// Placeholder for simple types from the extension that might need to be copied later
+// This file is kept separate to use in `tsx` files
+export { AuthInfo, DetailedSiteInfo, ProductJira } from '../../atlclients/authInfo';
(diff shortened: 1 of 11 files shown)


src/rovo-dev/api/extensionApiTypes.ts
```diff
@@ -0,0 +1,3 @@
+// Placeholder for simple types from the extension that might need to be copied later
+// This file is kept separate to use in `tsx` files
+export { AuthInfo, DetailedSiteInfo, ProductJira } from '../../atlclients/authInfo';
```
sdzh-atlassian, line 3: The gist of the change - later, instead of re-importing the types we'll be able to just copy them here

PR #1310, opened 2025-11-19
Fix for the new Jira filters when the new Settings page is enabled

marcomura:
### What Is This Change?

The root cause of the issue was the register command for `Filter` happening twice when the new Settings page is enabled, because the registration command was put in two different places.

This issue was likely caused by the bad organization of the register commands.

This change fixes the issue, and refactors the register commands to be cleaner and clearer.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`


sdzh-atlassian (approved):
> This change fixes the issue, and refactors the register commands to be cleaner and clearer.

This should _really_ not have been a 1 PR IMO - but I've seen this work so let's ship it :)

PR #1312, opened 2025-11-20
Fix primary site user info for Rovo Dev feedbacks

marcomura:
### What Is This Change?

This changes fixed the primary site UserInfo data that is used to send Rovo Dev feedbacks.

The UserInfo structure is now provided directly by the Rovo Dev webview provider, which comes from the same site that Rovo Dev used for authentication and billing purposes.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`

### DIFF
--- a/src/rovo-dev/rovoDevProcessManager.ts
+++ b/src/rovo-dev/rovoDevProcessManager.ts
@@ -4,7 +4,7 @@ import fs from 'fs';
 import net from 'net';
 import packageJson from 'package.json';
 import path from 'path';
-import { UserInfo } from 'src/atlclients/authInfo';
+import { UserInfo } from 'src/rovo-dev/api/extensionApiTypes';
 import { downloadAndUnzip } from 'src/rovo-dev/util/downloadFile';
 import { getFsPromise } from 'src/rovo-dev/util/fsPromises';
 import { waitFor } from 'src/rovo-dev/util/waitFor';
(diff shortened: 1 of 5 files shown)


src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -4,12 +4,13 @@ import fs from 'fs';
 import net from 'net';
 import packageJson from 'package.json';
 import path from 'path';
+import { UserInfo } from 'src/atlclients/authInfo';
```
sdzh-atlassian: Small NIT - would you mind importing `UserInfo` via `src/rovo-dev/api/extensionApiTypes` here and elsewhere?

```
// extensionApiTypes.ts
export { UserInfo } from 'src/atlclients/authInfo';

// here
import { ExtensionApi, ValidBasicAuthSiteData, UserInfo } from './api/extensionApi';
```

I've been using that file to keep track of small data types that would need to be copied over later.

Happy to do this myself in a follow-up, just figured it could be good to bring up for awareness :D

src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -272,7 +277,8 @@ export abstract class RovoDevProcessManager {
             if (totalBytes) {
                 this.setState({
                     state: 'Downloading',
-                    jiraSiteHostname: credentialsHost,
+                    jiraSiteHostname: credentials.host,
+                    jiraSiteUserInfo: credentials.authInfo.user,
```
sdzh-atlassian, line 281: QQ: Sorry, I may have not had enough coffee today - why do we need to change these on download progress change?

src/rovo-dev/rovoDevProcessManager.ts
marcomura: done :)

src/rovo-dev/rovoDevProcessManager.ts
marcomura: I simply added `jiraSiteUserInfo` everywhere `jiraSiteHostname` was available for simplicity

PR #1313, opened 2025-11-20
[AXON-1303, AXON-1363] Improve the UI for the Jira work item and the Create Jira issue page

sspanchenko:
### What Is This Change?

We got rid of the dismiss button. Depending on the error, we offer the user either retry or sign in.

image

image

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/webviews/components/issue/view-issue-screen/JiraIssuePage.tsx
```diff
@@ -906,6 +903,37 @@ export default class JiraIssuePage extends AbstractIssueEditorPage<Emit, Accept,
             return <Offline />;
         }
 
+        // If there's an error, show only the error banner without the content
```
sspanchenko: Hey @bwieger-atlassian-com I wanted to check with you how you feel about this logic here and on the issue-creation page. I remember we agreed yesterday that on both the `JiraIssuePage` and the `CreateIssuePage` we would show either the error banner or the content. However, today I started to doubt whether this is really the right approach - banners can represent different types of errors, and do we really want to hide the content for absolutely any error? Maybe you know a more appropriate way to handle this, or does this approach still make sense?

bwieger-atlassian-com: @sspanchenko 

If our errors are truly blocking errors, than hiding the content makes sense. 

However, it sounds like we don't have confidence that all of our errors are blocking error. 

So to move forward we can: 
1. Use the existing logic. Let the context always show 
2. OR we can update the non-blocking errors to be a warning message (or simply hide if no value to the user). By doing so, we can assert that only blocking errors are "error" messages to the user where the content is hidden.

marcomura (approved):
We probably want to iterate on the design for it, but for now we don't have anything from our designers.

Having a CTA button is a big improvement, thanks!
Let's ship it!

PR #1315, opened 2025-11-20
[AXON-1559]: Rovo dev start task immediately after clicking `Generate code` button

jwang19-atlassian:
### What Is This Change?

Run Rovo Dev right after clicking generate code button in issue page

https://www.loom.com/share/a498dbde8ade4aba9992b0055ff03c23

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`


src/webviews/jiraIssueWebview.ts
```diff
@@ -1276,7 +1276,7 @@ export class JiraIssueWebview
                                 url: issueUrl,
                             };
 
-                            await Container.rovodevWebviewProvider.setPromptTextWithFocus(promptText, jiraContext);
+                            await Container.rovodevWebviewProvider.invokeRovoDevAskCommand(promptText, [jiraContext]);
```
bwieger-atlassian-com, line 1279: We should be allowing users to add more context as appropriate. See how BBY does it

bwieger-atlassian-com (changes requested):
We should allow users to modify the prompt

jwang19-atlassian:
Close this PR because user can modify the prompt now, and we will improve to have the dialog like BBY in the future.

PR #1320, opened 2025-11-21
[AXON-1545] refactor rovodev analytics

sdzh-atlassian:
### What Is This Change?

> **Note**: no change to behaviour or presentation - refactoring only

This PR is a somewhat broad rework of how analytics are implemented, scoped to `rovodev`-related stuff.

The main reasoning behind this is to avoid dragging the [legacy implementation of analytics](https://github.com/atlassian/atlascode/blob/main/src/analytics.ts) from into the new `rovodev`, whatever shape that takes :D

| Change | Before | After |
|----|----|----|
| Explicit types for all relevant events | image |  image |
| Finalize and send events in a single step |  image |  image
| [Options object](https://medium.com/@bchadwickfrance/the-options-object-pattern-with-typescript-a14f33306ec8) across the board - no more magic with ordinal parameters | image |  image |
| Single entrypoint for analytics | -/- | `Telemetry`, `Performance`, and `Dwell` events are defined in the same place and all sent through `extensionApi.analytics`

What's NOT in this PR?
 * Actual automatic generation of the event types (doable, but will need to happen later)
 * `performanceEvent` in  `src/analytics.ts` still references rovodev but is now unused, it'll be cleaned up separately
 * We still use `trackEvent` from `src/analytics.ts` to enrich the events with stuff like userId - that will need to be moved separately

### How Has This Been Tested?

Somewhat extensively :D

 * All tests that have been moved around kept their assertions, meaning _there is literally no change to the shape of the events we actually send out_
 * Traced a bunch of event sends in debug
 * [still going] Dumped all events we send to check if the (currently untested) single-event-per-prompt logic holds. Honestly, this is kind of tricky, so I might actually add tests for that in a separate PR

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/rovo-dev/rovoDevTelemetryProvider.ts
+++ b/src/rovo-dev/rovoDevTelemetryProvider.ts
@@ -100,7 +100,7 @@ export class RovoDevTelemetryProvider {
     async fireTelemetryEvent(event: TelemetryEvent): Promise<void> {
         const eventId = `${event.subject}_${event.action}`;
 
-        if (!this.hasValidMetadata(event, this.metadata) && !this.canFire(eventId)) {
+        if (!this.hasValidMetadata(event, this.metadata) || !this.canFire(eventId)) {
             return;
         }
 


src/rovo-dev/analytics/events.ts
```diff
@@ -0,0 +1,175 @@
+// All Rovo Dev analytics events and types
```
sdzh-atlassian, line 1: This file replaces `rovodevAnalyticsTypes` with a different approach to event definition, but same content otherwise

src/rovo-dev/analytics/rovodevAnalyticsApi.test.ts
```diff
@@ -8,295 +7,22 @@ jest.mock('src/container', () => ({
...
-        const mockEvent = { type: 'test', data: 'value' };
-        await analyticsApi.sendTrackEvent(mockEvent);
+        const analyticsApi = new RovodevAnalyticsApi();
+        const mockEvent = { action: 'action', subject: 'subject', attributes: { abc: '123' } };
+        await analyticsApi.sendTrackEvent(mockEvent as any);
 
         expect(Container.analyticsClient.sendTrackEvent).toHaveBeenCalledWith(mockEvent);
     });
-
-    // Rovo Dev event tests
```
sdzh-atlassian, line 38: Moved to `src/rovo-dev/rovoDevTelemetryProvider.test.ts` without changes to the assertions

src/rovo-dev/analytics/rovodevAnalyticsApi.ts
```diff
@@ -1,161 +1,11 @@
...
-        return trackEvent('performanceEvent', 'atlascode', {
-            attributes: { tag, measure, ...(params || {}) },
-        });
-    };
-}
-
-// Rovo Dev events, previously from src/analytics.ts
-// TODO: refactor these, and particularly the usage in RovoDevTelemetryProvider, in a follow-up
-
-export function rovoDevEntitlementCheckEvent(isEntitled: boolean, type: string, source?: string) {
```
sdzh-atlassian, line 26: Deleted in favour of using explicit types and a discriminated union in `analyticsApi.sendTrackEvent`

src/rovo-dev/analytics/events.ts
```diff
@@ -0,0 +1,175 @@
...
+    // TODO: rovodev metadata fields here are different from other events, reconcile later?
+    export type PerformanceEvent = {
+        action: 'performanceEvent';
+        subject: 'atlascode';
+        attributes: {
+            tag: RovodevPerformanceTag;
+            measure: number;
+            rovoDevEnv: RovoDevEnv;
+            appInstanceId: string;
+            rovoDevSessionId: string;
```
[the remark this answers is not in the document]
sdzh-atlassian: Yeah I wish I could do that buddy :D

src/rovo-dev/rovoDevTelemetryProvider.ts
```diff
@@ -57,59 +48,80 @@ export class RovoDevTelemetryProvider {
...
 
-        if (!this._firedTelemetryForCurrentPrompt[funcName]) {
-            this._firedTelemetryForCurrentPrompt[funcName] = true;
+    // This function ensures that the same telemetry event is not sent twice for the same prompt
+    async fireTelemetryEvent(event: TelemetryEvent): Promise<void> {
+        const eventId = `${event.subject}_${event.action}`;
 
-            // add `rovoDevEnv` and `sessionId` as the first two arguments
-            params.unshift(this.rovoDevEnv, this.appInstanceId, this._chatSessionId);
+        if (!this.hasValidMetadata(event, this.metadata) && !this.canFire(eventId)) {
```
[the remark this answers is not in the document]
sdzh-atlassian: Oh wow, good catch

src/rovo-dev/rovoDevTelemetryProvider.ts
```diff
@@ -57,59 +48,80 @@ export class RovoDevTelemetryProvider {
...
+        return true;
+    }
 
-        // the following events can be fired multiple times during the same prompt
-        delete this._firedTelemetryForCurrentPrompt['rovoDevFileChangedActionEvent'];
-        delete this._firedTelemetryForCurrentPrompt['rovoDevCreatePrButtonClickedEvent'];
+    private canFire(eventId: string): boolean {
+        return (
+            // Allow multiple firings for these events
+            eventId === 'atlascode_rovoDevFileChangedAction' ||
```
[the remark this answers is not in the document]
sdzh-atlassian: This is actually correct, just a shorthand for validating 2 fields at once

src/rovo-dev/analytics/events.ts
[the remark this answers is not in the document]
sdzh-atlassian: Yeah I wish this was not the case hahah :)

src/rovo-dev/analytics/rovodevAnalyticsApi.ts
jwang19-atlassian: I saw we have same func in `analytics.ts` [file](https://github.com/atlassian/atlascode/blob/3937c22c05d03ddcce6af6372766a81eaacfc41e/src/analytics.ts#L242)
just curious are these two same event?

sdzh-atlassian: Yes! `entitlementCheckEvent` is something relevant to atlascode, despite having `rovodev` in the name - so it's not actually moved to `events.ts` here, and will just stay in `analytics.ts` for now

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -780,12 +792,15 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
         this._revertedChanges.push(...paths);
 
-        this._telemetryProvider.fireTelemetryEvent(
-            'rovoDevFileChangedActionEvent',
-            this._chatProvider.currentPromptId,
-            'undo',
-            files.length,
-        );
+        this._telemetryProvider.fireTelemetryEvent({
+            action: 'rovoDevFileChangedAction',
```
cabella-dot, line 811: wait, so we only want to fire this event once? What if a user does this type of action multiple times on the same prompt? This goes for the other events too. If we define an event already sent by `event.subject_event.action` the event will only fire on the first time it is done, right?

sdzh-atlassian: Not quite - that particular event, as well as is `rovoDevCreatePrButtonClickedEvent` [explicitly allowed to re-fire](https://github.com/atlassian/atlascode/blob/main/src/rovo-dev/rovoDevTelemetryProvider.ts#L96-L101)

Old code:
```
        // the following events can be fired multiple times during the same prompt
        delete this._firedTelemetryForCurrentPrompt['rovoDevFileChangedActionEvent'];
        delete this._firedTelemetryForCurrentPrompt['rovoDevCreatePrButtonClickedEvent'];

        if (!this._firedTelemetryForCurrentPrompt[funcName]) {
            this._firedTelemetryForCurrentPrompt[funcName] = true;
            // we fire the event here
        }
```

New code:
```
    private canFire(eventId: string): boolean {
        return (
            // Allow multiple firings for these events
            eventId === 'atlascode_rovoDevFileChangedAction' ||
            eventId === 'rovoDevCreatePrButton_clicked' ||
            // Otherwise, only allow if not fired yet
            !this._firedTelemetryForCurrentPrompt[eventId]
        );
    }
```

I think the intent is to only fire allowlisted events multiple times? @marcomura might know more on why it's like this, but all that behaviour should be preserved here :)

cabella-dot (approved):
lgtm

PR #1346, opened 2025-11-26
AXON-1593 fix the metric failure in bby for DAU

bwieger-atlassian-com:
### What Is This Change?
- BBY was failing to track usage because they stopped sending this env var: ROVODEV_BBY see this: https://app.amplitude.com/analytics/atlassian/chart/tp1gocz5?linkingDashboardId=rvcldgga&sharingId=cNrP8tJc
- fix: we can continue to rely on BBY_USERID, so changed the logic to do this

### Testing 
- Test locally emulatiing the BBY environment variable that I see most recently in BBY

### DIFF
--- a/.vscode/launch.json.example
+++ b/.vscode/launch.json.example
@@ -33,7 +33,7 @@
             "env": {
                 "ROVODEV_PORT": "8899",
                 "ROVODEV_ENABLED": "true",
-                "ROVODEV_BBY": "true",
+                "BBY_USERID": "test_user",
                 "ROVODEV_SANDBOX_ID": "bby_sandbox"
             },
             "preLaunchTask": "Start RovoDev Server"
--- a/src/siteManager.ts
+++ b/src/siteManager.ts
@@ -253,7 +253,7 @@ export class SiteManager extends Disposable {
     }
 
     public getFirstAAID(productKey?: string): string | undefined {
-        if (process.env.ROVODEV_BBY && process.env.BBY_USERID) {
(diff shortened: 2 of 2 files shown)


.vscode/launch.json.example
```diff
@@ -33,7 +33,7 @@
             "env": {
                 "ROVODEV_PORT": "8899",
                 "ROVODEV_ENABLED": "true",
-                "ROVODEV_BBY": "true",
+                "BBY_USERID": "test_user",
```
cabella-dot, line 36: You'll probably have to change when we interact with this variable

For example in `container.ts`: 

```ts
public static get isBoysenberryMode() {
        return !!process.env.ROVODEV_BBY;
    }
```

marcomura, line 36: ROVODEV_BBY is a compile-time variable we use to build the extension for BBY in the `atlascode-statlas-mirror` pipeline. It shouldn't be removed.

After build, this variable disappears from the code, so BBY doesn't need to set it.

src/siteManager.ts
```diff
@@ -253,7 +253,7 @@ export class SiteManager extends Disposable {
     }
 
     public getFirstAAID(productKey?: string): string | undefined {
-        if (process.env.ROVODEV_BBY && process.env.BBY_USERID) {
```
marcomura, line 256: looking at this change, something doesn't add up. In BBY this condition is
`if ("true" && process.env.BBY_USERID)` or something like that.

After the change it should be equivalent.

PR #1361, opened 2025-11-30
AXON-1527 Added the Create Bitbucket Pull Request button to the scm

teg-atlassian:
### What Is This Change?
The Source Control Management (SCM) extension has a "Create Pull Request" button that is only useful if the underlying rpo is github.  With this PR, we added a new button "Create Bitbucket Pull Request" that when clicked opens up the UI for creating a pull request for bitbucket repos.
Screenshot 2025-11-30 at 2 01 15 PM

### How Has This Been Tested?

Basic checks:

- [*] `npm run lint`
- [*] `npm run test`

Advanced checks: 

Recommendations:

demo
https://www.loom.com/share/f12631bd9ec54c56936b66890218f6d6

### DIFF
--- a/package.json
+++ b/package.json
@@ -680,7 +680,7 @@
             "scm/title": [
                 {
                     "command": "atlascode.bb.createPullRequest",
-                    "when": "config.atlascode.bitbucket.enabled",
+                    "when": "atlascode:isBitbucketCloudRepo && config.atlascode.bitbucket.enabled",
                     "group": "navigation"
                 }
 			],
--- a/src/bitbucket/bbContext.ts
+++ b/src/bitbucket/bbContext.ts
@@ -1,6 +1,7 @@
 import { Disposable, Event, EventEmitter, Uri } from 'vscode';
 
 import { DetailedSiteInfo, ProductBitbucket } from '../atlclients/authInfo';
+import { CommandContext, setCommandContext } from '../commandContext';
 import { bbAPIConnectivityError } from '../constants';
 import { Container } from '../container';
 import { Logger } from '../logger';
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -677,6 +677,13 @@
                     "when": "resourceScheme == atlascode.bbpr && config.atlascode.bitbucket.enabled"
                 }
             ],
+            "scm/title": [
+                {
+                    "command": "atlascode.bb.createPullRequest",
+                    "when": "config.atlascode.bitbucket.enabled",
```
bwieger-atlassian-com: This when seems too loose.

bwieger-atlassian-com: Shouldn't we have some logic about if the git repo is a bitbucket repo?

cabella-dot: Screenshot 2025-12-01 at 4 00 59 PM

Just as a reference, the github extension still provides the github PR creation button even if I am in `atlascode-issue-analysis`, a bb repo. This is a good point if we want ours to more intentional and I believe we should be

teg-atlassian: Pushed a new commit to address this request.

Usually we could have used the two context params: _scmProvider_ and _scmProviderRootUri_.  Unfortunately for git and bitucket, both repos have the value 'git' for scmProvider and when it comes to scmProverRootUri, that value doesn't include github.com or bitbucket.org.  It is just the plain name of the repo like 'atlascode'

So, I have created a new context called **IsBitBucketCloudRepo** which is boolean.

src/bitbucket/bbContext.ts
```diff
@@ -178,6 +182,10 @@ export class BitbucketContext extends Disposable {
         this._disposable.dispose();
     }
 
+    private static setIsBitBucketCloudRepo(isCloudRepo: boolean) {
```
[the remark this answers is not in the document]
teg-atlassian: accepted

cabella-dot (approved):
Looks good to me! Thanks for making that change

bwieger-atlassian-com (approved):
Nice! And way to include the loom demo!

PR #1371, opened 2025-12-03
fix: show session token in debug panel context

bwieger-atlassian-com:
### What Is This Change?
- as a developer, I need to know what the session token is for the Rovo Dev CLI
- So, show it in the debug panel 

https://www.loom.com/share/003c5ba9770e4be2a4bb75ffae205369


jwang19-atlassian (approved):
🚀

marcomura (approved):
Sweet! You read my mind, I was thinking to do something like that, too 😄 
Thanks!

PR #1394, opened 2025-12-08
[AXON-1644] Support for Full-Context mode in IDE/BBY

marcomura:
### What Is This Change?

This change implements an option in the Rovo Dev prompt box to enable/disable Full-Context mode:

| Menu | Enabled |
|----|-----|
| image | image |

This change includes
- Implementation of the setting in prompt box
- Implementation of the `isAtlassianUser` state to control internal only features
- Implementation of the `/v3/status` request API to obtain the user email in BBY

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -113,6 +113,7 @@
                 "p-queue": "^6.3.0",
                 "p-settle": "^3.1.0",
                 "p-timeout": "^3.2.0",
+                "parse-url": "^8.1.0",
                 "prosemirror-commands": "^1.7.1",
                 "prosemirror-dropcursor": "^1.8.2",
                 "prosemirror-example-setup": "^1.2.3",
--- a/package.json
+++ b/package.json
@@ -1700,6 +1700,7 @@
         "p-queue": "^6.3.0",
         "p-settle": "^3.1.0",
         "p-timeout": "^3.2.0",
+        "parse-url": "^8.1.0",
         "prosemirror-commands": "^1.7.1",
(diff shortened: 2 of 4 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -515,6 +515,17 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
         });
     }
 
+    private async sendProviderReadyEvent(userEmail: string | undefined) {
+        const yoloMode = await this.loadYoloModeFromStorage();
+        await this._webView!.postMessage({
+            type: RovoDevProviderMessageType.ProviderReady,
+            isAtlassianUser: !!userEmail?.endsWith('@atlassian.com'),
```
jwang19-atlassian, line 522: to check atlassian user, are we able to use the existed one in https://github.com/atlassian/atlascode/blob/e3f0c217631becff1b93efff8aa19fa9ce26b950/src/container.ts#L296 or they are diff usage?

marcomura: Yeah, they are different. `isAtlassianUser` is checking if the user is an atlassian employee by checking if any of the auth credentials is from an Atlassian account.

This one is specifically for the credentials used by Rovo Dev.

Also, this one needs to work on Boysenberry where VS Code isn't authenticated at all.

PR #1396, opened 2025-12-09
AXON-727: Prevented duplicate notification

BHulovatyi:
### What Is This Change?

Prevented duplicate notification after restarting VSCode 

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/views/notifications/bannerDelegate.test.ts
+++ b/src/views/notifications/bannerDelegate.test.ts
@@ -9,6 +9,9 @@ import {
     NotificationType,
 } from './notificationManager';
 
+// Helper to flush all pending promises when using fake timers
+const flushPromises = () => new Promise((resolve) => jest.requireActual('timers').setImmediate(resolve));
+
 jest.mock('../../commands', () => ({
     Commands: {
         ShowJiraAuth: 'ShowJiraAuth',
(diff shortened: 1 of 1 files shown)


src/views/notifications/bannerDelegate.ts
```diff
@@ -85,7 +85,11 @@ export class BannerDelegate implements NotificationDelegate {
         this.timer = undefined;
     }
 
-    private showNotification(notification: AtlasCodeNotification, yesText: string, yesAction: () => void) {
+    private async showNotification(notification: AtlasCodeNotification, yesText: string, yesAction: () => void) {
```
sspanchenko, line 94: @BHulovatyi 

`showNotification` is async but called without await in `aggregateAndShowNotifications`. This means `pile.clear()` executes before `markBannerShown` completes, making the `await` here somewhat ineffective. 

Is this intentional?

BHulovatyi: You are right that this could be cleaner. If VSCode closes while aggregateAndShowNotifications is running, some notifications might not have their markBannerShown completed yet

src/views/notifications/bannerDelegate.test.ts
```diff
@@ -149,6 +157,13 @@ describe('BannerDelegate', () => {
 
         // Advance the timer to trigger the display of notifications
         jest.advanceTimersByTime(SHORT_TIMEOUT / 2 + 1); // Simulate the passage of time
+        // Flush promises to allow async aggregateAndShowNotifications and showNotification calls to complete
+        // Need multiple ticks for the full async chain to resolve
+        await Promise.resolve();
```
sspanchenko: @BHulovatyi @bhorai-atl 
Hey team, do you know of a more concise way to do this instead of using a large number of resolves?

I read about `advanceTimersByTimeAsync`. I wonder if it can be used here?

om-ukr: Not sure about `advanceTimersByTimeAsync`, but here is one of the answers on stackoverflow that may be userfull:
image

BHulovatyi: Updated test

BHulovatyi: @sspanchenko @om-ukr Thanks for the comments

PR #1407, opened 2025-12-10
[AXON-1673] chore: add reusable DefinePlugin config

sdzh-atlassian:
### What Is This Change?

Move the duplicate environment setup logic from webpack configurations into a separate file.
There should be no changes to the logic or the builds

### How Has This Been Tested?

`npm run devcompile`
`npm run extension:install`

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/webpack.env.config.js
+++ b/webpack.env.config.js
@@ -1,0 +1,39 @@
+const webpack = require('webpack');
+const dotenv = require('dotenv');
+
+// Load environment variables from .env file
+dotenv.config();
+
+const ENV_VARS = [
+    // FX3 configuration
+    'ATLASCODE_FX3_API_KEY',
+    'ATLASCODE_FX3_ENVIRONMENT',
+    'ATLASCODE_FX3_TARGET_APP',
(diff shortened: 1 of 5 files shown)


webpack.env.config.js
```diff
@@ -0,0 +1,39 @@
...
+
+function createEnvPlugin({ nodeEnv, isBrowser = false }) {
+    const definitions = {
+        ...Object.fromEntries(
+            ENV_VARS.map((varName) => [`process.env.${varName}`, JSON.stringify(process.env[varName])]),
+        ),
+        'process.env.NODE_ENV': JSON.stringify(nodeEnv),
+    };
+
+    // Add process.browser for browser builds
```
marcomura, line 31: do we still need this?

sdzh-atlassian: Honestly no idea, same with `NODE_ENV` which isn't actually referenced anywhere - I just wanted to minimize the potentially intrusive changes with this PR and just kept everything as-is.

PR #1430, opened 2025-12-16
[AXON-1643] Run Rovo Dev as a subprocess, not in terminal

marcomura:
### What Is This Change?

- Move the execution of Rovo Dev from the terminal to a subprocess.
- Log the stderr to our servers 
- Present the stderr to the user for debugging locally 

https://www.loom.com/share/0a04f672485142ef8ba734b4868c84ae

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/src/rovo-dev/ui/common/DialogMessage.tsx
+++ b/src/rovo-dev/ui/common/DialogMessage.tsx
@@ -53,7 +53,7 @@ export const DialogMessageItem: React.FC<{
             parts.push(`\n${msg.statusCode}`);
         }
         if (msg.stackTrace) {
-            parts.push(`\n\nExtensionStack Trace:\n${msg.stackTrace}`);
+            parts.push(`\n\nExtension Stack Trace:\n${msg.stackTrace}`);
         }
         if (msg.stderr) {
             parts.push(`\n\nRovo Dev Stderr:\n${msg.stderr}`);


src/rovo-dev/rovoDevProcessManager.ts
```diff
@@ -512,62 +490,54 @@ class RovoDevTerminalInstance extends Disposable {
                         shellArgs.push('--server-env', 'staging');
                     }
 
-                    this.rovoDevTerminal = window.createTerminal({
-                        name: 'Rovo Dev',
-                        shellPath: this.rovoDevBinPath,
-                        shellArgs,
+                    let stderrData = '';
```
sdzh-atlassian, line 493: Would you happen to have an example of how this actually looks in the error message @marcomura?

marcomura: I didn't pipe the `stderrData` to the error message because I found it's too verbose.
It seems Rovo Dev is writing too much stuff in the stderr :(

But this is something that eventually I'd like to include in the errors / logs, so I'm passing it in the Terminated event just to be ready.

sdzh-atlassian:
Same note as before - I'm happy to get this into nightly as long as the happy path still generally works as expected :)

Love the much more verbose error info

PR #1437, opened 2025-12-18
AXON-1693: add error catching for tool return parsing

cabella-dot:
### What Is This Change?

When parsing tool return messages, sometimes the streaming breaks messages resulting in the breaking of the ui

See: https://atlassian.slack.com/archives/C08HHS3DQTU/p1766020874164619

This PR add error handling for broken tool return messages and better logging

### How Has This Been Tested?

Manually
Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,7 +78,6 @@
                 "@mui/styles": "^5.18.0",
                 "@mui/utils": "^5.17.1",
                 "@segment/analytics-node": "^2.1.3",
-                "@sentry/browser": "^9.40.0",
                 "@sentry/node": "^9.40.0",
                 "@speed-highlight/core": "^1.2.7",
                 "@vscode/codicons": "^0.0.40",
@@ -28142,10 +28141,9 @@
             }
         },
         "node_modules/@remix-run/router": {
-            "version": "1.23.0",
-            "resolved": "https://packages.atlassian.com/api/npm/npm-remote/@remix-run/router/-/router-1.23.0.tgz",
-            "integrity": "sha512-O3rHJzAQKamUz1fvE0Qaw0xSFqsA/yafi2iqeE0pvdFtCO1viYx8QL6f3Ln/aCCTLxs68SLf0KPM9eSeM8yBnA==",
-            "license": "MIT",
+            "version": "1.23.2",
+            "resolved": "https://packages.atlassian.com/api/npm/npm-remote/@remix-run/router/-/router-1.23.2.tgz",
+            "integrity": "sha512-Ic6m2U/rMjTkhERIa/0ZtXJP17QUi2CbWE7cqx4J58M8aA3QTfW+2UlQ4psvTX9IO1RfNVhK3pcpdjej7L+t2w==",
             "engines": {
                 "node": ">=14.0.0"
             }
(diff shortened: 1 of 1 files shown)


src/rovo-dev/ui/utils.tsx
```diff
@@ -109,128 +109,135 @@ export const modifyFileTitleMap: Record<string, ToolReturnInfo> = {
...
+                break;
+            default:
+                // For other tool names, we just return the raw content
+                resp.push({
+                    content: msg.tool_name,
+                });
+                break;
+        }
+    } catch (error) {
+        console.warn('Error parsing ToolReturnMessage:', error, msg);
```
bwieger-atlassian-com: Should this be an error? 

If yes, can we ensure we use something that will report this to our Error dashboard in Amplitude? (and later splunk / sentry)

bwieger-atlassian-com (commented):
See comment s

src/rovo-dev/ui/utils.tsx
cabella-dot: Yes, this try catch will make sure that if anything goes wrong during the parsing, our UI does not crash resulting in a user stuck with a blank screen. I would like to include the metrics for this error in a separate PR to get a work around for this UI crashing bug into prod as soon as possible. Wdyt

cabella-dot: addressed

src/rovo-dev/ui/rovoDevView.tsx
```diff
@@ -158,58 +158,73 @@ const RovoDevView: React.FC = () => {
         setPendingToolCallMessage('');
     }, [keepFiles, totalModifiedFiles]);
 
-    const handleAppendModifiedFileToolReturns = useCallback((toolReturn: RovoDevToolReturnResponse) => {
-        if (isCodeChangeTool(toolReturn.tool_name)) {
-            const msg = parseToolReturnMessage(toolReturn).filter((msg) => msg.filePath !== undefined);
+    const onError = useCallback(
+        (error: Error, errorMessage: string) => {
+            dispatch({
```
jwang19-atlassian, line 164: i saw there are couples of few other usage for dispatch, like `keepFiles`, `undoFiles`. Curious why not directly use `postMessage` here.

jwang19-atlassian: it looks like there are two different patterns in `rovoDevView`: 
1. use postMessage to directly post message to backend, e.g. `sendPrompt`
2. change the `outgoingMessage` state and use `useEffect` to postMessage to backend, e.g. `keepFiles` and `undoFiles`

Not sure if the 2nd one is redundant? or it's handling other things here

src/rovo-dev/ui/rovoDevView.tsx
```diff
@@ -158,58 +158,73 @@ const RovoDevView: React.FC = () => {
...
+                type: RovoDevViewResponseType.ReportRenderError,
```
marcomura, line 165: this message should carry the entire Error message instead of being re-created in the provider, so we can preserve the stack trace for debugging.

can you adjust the code to carry the original error message?

marcomura (approved):
left a comment

src/rovo-dev/ui/utils.tsx
```diff
@@ -107,131 +107,144 @@ export const modifyFileTitleMap: Record<string, ToolReturnInfo> = {
  *
  * @param msg - The tool-return object to parse.
  */
-export function parseToolReturnMessage(msg: RovoDevToolReturnResponse): ToolReturnParseResult[] {
+export function parseToolReturnMessage(
+    msg: RovoDevToolReturnResponse,
+    onError: (error: Error, msg: string) => void,
```
jwang19-atlassian, line 112: it looks like it's a props drilling for it.
i think it will be great to put this in context so there won't be props drilling and the hook can be reused in other components
https://react.dev/learn/passing-data-deeply-with-context

src/rovo-dev/ui/rovoDevView.tsx
cabella-dot: Yeah so we use dispatch for messaging initialized before postMessage is created. This helps with circular dependencies

src/rovo-dev/ui/utils.tsx
cabella-dot: This looks very useful! Thanks I'll look into this

src/rovo-dev/ui/rovoDevView.tsx
cabella-dot: I appended the original error message so we can see which tool call might have broken the rendering

src/rovo-dev/ui/utils.tsx
cabella-dot: I won't make this change in the PR bc with my understanding this would require deeper changes

src/rovo-dev/ui/rovoDevView.tsx
jwang19-atlassian: i see, i feel the dispatch way is like a workaround for circular dependencies...
`onMessageHandler` -> `clearChatHistory` -> 'keepFiles` -> 'postMessage` -> back to `onMessageHandler`
sounds like huge refactoring 🤯

PR #1439, opened 2025-12-18
AXON-449 - prevent auth notifications errors

sky-ocean-spirits:
### What Is This Change?

- Removed fetching notifications for the Server
- Added BasicAuth support to GraphQL client for Cloud API tokens
- Deduplicated credentials based on cloud username/email to ensure only one notification fetch is done per real user (Eg: the situation where a user has both an API token & OAuth signed in)

### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:


src/atlclients/authStore.ts
```diff
@@ -168,13 +168,26 @@ export class CredentialManager implements Disposable {
     public async getAllValidAuthInfo(product: Product): Promise<AuthInfo[]> {
```
bwieger-atlassian-com: Let's rename this function to reflect that this now only get "cloud" auth

sky-ocean-spirits: Created a separate function, since this one is used in [one more place](https://github.com/atlassian/atlascode/blob/main/src/container.ts#L300) and probably should be left as it was

bhorai-atl (approved):

PR #1449, opened 2025-12-30
AXON-1710 add replay analytics

bwieger-atlassian-com:
### What Is This Change?
- Add a replay analtyic event so that we can monitor the health of the replay API in BBY mode

Problem statement: Sometimes, the replay behaves unusually. Ideally, we would see this quickly through our metrics vs waiting for users to report it.

### DIFF
--- a/src/rovo-dev/rovoDevTelemetryProvider.ts
+++ b/src/rovo-dev/rovoDevTelemetryProvider.ts
@@ -31,6 +31,8 @@ export type TelemetryEvent =
     | PartialEvent<Track.ReplayCompleted>;
 
 export class RovoDevTelemetryProvider {
+    private static readonly EVENTS_WITHOUT_PROMPT_ID = new Set(['rovoDevNewSessionAction', 'rovoDevReplayCompleted']);
+
     private _chatSessionId: string = '';
 
     private _firedTelemetryForCurrentPrompt: Record<string, boolean> = {};
@@ -79,10 +81,8 @@ export class RovoDevTelemetryProvider {
             return false;
         }
 
-        // these events are the only events that don't need the promptId
-        const eventsWithoutPromptId = ['rovoDevNewSessionAction', 'rovoDevReplayCompleted'];
-        const hasPromptId = (event.attributes as any).promptId !== undefined;
-        if (!eventsWithoutPromptId.includes(event.action) && !hasPromptId) {
+        const hasPromptId = 'promptId' in event.attributes && event.attributes.promptId !== undefined;
+        if (!RovoDevTelemetryProvider.EVENTS_WITHOUT_PROMPT_ID.has(event.action) && !hasPromptId) {
             this.onError(new Error('Unable to send Rovo Dev telemetry: PromptId not initialized'));
             return false;
         }


src/rovo-dev/rovoDevChatProvider.ts
```diff
@@ -320,6 +320,15 @@ export class RovoDevChatProvider {
 
         if (replayBuffer.length > 0) {
             await this.processRovoDevReplayResponse(replayBuffer);
+
+            // Emit analytics for replay completion
+            this._telemetryProvider.fireTelemetryEvent({
```
marcomura: I think you want to send this event regardless of `replayBuffer.length` being empty or not, to indicate how long it took to render an empty replay

marcomura: Sending the telemetry from here will indicate how long it took to process the replay response from Rovo Dev, parse it, and send it to the webview, which is always going to be very fast since it doesn't include rendering time.

If you want to log how long it took to render it, you should send a signal back from the webview.

bwieger-atlassian-com: Yes, thank you for pointing that out!

bwieger-atlassian-com: Ah, not looking for how long, just whether or not we have messages from the replay API

src/rovo-dev/analytics/events.ts
```diff
@@ -147,6 +147,17 @@ export namespace Track {
         };
     };
 
+    export type ReplayCompleted = {
+        action: 'rovoDevReplayCompleted';
+        subject: 'atlascode';
+        attributes: {
+            rovoDevEnv: RovoDevEnv;
+            appInstanceId: string;
+            rovoDevSessionId: string;
```
sdzh-atlassian: You might want to rename this attribute to `sessionId` to bring this in line with the other events, or provide it explicitly in the invocation if the `sessionId` here is meant to be something different from other events.

If I'm reading this right, that's why the build is failing here:
```
Property 'rovoDevSessionId' is missing in type '{ messagePartsCount: number; }' but required in type 'Omit<{ rovoDevEnv: RovoDevEnv; appInstanceId: string; rovoDevSessionId: string; messagePartsCount: number; }, keyof CommonSessionAttributes>'.
    326 |                 action: 'rovoDevReplayCompleted',
    327 |                 subject: 'atlascode',
  > 328 |                 attributes: {
        |                 ^^^^^^^^^^
    329 |                     messagePartsCount: replayBuffer.length,
    330 |                 },
    331 |             });
```

This happens because of how the common attributes are defined in `rovoDevTelemetryProvider.ts` - arguably we actually _want_ them to be uniform across events, right?
```
export type CommonSessionAttributes = {
    rovoDevEnv: RovoDevEnv;
    appInstanceId: string;
    sessionId: string;
};
```

sdzh-atlassian: Just looked at the history - looks like this came from a suggestion form Atlassian[bot], eh? :D 

I think your original notation was right, it basically proposed to use the same naming as in `PerformanceEvent`, which is an outlier and marked as such ;D

If anything, I guess the takeaway here is to streamline `PerformanceEvent` down the line hahah

marcomura (approved):

PR #1460, opened 2026-01-06
[AXON-1721]-sentry-logging

teg-atlassian:
### What Is This Change?

This PR introduces Sentry logging.  

- It has encapulated the Sentry logging logic in `sentry.ts`.  
- We only log `errors` to Sentry.  
- Currently this logic is behind a feature flag `atlascode-sentry-logging`.  
- Sentry configuration is in ._env.example_.  To run it locally, please copy the configuration to your local ._env_ file

the logs in sentry can be seen at https://atlassian-2y.sentry.io/issues/errors-outages/?project=4510634894688256

Screenshot 2026-01-06 at 2 47 18 PM

### How Has This Been Tested?

**sentry.test.ts** with result shown below
Screenshot 2026-01-06 at 2 41 44 PM

logger.test.ts with result shown below
Screenshot 2026-01-06 at 2 40 53 PM

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     ],
     "main": "./build/extension/extension",
     "rovoDev": {
-        "version": "0.13.22"
+        "version": "0.13.25"
     },
     "scripts": {
         "vscode:uninstall": "node ./build/extension/uninstall.js",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 ### [Report an Issue](https://github.com/atlassian/atlascode/issues)
+## What's new in 4.0.15
+
+### Bug Fixes
+- Added missing labels for PR merge strategies(AXON-1733)
 
 ## What's new in 4.0.14
 
@@ -10,7 +14,7 @@
 
 ### Improvements
 
-- **Rovo Dev Updates**: 
+- **Rovo Dev Updates**:
   - Updated Rovo Dev to v0.13.22
   - Improved keyboard navigation for context and preference buttons (AXON-1697)
   - Made Rovo Dev interface responsive with better width properties and aligned settings actions
--- a/src/bitbucket/bitbucket-cloud/pullRequests.test.ts
+++ b/src/bitbucket/bitbucket-cloud/pullRequests.test.ts
@@ -476,7 +476,14 @@ describe('CloudPullRequestApi', () => {
                 data: {
                     destination: {
                         branch: {
-                            merge_strategies: ['merge_commit', 'squash', 'fast_forward'],
+                            merge_strategies: [
+                                'merge_commit',
+                                'squash',
+                                'fast_forward',
+                                'squash_fast_forward',
+                                'rebase_fast_forward',
+                                'rebase_merge',
+                            ],
                             default_merge_strategy: 'squash',
                         },
                     },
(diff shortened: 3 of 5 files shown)


src/sentry.ts
```diff
@@ -0,0 +1,141 @@
...
+                    scope.setContext('additional context', context.extra);
+                }
+
+                return scope;
+            });
+
+            return scope;
+        } catch (err) {
+            // Silently fail - don't break error logging
+            console.error('Failed to capture exception in Sentry:', err);
```
sdzh-atlassian, line 117: QQ - would it be worth it to have an analytics event here, just to know if this starts happening en masse?

package.json
```diff
@@ -1670,6 +1670,8 @@
         "@mui/styles": "^5.18.0",
         "@mui/utils": "^5.17.1",
         "@segment/analytics-node": "^2.1.3",
+        "@sentry/browser": "^9.40.0",
```
sdzh-atlassian: Do we actually need `@sentry/browser`? Looks like we're only sending stuff via the node client, right?

teg-atlassian: oh.. my first change was doing both.. i will fix that

src/sentry.ts
```diff
@@ -0,0 +1,141 @@
...
+    public async initialize(config: SentryConfig): Promise<void> {
+        // Check if Sentry logging feature flag is enabled
+        const isSentryLoggingEnabled = Container.featureFlagClient.checkGate(Features.SentryLogging);
+
+        // If not enabled, silently return without initializing
+        if (!config.enabled || !config.dsn || !isSentryLoggingEnabled) {
+            return;
+        }
+
+        if (!config.enabled || !config.dsn) {
```
sdzh-atlassian: Possibly unintended duplication?

src/container.ts
```diff
@@ -260,6 +261,25 @@ export class Container {
...
+        try {
+            const sentryConfig: SentryConfig = {
+                enabled: process.env.SENTRY_ENABLED === 'true',
+                dsn: process.env.SENTRY_DSN,
+                environment: process.env.SENTRY_ENVIRONMENT || 'development',
+                sampleRate: parseFloat(process.env.SENTRY_SAMPLE_RATE || '1.0'),
+                atlasCodeVersion: version,
+            };
+
+            if (sentryConfig.enabled && sentryConfig.dsn) {
```
sdzh-atlassian: Do we actually need this check here if we have the same thing in `initialize()` where it silently fails?

I'd suggest that we remove it in one of the places, especially seeing how the conditions are seemingly already out of sync (e.g. this doesn't account for the FF)

src/container.ts
```diff
@@ -260,6 +261,25 @@ export class Container {
...
+                await SentryService.getInstance().initialize(sentryConfig);
+                Logger.info('Sentry initialized successfully');
+            }
+        } catch (error) {
+            Logger.error(error as Error, 'Failed to initialize Sentry');
```
sdzh-atlassian: NIT: Shouldn't this be unreachable with the current implementation?

src/logger.ts
```diff
@@ -156,6 +156,25 @@ export class Logger {
     ): void {
         Logger._onError.fire({ error: ex, errorMessage, capturedBy, params, productArea });
 
+        // Capture to Sentry if initialized
+        try {
+            const { SentryService } = require('./sentry');
```
sdzh-atlassian: Is there any specific reasoning behind importing our own module this way? Could it be that this is a forgotten placeholder?

I'd also consider if we actually need `SentryService` to be a singleton 🤔 Realistically, we want all our error handling to flow here - so perhaps it would make more sense as an instance property of Logger?

src/sentry.ts
teg-atlassian: added

src/container.ts
teg-atlassian: removed it

src/sentry.ts
teg-atlassian: added analytics event

__tests__/setupTests.js
```diff
@@ -5,4 +5,4 @@ global.console = {
     debug: jest.fn(),
     info: jest.fn(),
     warn: jest.fn(),
-};
\ No newline at end of file
+};jest.mock('@sentry/node');
```
sdzh-atlassian: This looks a bit off, is this intended?

jwang19-atlassian (approved):
lgtm

PR #1463, opened 2026-01-07
[AXON-1730] Centralize the hardcoded extension ID

marcomura:
### What Is This Change?

Let there be only one place where 'atlassian.atlascode' is hardcoded.

### How Has This Been Tested?

- [X] `npm run lint`
- [X] `npm run test`
- [X] `manual tests`

### DIFF
--- a/__mocks__/packagejson.ts
+++ b/__mocks__/packagejson.ts
@@ -1,5 +1,7 @@
 module.exports = {
     default: {
+        name: 'atlascode',
+        publisher: 'atlassian',
         rovoDev: {
             version: '1.0.0'
         }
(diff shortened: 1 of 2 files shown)


src/errorReporting.ts
```diff
@@ -4,9 +4,10 @@ import { errorEvent } from './analytics';
 import { AnalyticsClient } from './analytics-node-client/src/client.min';
 import { TrackEvent } from './analytics-node-client/src/types';
 import { ErrorProductArea } from './analyticsTypes';
+import { ExtensionId } from './constants';
```
sdzh-atlassian, line 7: I wonder if it might make sense to have `package.json` as the only source of truth on this? For instance, where `ExtensionId` is defined, we can have something like:

```
import pkg from '../package.json';
export const ExtensionId = `${pkg.publisher}.${pkg.name}`;
```

marcomura: love it!! thanks!

sdzh-atlassian (approved):
LGTM ship it :D

sdzh-atlassian (approved):
Still LGTM :D

Issue #1333, opened 2025-11-25
Finding users while trying to mention doesn't work


kubaWilk:
**Describe the bug**
Trying to mention somebody in comment or jira issue doesn't work - it does not find any users. When I edit comments or issue description via VSCode extension i can see that mentions are done as "[accountid:UUID]" which are later transcribed element to an account full name as an anchor element (directing to browser with a given user's details page in a project). Issue can be seen in Jira Software project in a managed environment.

**To Reproduce**
1. Open Visual Code
2. Open any Jira issue using the extension
3. While adding a comment/editing description click on "@" button to mention somebody
4. A dropdown list shows up that's supposed to show search result. You can see it shows only "No options" as a placeholder in the background
5. Type in user's Full Name/Surname/e-mail associated with the account you are looking for
6. No matches show up. 

**Expected behavior**
Seeing users I can mention while typing in their name/e-mail address

**Screenshots**

<img width="801" height="760" alt="Image" src="https://github.com/user-attachments/assets/e226d4ba-0b1f-444e-95ba-5016bdae5604" />
No results show right after clicking a "@" button

<img width="870" height="757" alt="Image" src="https://github.com/user-attachments/assets/1e441ba7-9c9b-4b5e-b066-26f30ebfddd2" />
"Jakub" is a pretty popular polish name which should show many results yet there's none.

<img width="799" height="315" alt="Image" src="https://github.com/user-attachments/assets/7ab052f0-09d0-44d6-8278-eef4a1ab5fbe" />
Same situation as the first screenshot however in comment section.

<img width="796" height="342" alt="Image" src="https://github.com/user-attachments/assets/fe7d82b1-6115-46c0-b42d-d7ad7998a834" />
Same situation as the second screenshot however in comment section.

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (4.0.9) 
 - Visual Studio Code Version (1.106.2) 
 - Local installation of VSCode and the plugin 
 - Windows 11 24H2 (26100.7171)

sspanchenko:
@kubaWilk 
Thank you for reporting the issue.
Fix should be available in the next versions.

PR #1467, opened 2026-01-07
AXON-1713: No users show up on the search list when using @mention

sspanchenko:
### What Is This Change?

**Summary**:
Fixed a regression where mentions (`@user`) stopped working in Jira issue comments when using the legacy editor.

**Root cause**:
A recent refactoring for the new Atlaskit Editor introduced an additional parameter to the user search function. This caused a conflict with the existing mention picker component, which resulted in malformed API requests and invalid data format being sent to Jira.

**Impact**:
Users were unable to mention colleagues in comments - the user list appeared empty, and manually typed mentions failed to post.

**Resolution**:
Corrected the data flow between components and ensured the WikiMarkup-to-ADF conversion produces valid output that Jira API accepts.

**QA Demo:**
https://www.loom.com/share/7c1ca96ede8742c0bad6447b858533b0

**Source**: https://github.com/atlassian/atlascode/issues/1333

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added missing labels for PR merge strategies(AXON-1733)
 - Fixed authentication error
+- Fixed user mentions not working in Jira issue comments when using editor
 
 ## What's new in 4.0.14
 


src/webviews/components/issue/common/adfToWikimarkup.ts
```diff
@@ -117,16 +118,27 @@ function sanitizeAdf(node: AdfNode): AdfNode {
...
 
-    // Remove null or undefined localId attributes - Jira API doesn't accept them
+    // Remove null/undefined attributes - Jira API doesn't accept them
     if (sanitized.attrs) {
-        if (sanitized.attrs.localId === null || sanitized.attrs.localId === undefined) {
-            // eslint-disable-next-line no-unused-vars
-            const { localId, ...restAttrs } = sanitized.attrs;
-            sanitized.attrs = restAttrs;
+        const cleanedAttrs: Record<string, any> = {};
+        for (const [key, value] of Object.entries(sanitized.attrs)) {
```
om-ukr: nit:
Can we use with `Object.entries(sanitized.attrs).reduce((acc, [key, value])=> {})` instead of `for of` here? 
We accumulating values here, so reduce for me sounds preferable in this case.

src/webviews/components/pullrequest/PopoutMentionPicker.tsx
```diff
@@ -70,7 +70,8 @@ export default class PopoutMentionPicker extends React.Component<
                     components={{ Option: UserOption, DropdownIndicator, IndicatorSeparator: null }}
                     menuIsOpen
                     onChange={this.onSelectChange}
-                    loadOptions={this.props.loadUserOptions}
+                    loadOptions={(inputValue: string) => this.props.loadUserOptions(inputValue)}
```
om-ukr, line 73: Is there a reason why do we need a wrapper function here?

om-ukr (approved):

src/webviews/components/pullrequest/PopoutMentionPicker.tsx
sspanchenko: @om-ukr 

This wrapper became necessary after the refactoring in `JiraIssuePage.tsx` where `fetchAndTransformUsers` was modified to accept an optional second parameter `accountId`:
```javascript
fetchAndTransformUsers = async (input: string, accountId?: string): Promise<UserOption[]>
```

Before this change, the function only took `inputValue`, so passing it directly to `loadOptions` worked fine. But `AsyncSelect` internally calls `loadOptions(inputValue, callback)` with two arguments. After adding the `accountId` parameter, that callback function started being interpreted as `accountId`, causing malformed API requests.

So the wrapper is essentially a fix for the unintended side effect of adding that second parameter - it ensures backward compatibility with how `AsyncSelect` invokes `loadOptions`.

src/webviews/components/issue/common/adfToWikimarkup.ts
sspanchenko: @om-ukr 
 Good point, changed to reduce. Thanks!

om-ukr (dismissed):

PR #1476, opened 2026-01-08
AXON-1600: Unable to stop response when "Rate limit exceeded"

jwang19-atlassian:
### How Has This Been Tested?

Basic checks:

- [ ] `npm run lint`
- [ ] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/rovo-dev/rovoDevChatProvider.ts
+++ b/src/rovo-dev/rovoDevChatProvider.ts
@@ -301,6 +301,14 @@ export class RovoDevChatProvider {
         const replayBuffer: RovoDevResponse[] = [];
 
         while (replayInProgress) {
+            // Check for cancellation during replay
+            if (this.pendingCancellation) {
+                await reader.cancel();
+                isDone = true;
+                this._pendingCancellation = false;
+                break;
+            }
+
             const { done, value } = await reader.read();
             if (done) {
                 isDone = true;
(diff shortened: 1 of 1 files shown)


marcomura:
Is there any indication that this issue is caused by our code?
If calling the API for cancellation doesn't work, it's a Rovo Dev issue that should be fixed there.

src/rovo-dev/rovoDevChatProvider.ts
```diff
@@ -301,6 +301,14 @@ export class RovoDevChatProvider {
         const replayBuffer: RovoDevResponse[] = [];
 
         while (replayInProgress) {
+            // Check for cancellation during replay
+            if (this.pendingCancellation) {
```
marcomura, line 305: I'm concerned about this logic.

You are setting pending cancellation to false when Rovo Dev didn't yet confirm the cancellation.
Which means, you will allow the user to send a new prompt, which is going to fail because there is another prompt being generated (since cancellation didn't happen yet).
The existing logic is blocking the user from sending a new prompt, which is intentional.

Also, even closing the socket with Rovo Dev doesn't result in a cancellation. Rovo Dev can still generate the response when the socket is disconnected.

Moreover, the cancellation API can return a failure. In a perfect world it shouldn't, but it's possible.
In that case we want to surface the error to the user and let them know that the stream is still happening.

PR #1498, opened 2026-01-14
[AXON-1769] Allow users to manually restart their Rovo Dev Process

cabella-dot:
### What Is This Change?

This PR implements the logic to manually restart a rovo dev process through a command.

This will allow for users who are stuck in a non-recoverable bug to be able to hard reset their rovo dev system

Screenshot 2026-01-13 at 3 26 57 PM

Event to track process restarts: https://data-portal.internal.atlassian.com/analytics/registry/95588?tabIndex=Details
### How Has This Been Tested?

Manually

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

---

### DIFF
--- a/src/rovo-dev/rovoDevWebviewProvider.ts
+++ b/src/rovo-dev/rovoDevWebviewProvider.ts
@@ -25,10 +25,13 @@ import {
 } from 'vscode';
 
 import { GitErrorCodes } from '../typings/git';
+import { FeatureFlagClient } from '../util/featureFlags/featureFlagClient';
+import { Features } from '../util/features';
 import { RovodevCommandContext, RovodevCommands } from './api/componentApi';
 import { DetailedSiteInfo, ExtensionApi, MinimalIssue } from './api/extensionApi';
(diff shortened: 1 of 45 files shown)


src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -516,7 +516,8 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
                         break;
 
                     case RovoDevViewResponseType.StartNewSession:
-                        await this.executeNewSession();
+                        const force = e.force ?? false;
```
cabella-dot: This is for future PRs

marcomura:
Overall looks good, but there are some cases to handle.

E.g., if you restart the process while a prompt is generating, you will see some glitch like the panel clears, then more output is printed. Then if you submit a prompt right away, the API gives an error (because the process hasn't started).

Also, ideally it should automatically reload the previous session and continue from there.
We can sync about this later as I still have to merge the session manager.

marcomura (approved):
looks good!

PR #1515, opened 2026-01-19
AXON-1798: add rate limit for ENOTFOUND errors while refreshing tokens

om-ukr:
### What Is This Change?
This PR limits refresh token attempts in case ENOTFOUND was reseived on 5 previous requests.

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/atlclients/authStore.ts
+++ b/src/atlclients/authStore.ts
@@ -543,13 +543,13 @@ export class CredentialManager implements Disposable {
                     Logger.debug(
                         `Skipping token refresh for credentialID: ${site.credentialId} due to previous failures.`,
                     );
+                    return undefined;
                 } else {
                     // reset attempts count after delay has passed
                     if (this._failedRefreshCache.has(site.credentialId)) {
                         this._failedRefreshCache.delete(site.credentialId);
                     }
                 }
-                return undefined;
             }
         }
 


src/atlclients/authStore.ts
```diff
@@ -536,9 +536,19 @@ export class CredentialManager implements Disposable {
...
+                    Logger.debug(
+                        `Skipping token refresh for credentialID: ${site.credentialId} due to previous failures.`,
+                    );
+                } else {
+                    // reset attempts count after delay has passed
+                    if (this._failedRefreshCache.has(site.credentialId)) {
+                        this._failedRefreshCache.delete(site.credentialId);
+                    }
+                }
                 return undefined;
```
sspanchenko: @om-ukr 
Shouldn't `return undefined` be inside the `if (timeSinceLastAttempt < RETRY_DELAY)` block? Currently it returns even after the cache is cleared.

Before, `return undefined` was only executed when both `attemptsCount > 5` AND `time < RETRY_DELAY`. Now it executes whenever `attemptsCount > 5`, regardless of time. Is it expected?

om-ukr: @sspanchenko This `return` only influences when we want to make a call to API on this attempt or will be waiting for the next one. 
Anyway, nice catch, thank you.
I moved it inside the if statement to speed up the token refresh.

PR #1534, opened 2026-01-23
AXON-1817: Fix yaml.schemas and Setting sync conflicts

### LINKED ISSUE #155: yaml.schemas and Setting sync conflicts
**Describe the bug**
When use "Jira and Bitbucket" on different VSCodes of different system users System sync always output merge conflict in settings file, because of your "yaml.schemas" settings, which has absolute path to user's profile directory + version of extension.

**To Reproduce**
1. Setup for two different users VSCode with "Jira and Bitbucket"
2. Setup both VSCodes to same Microsoft account
3. Open first and sync settings
4. Open second and sync settings

**Expected behavior**
No problem sync

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (3.4.11) 
 - Visual Studio Code Version (1.97.2) 
 - Is this a local or cloud environment? (Local) 
 - If Local, OS: Windows

**Additional context**
VSCode has variables for home directory: https://code.visualstudio.com/docs/editor/profiles#_common-questions
And version you can hide inside the json or make it handler more flexible.

om-ukr:
### What Is This Change?
Closes #155 

### How Has This Been Tested?

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

### DIFF
--- a/src/pipelines/yaml/pipelinesYamlHelper.ts
+++ b/src/pipelines/yaml/pipelinesYamlHelper.ts
@@ -2,7 +2,7 @@ import { Logger } from 'src/logger';
 import { commands, ConfigurationTarget, Extension, extensions, Uri, window, workspace } from 'vscode';
 
 import { Resources } from '../../resources';
-import BBPipelinesSchema from './pipelines-schema.json' with { type: 'json' };
+import * as BBPipelinesSchema from './pipelines-schema.json';
 
 const VSCODE_YAML_EXTENSION_ID = 'redhat.vscode-yaml';
 
--- a/src/pipelines/pipelines-schema.test.ts
+++ b/src/pipelines/pipelines-schema.test.ts
@@ -1,7 +1,7 @@
 import Ajv from 'ajv';
 import addFormats from 'ajv-formats';
 
-import * as schema from '../../resources/schemas/pipelines-schema.json';
+import * as schema from './yaml/pipelines-schema.json';
 
 describe('Pipeline schema', () => {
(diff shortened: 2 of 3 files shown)


src/pipelines/yaml/pipelinesYamlHelper.ts
```diff
@@ -29,14 +31,40 @@ async function addPipelinesSchemaToConfigAtScope(
...
+                // remove 'yaml.schemas' entirely
+                await workspace
+                    .getConfiguration()
+                    .update(YAML_SCHEMA_CONFIG_NAME_OF_VSCODE_YAML_EXTENSION, undefined, scope);
+            }
+        } catch (error) {
+            Logger.error(error, `Attempt to remove old 'yaml.schemas' configuration failed: ${error.message}`);
         }
-    });
-    newValue[key] = value;
```
BHulovatyi, line 38: You are deleting outdated entries, updating/removing the config, but you never add newValue[key] = value; as it was before. Is it expected?

om-ukr: Yes this is expected. 
PR is removing old approach with adding path to json schema to configuration. Instead we will relay on yamlPlugin.registerContributor [call](https://github.com/atlassian/atlascode/pull/1534/files#diff-07926156abcce11b95ee80abfd6a2614771d9c68031b26ba34b19f8c9a29efddR115).
So this part of code is going to make sure we remove old configuration in case it is exist in user config.

PR #1585, opened 2026-02-09
 AXON-1865: Fix incorrect Rovo Dev Log File Path

jwang19-atlassian:
### What Is This Change?

Rovo Dev changes it's logs file path to `.rovodev/logs/rovodev.log` [here](https://bitbucket.org/atlassian/acra-python/pull-requests/2219/diff), so this PR is to address this change in IDE
Because old users may still have there log file in `.rovodev/rovodev.log` so we need backward compatibility, so the change in this PR is:
1. Read the log path from `config.yml` file
2. if it's not there, fallback to find `.rovodev/rovodev.log` and `.rovodev/logs/rovodev.log`

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

---

### DIFF
--- a/src/rovo-dev/rovoDevUtils.ts
+++ b/src/rovo-dev/rovoDevUtils.ts
@@ -75,7 +75,7 @@ function getRovoDevLogFilePath(): string | undefined {
         return undefined;
     }
     const configPath = getLogPathFromConfig(home);
-    if (configPath && fs.existsSync(configPath)) {
+    if (configPath) {
         return configPath;
     }
 


src/rovo-dev/rovoDevUtils.ts
```diff
@@ -18,14 +19,39 @@ const FriendlyName: Record<SupportedConfigFiles, string> = {
     'rovodev.log': 'Rovo Dev log file',
 };
 
+function getRovoDevFilePath(home: string, configFile: SupportedConfigFiles): string {
```
marcomura: why introducing this function?

jwang19-atlassian: my init thought was keep it DRY in `getLogPathFromConfig` and `openRovoDevConfigFile` func, but i think it can be removed since it just a one liner, will remove it

src/rovo-dev/rovoDevUtils.ts
```diff
@@ -52,7 +74,15 @@ function getRovoDevLogFilePath(): string | undefined {
     if (!home) {
         return undefined;
     }
-    return path.join(home, '.rovodev', 'rovodev.log');
+    const configPath = getLogPathFromConfig(home);
+    if (configPath && fs.existsSync(configPath)) {
```
marcomura: if the log path is set, but it doesn't exist (yet), should it still return the configured path anyway?

if the new path is set, but it returns the old log, it can be a confusing experience.

jwang19-atlassian: removed the existsSync in https://github.com/atlassian/atlascode/pull/1585/changes/1d1395f359e7d3dd3e968f7dffc3afd7200553e0 to treat path in config as source of truth
only fallback to `.rovodev/rovodev.log` and `.rovodev/logs/rovodev.log` when the path is not config

PR #1593, opened 2026-02-10
Speed up build 

bwieger-atlassian-com:
### What Is This Change?
Speed up the build. 

From about 100 seconds to 24 seconds

Tested via `time npm run devcompile`

---


teg-atlassian:
similar change is not in [the prod file.](https://github.com/atlassian/atlascode/blob/main/webpack.extension.prod.js) is that by design?

PR #1612, opened 2026-02-16
fix: prevent markdown rendering error by ensuring string input

Blastoplex:
Guard against null or undefined markdown input by defaulting to empty string before rendering. This fixes the "Input data should be a String" error caused when markdown content is missing or null in DialogMessageItem, MarkedDown component, and ToolReturnParsedItem components.

image

https://atlassian.slack.com/archives/C09DA8EUW9F/p1770768423830179

devloped in RDiJ and tested with npm run test:react

### DIFF
--- a/src/rovo-dev/ui/common/MarkedDown.test.tsx
+++ b/src/rovo-dev/ui/common/MarkedDown.test.tsx
@@ -1,0 +1,39 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { MarkedDown } from './common';
+
+describe('MarkedDown', () => {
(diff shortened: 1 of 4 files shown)


mattcolman:
How did you arrive at this fix? Did you prove that one of these was null or undefined in the problem cases?
Can we be test driven here and write tests for the cases we saw?

Blastoplex:
yep validating atm. per the types, its expected that the message content could be null/undefined

Will pop in some tests also

Blastoplex:
Updated test cases and validated they failed when the fix wasnt in.

image

amarg-at (approved):

PR #1645, opened 2026-02-23
Unblock CI

sdzh-atlassian:
### What Is This Change?

 * Added a couple overrides for the non-dev dependencies to address the [minimatch advisory](https://github.com/advisories/GHSA-3ppc-4f35-3m26)
 * Bumped the CI npm audit level from `moderate` to `critical` so that the pipelines would unblock

Background: a couple niche packages we use have older dependencies that can't be easily bumped

Created AXON-1932 to return to it once CI is unblocked

### How Has This Been Tested?

Just the CI and E2E tests here - didn't do any manual testing outside of that


teg-atlassian (dismissed):
taking back the approval so that it is not accidentally merged

PR #1650, opened 2026-02-24
Handling-multi-byte-unicode-characters

teg-atlassian:
### What Is This Change?

In the process of the backend RovoDev sending a response message and Atlascode receiving the message, the following transformations occur

- HTTP delivers raw bytes in arbitrary-sized chunks
- The SSE (Server Side Event) parser splits on [\n\n](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) - but this split happens at the byte level, not character level
- If [\n\n](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) falls in the middle of a multi-byte UTF-8 character, the split corrupts the data

This sometimes split tokens such as `****tool_name****` as `***t'` and `ool_name****` and this results in Atlascode throwing error and the session failing.

A more detailed (a little bit long) discussion about the problem and the solution can be found [here](https://docs.google.com/document/d/1vFhwnrsUfpa2DK-XWY2bzkNt9OLCecpjPZirh9jwt70/edit?tab=t.0).
With this PR, we
1. safely parse the the string: we don't assume the string is json
2. when the parsing fails, we attempt again by combining different consecutive chunks.  **So, the trick is just try different combinations.**

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`
- [x]  `new tests`

Advanced checks: 

Recommendations:

---

### DIFF
--- a/src/rovo-dev/client/responseParser.ts
+++ b/src/rovo-dev/client/responseParser.ts
@@ -384,9 +384,25 @@ export class RovoDevResponseParser {
                 continue;
             }
 
+            // Parse JSON data safely - if it fails due to incomplete UTF-8 sequences from chunking,
+            // put it back in the buffer to wait for more data
+            let parsedData: any;
+            if (match[2]) {
+                try {
+                    parsedData = JSON.parse(match[2]);
+                } catch {
+                    // JSON parse failed - likely due to incomplete multi-byte UTF-8 character at chunk boundary
+                    // Put this chunk back in the buffer and wait for more data
+                    this.buffer = chunkRaw + '\n\n' + this.buffer;
+                    continue;
+                }
+            } else {
+                parsedData = '';
+            }
+
             const chunk: RovoDevSingleChunk | RovoDevPartStartChunk | RovoDevPartDeltaChunk = {
                 event_kind: match[1].trim() as any,
-                data: match[2] ? JSON.parse(match[2]) : '',
+                data: parsedData,
             };
 
             let tmpChunkToFlush: RovoDevResponse | undefined;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 - Added warning message when no git repository selected on Start Work page
 - Bumped Rovo Dev version to v0.13.47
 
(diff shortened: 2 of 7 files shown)


src/rovo-dev/client/responseParser.ts
```diff
@@ -384,9 +384,25 @@ export class RovoDevResponseParser {
...
+            // Parse JSON data safely - if it fails due to incomplete UTF-8 sequences from chunking,
+            // put it back in the buffer to wait for more data
+            let parsedData: any;
+            if (match[2]) {
+                try {
+                    parsedData = JSON.parse(match[2]);
+                } catch {
+                    // JSON parse failed - likely due to incomplete multi-byte UTF-8 character at chunk boundary
+                    // Put this chunk back in the buffer and wait for more data
+                    this.buffer = chunkRaw + '\n\n' + this.buffer;
```
marcomura, line 396: match[2] is the end of the chunkRaw.
If match[2] can't be parsed, adding it back to the buffer followed by '\n\n' will keep it broken.

Let's discuss about this

teg-atlassian: this code is wrapped in [another loop,](https://github.com/atlassian/atlascode/blob/Handling-multi-byte-unicode-characters/src/rovo-dev/rovoDevChatProvider.ts#L378-L379) .  So, as you said match[2] is the last token for this buffer but we get new data since we are in the loop.

bwieger-atlassian-com:
My thought here is that this should be solved at the Rovo Dev Server layer, not at the client level... but that's just a first impression. Worth chatting with Tim Esler on this.

marcomura:
Agree with @bwieger-atlassian-com that any issue in the response should be fixed at Rovo Dev level.

However, looking at the telemetry, I don't believe the response is split incorrectly, but I think the tool-response may be responding with a different format than what we are expecting (e.g., string instead of json).

PR #1667, opened 2026-02-27
Axon 1855 teamcamp updated files should not be tracked when replaying a restored session

Migelangel04:
### What Is This Change?

This change address the bug where modified files from a previous chat are being put in the modified files component of the UI. This bug caused the user to see the optional to revert or keep files that could have been deleted, modified by the user, or committed in past commits (a user should not see modified files in a new chat).

Now when a user restores a previous chat, that component will not render.

### How Has This Been Tested?

Since I input an optional parameter to the defined types, I tested for when if it is true, it would not call the `handleAppendModifiedFileToolReturns` in the webview. This was tested manually by deleting, modifying, and restoring chats. Moreover I added unit tests to test the replay responses and if utils would call the described function or not.

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 
 - Rovo Dev: Agent model selection both via /models command, and dedicated drop-down menu
 
+### Bug Fixes
+
+- Fixed a bug where modified files of previous Rovo chats would appear in the tool component.
+
 ## What's new in 4.0.21
 
 ### Bug Fixes


src/rovo-dev/rovoDevChatProvider.ts
marcomura, line 424: this method is also called in Boysenberry to execute the replay on start-up, and for this scenario we do want to track modified files

bwieger-atlassian-com (commented):
We can merge this if we want. But in a very short moment, we will have Rovo Dev keeping track of all of the file change / restore logic. Which will work with replay as well. 

My soft recommendation is to not merge this, as it introduces complexity. And instead wait for the updated APIs from Rovo Dev CLI to manage the changed file management

PR #1683, opened 2026-03-05
AXON-1844 improve fix explain with rovodev

bwieger-atlassian-com:
### What Is This Change?
- Provide more context to the explain & fix with Rovo Dev so that it can be more successful 
- Update how the message is sent to the frontend so that it is always visible 
Recommendations:

---
Rovo Dev code review: <strong>Rovo Dev is reviewing this pull request…</strong>
Refresh the page in a few minutes to see the results.

### DIFF
--- a/src/rovo-dev/rovoDevCodeActionProvider.ts
+++ b/src/rovo-dev/rovoDevCodeActionProvider.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { Logger } from 'src/logger';
 import * as vscode from 'vscode';
 
 import { RovodevCommands } from './api/componentApi';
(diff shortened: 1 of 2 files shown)


src/rovo-dev/rovoDevCodeActionProvider.ts
```diff
@@ -83,4 +87,26 @@ export class RovoDevCodeActionProvider implements vscode.CodeActionProvider {
         };
         return action;
     }
+
+    private buildEnhancedPromptSync(
+        basePrompt: string,
+        document: vscode.TextDocument,
+        range: vscode.Range,
+        diagnostics: readonly vscode.Diagnostic[],
+        intent: 'fix' | 'explain',
```
teg-atlassian, line 96: nit: can we create an interface where this type can be defined so, we can extend it later if the need arises

teg-atlassian (approved):

cabella-dot (approved):
Lgtm!

PR #1720, opened 2026-03-13
FLOW-1731 replace giant atlassian loader

Migelangel04:
### What Is This Change?

Loader for the RovoDev chat before start up.

Before:
Screenshot 2026-03-13 at 1 02 04 PM

After:
Screenshot 2026-03-13 at 1 00 01 PM

### How Has This Been Tested?

Manually:
https://www.loom.com/share/ad95f43310414faeb069ed7b80782474

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

---

### DIFF
--- a/src/react/atlascode/common/AtlLoader.tsx
+++ b/src/react/atlascode/common/AtlLoader.tsx
@@ -19,7 +19,12 @@ const styles = {
         },
     },
     spinnerContainer: {
-        height: '100vh',
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        width: '100%',
+        height: '100%',
+        minHeight: '200px',
     },
     spinner: {
         width: '48px',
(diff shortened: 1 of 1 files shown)


src/react/atlascode/common/AtlLoader.tsx
```diff
@@ -18,13 +18,51 @@ const styles = {
...
 };
 
 const useStyles = makeStyles(styles);
 
-export const AtlLoader: React.FunctionComponent = () => {
+interface AtlLoaderProps {
+    variant?: 'default' | 'rovoDev';
+}
+
+export const AtlLoader: React.FunctionComponent<AtlLoaderProps> = ({ variant = 'default' }) => {
```
jwang19-atlassian, line 55: should we rename this to distinguish jira/BB one?

Migelangel04: I am open to suggestions. I could just change the prop name from varient to isRovo and change from string to boolean. I choose default because it is the loader for every other webview in the app

jwang19-atlassian: oh actually i mean the component name, i may be wrong but isn't this component only for rovo dev or it's also used in jira/BB

Migelangel04: This is used all around Atlascode so yes also in Jira/BB for instance in the settings loading or create issue. I just modified the component for the RovoDev use case.

PR #1735, opened 2026-03-16
AXON-1980 UI for the invoke_subagents tool

Migelangel04:
### What Is This Change?

Added UI support for invoke_subagents in RovoDev chat.
https://www.loom.com/share/55cf2cf69f6f492c87c40c4740db47af

The new UI at a glance:
Screenshot 2026-03-19 at 12 51 44 PM

Screenshot 2026-03-19 at 12 51 58 PM

### How Has This Been Tested?

NOTE: For the invoke_subagents tool has inconsistent behavior due to an external factor that is being investigated current (nothing to do with these changes it was present before). The loom shows a bit a what I mean.

Aside from manual testing like shown in the loom, the two unit tests changed in this PR reflect the UI changes for the ToolCallItem component to render the correct dialog when the agent invokes subagents and displays it accordingly. The second file that was the util to test that the parsing of the data is correct and sends it correctly while testing the empty and missing fields edge cases (like missing description, subagent name, task name).

Basic checks:

- [X] `npm run lint`
- [X] `npm run test`

Advanced checks: 

Recommendations:

---

### DIFF
--- a/src/rovo-dev/ui/utils.tsx
+++ b/src/rovo-dev/ui/utils.tsx
@@ -8,6 +8,7 @@ import {
 } from 'src/rovo-dev/client';
 
 import { RovoDevContextItem } from '../rovoDevTypes';
+import { TodoItem } from './tools/ToDoList';
 
 /**
  * Creates a keyboard event handler that triggers onClick when Enter or Space is pressed.
@@ -41,6 +42,7 @@ export type ChatMessage =
     | RovoDevTextResponse
     | RovoDevToolCallResponse
     | RovoDevToolReturnResponse
+    | RovoDevUpdateTodoResponse
     | RovoDevRetryPromptResponse
     | RovoDevExitPlanModeMessage;
 
@@ -93,6 +95,13 @@ export interface ToolPermissionDialogMessage extends AbstractDialogMessage {
     mcpServer?: string;
     toolCallId: string;
 }
+export interface RovoDevUpdateTodoResponse {
+    event_kind: 'update_todo';
+    todos: TodoItem[];
+    tool_call_id: string;
+    timestamp: string;
+    toolCallMessage: RovoDevToolCallResponse;
+}
 
 export type DialogMessage = ErrorDialogMessage | WarningInfoDialogMessage | ToolPermissionDialogMessage;
 
@@ -102,6 +111,7 @@ export interface ToolReturnParseResult {
     filePath?: string;
     title?: string;
     type?: 'modify' | 'create' | 'delete' | 'open' | 'bash' | 'move' | 'subagents';
+    todoData?: TodoItem[];
 }
 
 export type Response = ChatMessage | ChatMessage[] | null;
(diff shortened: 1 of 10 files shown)


src/rovo-dev/ui/utils.tsx
```diff
@@ -296,6 +302,18 @@ export function parseToolReturnMessage(
...
                 });
                 break;
+            case 'invoke_subagents':
+                const subagentArgs = safeJsonParse<SubagentArgs>(msg.toolCallMessage.args);
+                const subagentNames = subagentArgs?.subagent_names || [];
+                const taskNames = subagentArgs?.task_names || [];
+                for (let i = 0; i < subagentNames.length; i++) {
+                    resp.push({
+                        content: `Subagent: ${subagentNames[i]}`,
+                        title: taskNames[i] || '',
```
marcomura, line 322: how would it look like with an empty title?

src/rovo-dev/ui/rovoDevView.tsx
```diff
@@ -264,8 +266,25 @@ const RovoDevView: React.FC = () => {
                     const last = messages.at(-1);
                     if (last?.event_kind === 'tool-call') {
                         setPendingToolCallMessage(parseToolCallMessage(last.tool_name));
+                        if (last.tool_name === 'invoke_subagents') {
+                            try {
+                                const args = JSON.parse(last.args);
```
marcomura: can we use `safeJsonParse` here instead of try/catch ?

src/rovo-dev/ui/rovoDevView.tsx
```diff
@@ -264,8 +266,25 @@ const RovoDevView: React.FC = () => {
...
+                                const subagentNames: string[] = args.subagent_names || [];
+                                const taskNames: string[] = args.task_names || [];
+                                const tasks: SubagentInfo[] = subagentNames.map((name, i) => ({
+                                    subagentName: name,
+                                    taskName: taskNames[i] || '',
```
marcomura: do we want to push a SubagentInfo object with taskName === '', or should we skip it if taskName is not assigned?

src/rovo-dev/ui/utils.tsx
Migelangel04: UI would look like:
Delegating tasks to subagents
Subagent: Subagent_Name ()
...

Basically from the picture above it would not have the title.

src/rovo-dev/ui/rovoDevView.tsx
Migelangel04: I will give it a go

src/rovo-dev/ui/rovoDevView.tsx
Migelangel04: I would say the taskName is the most important part of the object. The subagent name is not very descriptive as it is something generic like "Explore", "Search", etc and if it running multiple explores like in the PR description example, it does not give much info to the user. A better solution is to ensure each object has a taskName and do the logic as usual. If one is missing, send in an empty SubagentInfo array so that way all the user sees is "Delegating tasks to subagents"

marcomura (approved):

PR #1776, opened 2026-04-13
[AXON-2033] Fix Rovo Dev chat when vscode is powered by Firefox

marcomura:
### Problem
When running the extension in a Firefox-based webview engine (e.g. code-server), users encountered a ChunkLoadError that prevented the RovoDev panel from loading:

```
ChunkLoadError: Loading CSS chunk 4712 failed.
(error: .../build/atlascodeRovoDev.f6b443f0dd5e31a4cd9d.css)
```

This happened because `MiniCssExtractPlugin` and `CompiledExtractPlugin` extract CSS into separate files that are fetched dynamically at runtime by inserting `<link>` tags via JavaScript. Firefox fires `onerror` on these dynamically-inserted `<link>` elements when the resource URL uses the `vscode-resource` scheme, causing webpack’s chunk loader to throw. Chromium-based engines handle this fine.

### Fix
Replaced `MiniCssExtractPlugin.loader` with `style-loader` for all CSS rules in the React webpack bundles (`webpack.react.prod.js`, `webpack.react.dev.js`), and set `extract: false` on `@compiled/webpack-loader`.

`style-loader` injects CSS as `<style>` tags directly into the DOM instead of fetching external CSS files — eliminating the dynamic `<link>` loading entirely. This works correctly because the webview CSP already allows 'unsafe-inline' for style-src.

### Changes
- webpack.react.prod.js / webpack.react.dev.js: replaced MiniCssExtractPlugin.loader → style-loader; set extract: false in @compiled/webpack-loader; removed MiniCssExtractPlugin, CompiledExtractPlugin, and CSSMinimizerPlugin

---


webpack.react.prod.js
```diff
@@ -136,7 +125,7 @@ module.exports = {
     ],
     performance: {
         maxEntrypointSize: 350000,
-        maxAssetSize: 12582912, // 12 MiB for atlaskit chunk
+        maxAssetSize: 13631488, // 13 MiB for atlaskit chunk
```
mattcolman, line 128: why this small bump?

mattcolman:
@marcomura did you test this change visually in all environments? Does this affect vscode IDE, boysenberry (Chrome, Firefox, Safari)?

webpack.react.prod.js
marcomura: the css is now inlined within one of the main asset instead of being chunked independently

marcomura:
Tested with VS Code (both debug and release), and in browser via code-server with Chrome and Firefox.

jwang19-atlassian (approved):

PR #1786, opened 2026-04-16
Revert "Revert "FLOW-1938: Remove Create Pull Request button and related code from chat extension - (Rovo Dev)""

cindy-atl:
Reverts atlassian/atlascode#1783

Pollinators are now also using Create Pull request button in the session header instead of the chat so we can remove the one in the chat

---


mattcolman (approved):

src/rovo-dev/rovoDevWebviewProvider.ts
```diff
@@ -383,32 +372,6 @@ export class RovoDevWebviewProvider extends Disposable implements WebviewViewPro
...
-                            attributes: {
-                                promptId: this._chatProvider.currentPromptId,
-                                prCreated: e.pullRequestCreated,
-                            },
-                        });
-                        break;
-
-                    case RovoDevViewResponseType.CheckGitChanges:
-                        if (!this._prHandler) {
-                            await webview.postMessage({
```
mattcolman, line 399: what was the postMessage used for?

cindy-atl: My understanding of how this works is that before we show the Create Pull request button in the chat, we also check if there are any actual changes. This event seems to be listened to in the chat stream which toggles the hasChanges state based on the response.

PR #1790, opened 2026-04-21
FLOW-2024: expose http endpoint for sending a prompt to rovodev

cindy-atl:
### What Is This Change?

- Start an http server with endpoint to send a prompt to Rovodev so that it is visible in the chat 
- Also surface endpoint for health status 

Jira: https://softwareteams.atlassian.net/browse/AXON-1621 

### How Has This Been Tested?

Tested by running locally 
- Rovodev is initialised properly
- User can still type prompts from the chatbox
- Hitting the new endpoint shows both user prompt and responses in the chat 
- Hitting the endpoint while an existing chat is in process returns 409 (no change to chat) 
- Reloading page shows the full chat history 

---
Rovo Dev code review: <strong>Rovo Dev is reviewing this pull request…</strong>
Refresh the page in a few minutes to see the results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ### [Report an Issue](https://github.com/atlassian/atlascode/issues)
 
+## What's new in 4.0.29
+
+### Features
+
+- **RovoDev (BBY)**: Exposed a local HTTP endpoint `POST http://127.0.0.1:9999/rovodev/chat` that allows external services (e.g. DevAI Sandbox) to send prompts to the Rovo Dev chat UI. Returns `200` on success or `409` if the agent is already busy.
+- **RovoDev (BBY)**: Added `GET http://127.0.0.1:9999/rovodev/health` healthcheck endpoint that returns server status and whether the agent is currently busy.
+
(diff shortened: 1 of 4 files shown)


src/rovo-dev/rovoDevLocalServer.ts
```diff
@@ -0,0 +1,78 @@
+import express, { Request, Response } from 'express';
+import * as http from 'http';
+import { Logger } from 'src/logger';
+import { Disposable } from 'vscode';
+
+export const ROVODEV_LOCAL_SERVER_PORT = 9999;
```
marcomura: nit: allow it to be configured via environment variable

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ### [Report an Issue](https://github.com/atlassian/atlascode/issues)
 
+## What's new in 4.0.29
```
marcomura: I would undo the changelog here since it's not a user-facing feature, and it's not something we need to advertise

marcomura (dismissed):

marcomura (approved):

PR #1814, opened 2026-04-30
AXON-2035: fix: safely stringify tool args in ToolCallBody to prevent React erro…

dchiew-atl:
### What Is This Change?

The customer has been getting this error 
image
likely because React is trying to render and object rather than a string. I've put some guards to JSON.stringify it in that situation so we can see the real error.

### How Has This Been Tested?

Tested before with some mock data:
SCR-20260430-lcyh

Tested after:
SCR-20260430-lelc

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`
- [x] `manual check`

---

### DIFF
--- a/src/rovo-dev/ui/common/DialogMessage.tsx
+++ b/src/rovo-dev/ui/common/DialogMessage.tsx
@@ -30,6 +30,16 @@ function safeJsonParse<T = any>(value: string | T | null | undefined): T {
     return typeof value === 'string' ? JSON.parse(value) : value;
 }
 
+function toDisplayString(value: unknown): string {
+    if (typeof value === 'string') {
+        return value;
+    }
+    if (typeof value === 'number') {
+        return String(value);
+    }
+    return JSON.stringify(value);
+}
+
 export const DialogMessageItem: React.FC<{
     msg: DialogMessage;
     isRetryAfterErrorButtonEnabled?: (uid: string) => boolean;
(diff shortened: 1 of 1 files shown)


amarg-at (approved):

src/rovo-dev/ui/common/DialogMessage.tsx
```diff
@@ -343,7 +353,7 @@ const ToolCallBody: React.FC<{
     if (toolName === 'bash') {
         return (
             <pre style={{ margin: '0' }}>
-                <code style={{ maxWidth: '100%' }}>{jsonArgs.command}</code>
+                <code style={{ maxWidth: '100%' }}>{toDisplayString(jsonArgs.command)}</code>
```
amarg-at, line 356: what is command expected to be? is its shape known? should we try to unwrap it, or is this just to unblock quickly?

cindy-atl (approved):

ccallcottstevens (approved):

src/rovo-dev/ui/common/DialogMessage.tsx
dchiew-atl: it's just to unblock quickly, because we couldn't actually see the full error (because of this bug!). it was an intermittent error so I suspect the LLM may have generated a non-standard response shape. Once we get this fix in we should have more info.

Issue #1818, opened 2026-05-04
Extension constantly reports "There was an error connecting to bitbucket cloud"


crobarcro:
**Describe the bug**
Extension constantly reports "There was an error connecting to bitbucket cloud. Please log in again" and forces me to login again after a fairly short time, at least every day for example. 

**To Reproduce**
1. Install Extension
2. log in using Oauth
3. Wait some time, maybe overnight
4. Error message will appear and you will need to re-login

**Expected behavior**
Logins to remain persistent for a longer period of time.

**Screenshots**

<img width="587" height="243" alt="Image" src="https://github.com/user-attachments/assets/729aeda2-d6c8-4af4-a70b-0b1d60a34a56" />

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (3,140,171) 
 - Visual Studio Code Version: 
Version: 1.108.2 (system setup)
Commit: c9d77990917f3102ada88be140d28b038d1dd7c7
Date: 2026-01-21T13:52:09.270Z
Electron: 39.2.7
ElectronBuildId: 12953945
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Windows_NT x64 10.0.26200

 - Is this a local or cloud environment? (Local) 
 - If Cloud, Browser (N/A)
 - If Local, OS (windows)

PR #1719, opened 2026-03-12
ensure-no-refresh-for-invalidated-credentials

teg-atlassian:
### What Is This Change?

When a credentialId is invalid (because we got a 403 or 401 error), we can't keep refreshing the token since it will keep getting 403 or 401 error.  In this PR, we skip those credentialIds in future refreshes once we identify them as being invalid. 

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`
- [x]  `new tests`

Advanced checks: 
- [ ] If Atlassian employee & Bitbucket changes: did you test with DC in mind? [See Instructions](https://www.loom.com/share/71e5d17734a547f68fd6128be6cd760e?sid=835e58a7-1240-498d-b2d7-fa7fdf8ffa36)

Recommendations:
- [ ] Update the CHANGELOG if making a user facing change

---
Rovo Dev code review: <strong>Rovo Dev couldn't review this pull request</strong>
Upgrade to Rovo Dev Standard to continue using code review.

Issue #1635, opened 2026-02-19
Jira and Bitbucket keep disconnecting


mankowitz:
**Describe the bug**
After a few hours, I lose connection to both Jira and Bitbucket 

**To Reproduce**
1. Sign into Jira and Bitbucket using OAuth
2. Develop
3. Try to commit or view tickets
4. Notification that connection to Bitbucket or OAuth is missing

**Expected behavior**
As recently as a few weeks ago, I could stay logged in for a month or more

**Screenshots**
<img width="859" height="192" alt="Image" src="https://github.com/user-attachments/assets/7c898603-81fc-4598-9099-57df7813961b" />

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version 4.0.19
 - Visual Studio Code Version: 1.109.4 (user setup)
 - Is this a local or cloud environment? Local
 - Version: 1.109.4 (user setup)
 - OS: Windows_NT x64 10.0.26200 
 
**Additional context**
N/A

boycce:
Yes, this just started happening for me too, a right pain! I

In my Atlassian UI Settings -> Bitbucket tab, it says "Credential Error" in the hover popup.

- Atlassian Plugin Version 4.0.19
- Visual Studio Code Version: 1.105.1 (user setup) / Cursor 2.4.31
- Is this a local or cloud environment? Local
- OS: Darwin arm64 24.6.0

<img width="375" height="80" alt="Image" src="https://github.com/user-attachments/assets/42d1aa4c-bc2b-404f-a972-99e3bb9900e7" />

deRaaf:
I have this problem too. Having to login again and again on every project is getting really frustrating.

obaidch24:
I'm facing the same issue; it's very annoying.

PR #1907, opened 2026-07-14
Recover OAuth credentials marked invalid instead of requiring a manual re-login

### LINKED ISSUE #1827: Authentication needed every time a new folder is opened.  this message appears "There was an error connecting to Bitbucket Cloud. Please log in again."
**Describe the bug**
This error message appears constantly in VS Code: "There was an error connecting to Bitbucket Cloud. Please log in again."

**To Reproduce**
1. Go to 'VS Code and open any folder.'
2. Connect to Atlassian using LOGIN WITH OAUTH
3. Open a different folder.
4. See error: "There was an error connecting to Bitbucket Cloud. Please log in again."

**Expected behavior**
No error message should appear when switching to a new folder.  No error message should appear after the first-time logging in.

**Screenshots**

Image

**Environment and Version (please complete the following information):**
 - Atlassian Plugin Version (4.0.28) 
 - Visual Studio Code Version (1.116.0) 
 - Is this a local or cloud environment?  Local
 -  Local, OS is Windows

**Additional context**

NSExceptional:
### What Is This Change?

Fixes #1827 (likely also #1818, and part of #1635).

**Symptom:** "There was an error connecting to Bitbucket Cloud. Please log in again." appears in *every* new VS Code window, even though the user is signed in and their stored refresh token is still valid. Bitbucket features silently stop working while the auth UI continues to show the user as logged in.

**Root cause:** once `refreshAccessToken` persists `AuthInfoState.Invalid` to secret storage, the credential can never recover:

- `softRefreshOAuth` returns early for `Invalid` credentials, so a refresh is never attempted again (#1719).
- `refreshAccessToken` also skips `Invalid` credentials and re-seeds the in-session `permanentFailure` cache, so restarting VS Code doesn't help either (#1740).
- The only code path that sets a credential back to `Valid` is a fresh login (`loginManager.saveDetails`).

Meanwhile `ClientManager.createClient` shows the modal once per `ClientManager` instance (`hasWarnedOfFailure`), and a new instance is created per window — hence "every new window", indefinitely.

The `Invalid` flag is easy to acquire spuriously: `OAuthRefesher.getNewTokens` classifies **any** 401 from the token endpoint as permanent (`shouldInvalidate`), including transient gateway/proxy 401s and RFC 6749 `invalid_client` responses that say nothing about the user's refresh token. (Until #1747 any 403 did too.) On an affected install the credential had been stuck `Invalid` since March; replaying the stored refresh token against `https://bitbucket.org/site/oauth2/access_token` returned **HTTP 200 with fresh tokens** — the invalidation was bogus, but the extension refused to retry for months.

**The change:** treat persisted `Invalid` as "needs re-validation" rather than "dead forever":

- `softRefreshOAuth` no longer skips `Invalid` OAuth credentials; it proceeds to attempt a refresh.
- `refreshAccessToken` no longer short-circuits on `Invalid`; the existing `_failedRefreshCache` backoff still bounds retries (a refresh that comes back `shouldInvalidate` marks `permanentFailure`, so a genuinely dead token is attempted only once per session).
- A successful refresh now explicitly sets `state = AuthInfoState.Valid` before saving, so a spuriously-invalidated credential self-heals on the next `getAuthInfo` call — typically at window startup, before the modal would have been shown.

Behavior for genuinely dead refresh tokens is unchanged: the recovery attempt fails, the credential stays `Invalid`, and the existing modal still appears.

### How Has This Been Tested?

- Updated the two unit tests that asserted the old skip-on-`Invalid` behavior, and added coverage for: recovery success (state persisted back to `Valid`), permanent failure (stays `Invalid` + `permanentFailure` cached), and no same-session retry after a permanent failure. `authStore.test.ts` passes (53/53); the rest of the unit suite is unaffected.
- Confirmed the bug empirically on an affected Cloud install: decrypted the stored credential (`state: 1`) and replayed its refresh token against the Bitbucket token endpoint — HTTP 200, still valid — establishing that the persisted `Invalid` was spurious.
- Built the patched extension and installed it locally against that same install for manual verification that the credential re-validates on startup and the modal no longer appears.

Basic checks:

- [x] `npm run lint`
- [x] `npm run test` (unit)

Advanced checks:
- [ ] If Atlassian employee & Bitbucket changes: did you test with DC in mind? — n/a (external contributor; change is OAuth-cloud-path only, DC/basic/PAT auth is untouched)

Recommendations:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- **Authentication**: OAuth credentials that were previously marked invalid are now re-validated by attempting a token refresh instead of being skipped forever. A credential invalidated by a transient token-endpoint error now self-heals on the next refresh, so users no longer see a persistent "There was an error connecting to ... Please log in again." message in every new window despite still being signed in.
 - **RovoDev**: Hid stack traces, stderr, and log details from external users while preserving them for Atlassian users.
 - **RovoDev (BBY)**: Fixed `ROVODEV_REBRAND_JCA` env var handling so the "Jira Coding Agent" rebrand works correctly in webviews.
 - **Notifications**: Fixed `atlassianNotificationNotifier` to correctly flush all promise levels, resolving a test reliability issue.
--- a/src/atlclients/authStore.test.ts
+++ b/src/atlclients/authStore.test.ts
@@ -1169,8 +1169,7 @@ describe('CredentialManager', () => {
             expect(Logger.debug).toHaveBeenCalledWith(expect.stringContaining('permanent previous failure'));
         });
 
-        it('should skip refreshAccessToken when OAuth credentials are already invalid', async () => {
-            const Logger = require('../logger').Logger;
+        it('should attempt a recovery refresh when OAuth credentials are marked invalid', async () => {
             const site = { ...mockJiraSite, isCloud: true };
             const invalidOAuthInfo: OAuthInfo = {
                 ...mockOAuthInfo,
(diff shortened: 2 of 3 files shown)


twocs:
I agree with the solution to stop considering AuthInfoState.Invalid as a final failed state.

I note that OAuth 2 should return error responses with an HTTP 400 status code along with `error` and `error_description` parameters.

https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

If the Atlassian OAuth 2 server provides these parameters, the auth state can potentially be refined to contain more than simply Valid or Invalid. Unfortunately, the Atlassian documentation does not contain enough information about the various error/error_description that may be returned in their implementation so I agree that Invalid could be retried. 
https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/#use-a-refresh-token-to-get-another-access-token-and-refresh-token-pair

PR #1915, opened 2026-07-22
FLOW-2627: hid live-preview button while agent is running to prevent …

ccallcottstevens:
…double click

### What Is This Change?
The Live Preview button is set to hide when the tool-call is working. Hiding it earlier so that the user doesn't get impatient, click it again and cause the error.

The button in the error also does nothing. We don't want to try again at this point, so I removed this button.
Added a check so if the tool-call actually fails we show the Live-preview button again. 

Added tests

### How Has This Been Tested?

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

Advanced checks: 

Recommendations:

---

### DIFF
--- a/src/rovo-dev/ui/rovoDevView.tsx
+++ b/src/rovo-dev/ui/rovoDevView.tsx
@@ -675,11 +675,13 @@ const RovoDevView: React.FC = () => {
         });
     }, [postMessage]);
 
-    const onCreateLivePreview = useCallback((): void => {
+    // returns true if the live preview was started, false if it was skipped
+    // (agent no longer idle) so the caller can reset its own state
+    const onCreateLivePreview = useCallback((): boolean => {
         // only start when idle, and hide the button immediately to avoid a
         // double-click race that triggers an HTTP 409 ("agent busy")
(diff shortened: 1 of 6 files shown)


src/rovo-dev/ui/rovoDevView.tsx
```diff
@@ -675,6 +675,21 @@ const RovoDevView: React.FC = () => {
         });
     }, [postMessage]);
 
+    const onCreateLivePreview = useCallback((): void => {
+        // only start when idle, and hide the button immediately to avoid a
+        // double-click race that triggers an HTTP 409 ("agent busy")
+        if (currentState.state !== 'WaitingForPrompt') {
+            return;
```
dchiew-atl: if we `return` here does the user get any feedback?

ccallcottstevens: Good catch — returning early left the button stuck with no feedback. Fixed it: onCreateLivePreview now returns a boolean, and the button resets its spinner if the preview didn’t start, so it’s clickable again.

dchiew-atl (approved):

PR #1917, opened 2026-07-26
JOLT-3435 Update message to remove credit total and remaining for Rovo Credits

atlassianwalmsley:
### What Is This Change?

Updating messaging for the Rovo Dev /usage command

### How Has This Been Tested?

Tested locally and increased unit test coverage.

Basic checks:

- [x] `npm run lint`
- [x] `npm run test`

### DIFF
--- a/src/rovo-dev/rovoDevUtils.ts
+++ b/src/rovo-dev/rovoDevUtils.ts
@@ -163,13 +163,22 @@ export function usageJsonResponseToMarkdown(response: RovoDevUsageResponse): {
         }
     } else {
         // If credit_total returns -1, it means unlimited credits
+        // If credit_total is 0 or missing, omit Remaining and Total lines
+        const hasValidTotal =
+            data.credit_total !== undefined &&
+            data.credit_total !== null &&
+            data.credit_total !== 0 &&
+            !Number.isNaN(data.credit_total);
+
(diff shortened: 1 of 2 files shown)


src/rovo-dev/rovoDevUtils.ts
```diff
@@ -163,13 +163,22 @@ export function usageJsonResponseToMarkdown(response: RovoDevUsageResponse): {
         }
     } else {
         // If credit_total returns -1, it means unlimited credits
+        // If credit_total is 0 or missing, omit Remaining and Total lines
+        const hasValidTotal =
+            data.credit_total !== undefined &&
+            data.credit_total !== null &&
+            data.credit_total !== 0 &&
```
lkyles-atlassian, line 170: Just thinking about internal users where dev infra might cap to 0. Not sure if we want to show in that case. 🤔 

I guess as long as the messaging covers that their limit is exceeded then that is probably enough.

lkyles-atlassian (approved):

src/rovo-dev/rovoDevUtils.ts
atlassianwalmsley: Yeah messaging covers that, this just fixes the awkward look where it's like:
used: 200,000
remaining: 0
total: 0

ccallcottstevens (approved):

PR #1932, opened 2026-08-24
Fix credential collision when connecting two Server/Data Center sites with the same PAT username

luisdat:
## Summary
`saveDetailsForSite` in `src/atlclients/loginManager.ts` computed the stored credential ID's `siteId` as `site.product.key` (e.g. `"jira"`) for PAT (Personal Access Token) logins, instead of the site host — unlike Basic Auth, which already used the host:

```ts
const siteId = isBasicAuthInfo(credentials) ? baseLinkUrl : site.product.key;
```

Two different Jira or Bitbucket Server/Data Center instances authenticated via PAT with the same username therefore hash to the identical credential ID (`CredentialManager.generateCredentialId` is `md5(siteId::username)`), so connecting a second such site silently overwrites the first site's stored token — both in the in-memory credential store and in VS Code SecretStorage. The first site then fails auth on its next request and gets marked invalid/disconnected.

This is easy to hit in practice: any organization running multiple on-prem Jira/Bitbucket instances where users authenticate with the same corporate/LDAP username on each.

`saveDetailsForSite` is only ever used for Server/Data Center (non-cloud) sites, so the fix always scopes the credential ID to the site's host, matching the existing Basic Auth behavior:

```ts
const siteId = baseLinkUrl;
```

The OAuth/cloud shared-credential-per-account behavior (a separate, deliberate design for cloud accounts with multiple accessible resources) lives elsewhere and is untouched by this change.

## Test plan
- [x] Added a regression test in `src/atlclients/loginManager.test.ts` verifying two different Server/DC hosts sharing the same PAT username now produce distinct credential IDs.
- [x] `npm test -- src/atlclients/loginManager.test.ts src/atlclients/authStore.test.ts` passes.
- [x] Manually verified: connected two separate Jira Data Center instances via PAT with the same username; both now stay connected simultaneously (previously the first would disconnect with an auth error as soon as the second was added).

---
The pull request author does not have access to Rovo Dev.


luisdat:
Closing to open a fresh PR with a clean CLA check.

